PR #285, opened 2022-01-14
add support for standard notes in importer

ammarahm-ed:


packages/importer/src/providers/standardnotes/types.ts
```diff
@@ -0,0 +1,109 @@
...
+  "Code Editor": {
+    type: "code",
+  },
+  "Bold Editor": {
+    type: "html",
+  },
+  "Simple Markdown Editor": {
+    type: "markdown",
+  },
+};
```
thecodrr, line 77: This doesn't belong here. `types.ts` file should only contain types.

packages/importer/src/providers/standardnotes/types.ts
```diff
@@ -0,0 +1,109 @@
+export type ContentTypes = "Note" | "Tag" | "SN|SmartTag" | "SN|Component";
```
thecodrr, line 1: Instead of a union type here, let's convert this into an `enum`. That `enum` can be used for safe comparision.

packages/importer/src/providers/standardnotes/types.ts
```diff
@@ -0,0 +1,109 @@
...
+  service: string;
+  account: string;
+  secret: string;
+  notes: string;
+  password: string;
+};
+
+export type BackupType = {
+  version: string;
+  items: StandardnotesNote[];
```
thecodrr, line 36: These should be optional.

packages/importer/src/providers/standardnotes/index.ts
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  async process(
+    files: File[],
+    settings: ProviderSettings
+  ): Promise<ProviderResult> {
+    return iterate(this, files, (file, notes) => {
+      if (file.name !== "Standard Notes Backup and Import File.txt")
+        return Promise.resolve(true);
+      let data: BackupType = <BackupType>JSON.parse(file.text);
```
thecodrr, line 201: Check if the parsed data is valid and includes the required properties.

We should also do a version check here and give error.

thecodrr (changes requested):
Not a full review.

packages/importer/src/providers/standardnotes/index.ts
```diff
@@ -0,0 +1,248 @@
...
+  buildTableWithRows(rows: string) {
+    return `<div class="table-container">
+    <table>
+      <tbody>
+        ${rows}
+      </tbody>
+    </table>
+  </div>
+  <p><br data-mce-bogus="1"/></p>`;
+  }
```
thecodrr, line 135: This should not belong inside the class. It is general enough that we can move it to a utility.

Also:
1. Add `contenteditable="false"` on `div` tag
2. Add `contenteditable="true"` on `table` tag

packages/importer/src/providers/standardnotes/types.ts
```diff
@@ -0,0 +1,75 @@
...
+  Note = "Note",
+  Tag = "Tag",
+  SmartTag = "SN|SmartTag",
+  Component = "SN|Component" 
+}
+
+export type EditorType = {
+  type: "code" | "html" | "text" | "json" | "markdown";
+  jsonFormat?: "table" | "token";
+  mode?: string;
```
thecodrr, line 12: We can do with a better name here. `mode` is used mainly for providing the language for the code block right? Perhaps we should name it `language` or something.

packages/importer/src/providers/standardnotes/index.ts
```diff
@@ -0,0 +1,248 @@
...
+          )}`;
+          return {
+            data: html,
+            type: ContentType.HTML,
+          };
+        } else {
+          let spreadsheet = <SpreadSheet>JSON.parse(data);
+          let html = ``;
+          for (let sheet of spreadsheet.sheets) {
+            if (sheet.rows.length > 0) {
```
thecodrr, line 203: Let's use inverted condition to reduce indentation.

packages/importer/src/providers/standardnotes/index.ts
```diff
@@ -0,0 +1,248 @@
...
+                  let cellAtCol = rowAtIndex.cells.find(
+                    (cell) => cell.index === col
+                  );
+                  // create an empty cell to fill index
+                  if (!cellAtCol) cellAtCol = { value: "", index: col };
+                  cells.push(`<td>${cellAtCol.value || ""}</td>`);
+                }
+                rows.push(`<tr>
+                ${cells.join("")}
+                </tr>`);
```
thecodrr, line 229: It would be best if we refactor the html construction out of the logic.

packages/importer/src/providers/standardnotes/index.ts
```diff
@@ -0,0 +1,248 @@
...
+
+          let html = `
+          ${tokens.map((token) =>
+            this.buildTableWithRows(
+              Object.keys(token)
+                .map(
+                  (key) => `<tr>
+              <th style="width:130px" >${key}</th>
+              <td>${token[key] || ""}</td>
+            </tr>`
```
thecodrr, line 190: We should refactor the html construction out of the main logic ‒ one mistake here and it'll break the whole note.

packages/importer/src/providers/standardnotes/index.ts
```diff
@@ -0,0 +1,248 @@
...
+          data: converter.makeHtml(data),
+          type: ContentType.HTML,
+        };
+      case "code":
+        let language = editorType.mode || "plaintext";
+        if (language === "htmlmixed") language = "html";
+        let code = hljs.highlightAuto(data, [language]);
+        let html = `<pre class="hljs language-${language}">
+          ${code.value}
+        </pre>`;
```
thecodrr, line 173: We should refactor the html construction out of the main logic ‒ one mistake here and it'll break the whole note.

thecodrr (changes requested):

thecodrr:
All looks good. Can we resolve the conflicts and regenerate the test snapshot?

PR #311, opened 2022-01-26
Add `vericrypt` app

thecodrr:
Still work in progress but the basic prototype is ready. ~~There is no actual decryption happening yet due to the pending work of open sourcing `nncrypto`.~~ The `crypto` module has been open-sourced in #329.

### What is Vericrypt?

Vericrypt is an open source utility to help you verify that Notesnook really is encrypting your data. It does this by helping you extract & decrypt raw encrypted data so you can see for yourself how encryption works.

### How does it work?

Vericrypt provides the necessary user interface to help anyone verify the encryption claims in 4 steps. The user is in complete control of the process and does all the data extraction & insertion manually. There are no external tools involved, no network requests required, and no bloatware. The end result is a decrypted blob of your data which you can go through and check easily.

### How can I be sure it does what it says?

Since Vericrypt is open source, you or anyone will be able to verify that the buttons & actions are actually doing what they say they are. Aside from that, you'll be extracting raw data right from the source (i.e. database, network requests etc.) so there will be 100% surety that Notesnook is in no way interfering with the process.


matttheba:
Thank you for doing this!

PR #861, opened 2022-09-03
feat: download app icon

mohsinalimat:
Show download app icon in readme file.


thecodrr:
Hey, I don't think there is a need for this on the main readme. The mobile app has it's own README which does include the download links. Further more the main README has a link to the downloads page at the very top.

PR #902, opened 2022-09-06
editor: insert blocks with paragraph to prevent dead end

### LINKED ISSUE #779: Blank line below an object
This is not a bug but it is annoying.

When you insert an object in a note (Table, Code, Quote, Math, etc) you can not write below it, unless you have put blank lines before insert it. 

I mean, you need to create the note or open one, insert (in example) 3 blank lines (enters), go back to line 1 or 2, insert the object you want and THEN continue writing in line 3. 
If you don't do this, you can not insert new lines below the object. 

It is easy to fix if you remember to insert the lines before the object, but if not, you need to cut the object, insert the lines and then paste the object again.

Again, not a bug, just annoying.

Thanks!

**Device information:**
App version: 2.1.4-889a8d98-web
OS: Windows 10 64-bit
Browser: Firefox 103.0
Pro: true

### LINKED ISSUE #880: Math Block Dead End
### What happened?

* In the mobile app
* adding a math Block at the end of the note 
* pretty much makes nothing else addable.

I got stuck trying to add a text below math Block and wad not able to do that. Because there is no way to get below ob the math Block. (In mobile app)
In the web version, ctrl+enter takes the cursor below the block.

I think automatically adding one or two blank lines below the math Block, at the time of creating a math equation, should work fine... As we can tap there to go there.

### Version

V2.1.8

### Platform/OS

Android

### Relevant log output

_No response_

ammarahm-ed:
Blocks like math/codeblock/blockquote should be inserted with a paragraph at the end. This handles the following cases:

1. When block is inserted at the start of a note, a paragraph will be added at the end
2. When block is inserted between two paragraphs, paragraph will not be added
3. When block is  inserted at the end of note, paragraph will be added.
4. If selection size is greater than 0, the current line will become part of the block

Fixes #880 
Fixes #779

### DIFF
--- a/apps/mobile/package-lock.json
+++ b/apps/mobile/package-lock.json
@@ -11,9 +11,6 @@
         "native/",
         "app/"
       ],
-      "dependencies": {
-        "qclone": "^1.1.0"
-      },
       "devDependencies": {
         "patch-package": "^6.4.7",
         "typescript": "^4.8.2"
@@ -614,6 +611,7 @@
       "version": "7.18.6",
       "resolved": "https://registry.npmjs.org/@babel/plugin-bugfix-safari-id-destructuring-collision-in-function-expression/-/plugin-bugfix-safari-id-destructuring-collision-in-function-expression-7.18.6.tgz",
       "integrity": "sha512-Dgxsyg54Fx1d4Nge8UnvTrED63vrwOdPmyvPzlNN/boaliRP54pm3pGzZD1SJUwrBA+Cs/xdG8kXX6Mn/RfISQ==",
+      "dev": true,
       "dependencies": {
         "@babel/helper-plugin-utils": "^7.18.6"
       },
@@ -628,6 +626,7 @@
       "version": "7.18.9",
       "resolved": "https://registry.npmjs.org/@babel/plugin-bugfix-v8-spread-parameters-in-optional-chaining/-/plugin-bugfix-v8-spread-parameters-in-optional-chaining-7.18.9.tgz",
       "integrity": "sha512-AHrP9jadvH7qlOj6PINbgSuphjQUAK7AOT7DPjBo9EHoLhQTnnK5u45e1Hd4DbSQEO9nqPWtQ89r+XEOWFScKg==",
+      "dev": true,
       "dependencies": {
         "@babel/helper-plugin-utils": "^7.18.9",
         "@babel/helper-skip-transparent-expression-wrappers": "^7.18.9",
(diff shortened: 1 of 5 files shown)


packages/editor/src/index.ts
```diff
@@ -70,12 +70,20 @@ import { Paragraph } from "./extensions/paragraph";
...
 import { DateTime } from "./extensions/date-time";
+import { EditorView } from "prosemirror-view";
 
 const CoreExtensions = Object.entries(TiptapCoreExtensions)
   // we will implement our own customized clipboard serializer
   .filter(([name]) => name !== "ClipboardTextSerializer")
   .map(([, extension]) => extension);
 
+EditorView.prototype.updateState = function updateState(state) {
+  //@ts-ignore
```
tortutales, line 81: why are you ignoring TS?

apps/mobile/package-lock.json
```diff
@@ -11,9 +11,6 @@
         "native/",
```
tortutales, line 11: There shouldn't be changes here as you didn't add any change in original `package.json`file

packages/editor/src/toolbar/tools/utils.tsx
```diff
@@ -38,3 +40,132 @@ export function menuButtonToTool(
...
 }
+
+/**
+ * Inserts a paragraph at the end of selection.
+ */
+function insertParagraphAtTheEndOfSelection(editor: Editor, tr: Transaction) {
+  if (!editor.current) return;
+  const docSize = editor.current.state.doc.content.size;
+  let selectionEnd = tr.selection.to + 1;
+  selectionEnd = docSize < selectionEnd ? docSize - 1 : selectionEnd;
```
tortutales, line 51: This can be in the same like and as a `const`

ammarahm-ed: This will require writing `tr.selection.to + 1` twice and I want to avoid doing that.

apps/mobile/package-lock.json
ammarahm-ed: Agreed

packages/editor/src/index.ts
ammarahm-ed: This is temporary and related to internal api of ProseMirror for which typings are not available. Without adding this piece of code, the editor crashes during development of every refresh. I will probably add a comment here explaining the reason.

thecodrr:
This won't work with nodes inserted via markdown shortcuts & keyboard shortcuts. I think a better solution for this would be to create a ProseMirror extension that handles this at a lower level based on transactions. It'll be slightly more complex but it'll work for every case.

PR #1190, opened 2022-10-13
Allow exporting all notes as html, md or txt

thecodrr:
Bulk exporting PDF is not yet supported. Locked notes will also not be exported.


thecodrr:
@ammarahm-ed you'll need to add this on your end as well.

PR #1228, opened 2022-10-22
editor: prevent tables from freezing the editor

dylhack:


packages/editor/src/extensions/table/component.tsx
```diff
@@ -125,15 +126,19 @@ function TableRowToolbar(props: TableToolbarProps) {
...
-        location: "left",
-        target: currentRow,
-        align: "start",
-        xOffset: -5,
-        yOffset: -3
-      });
-      rowToolsRef.current.style.top = `${pos.top}px`;
-      rowToolsRef.current.style.left = `${pos.left}px`;
+      try {
+        const pos = getPosition(rowToolsRef.current, {
```
thecodrr, line 130: I think it might be a better idea to handle this at the `getPosition` level instead of wrapping this in a try-catch.

packages/editor/src/extensions/table/component.tsx
```diff
@@ -37,6 +37,7 @@ import { getToolDefinition } from "../../toolbar/tool-definitions";
 import { getPosition } from "../../utils/position";
 import { findSelectedDOMNode } from "../../toolbar/utils/prosemirror";
 import { DesktopOnly } from "../../components/responsive";
+import { logger } from "nx/src/utils/logger";
```
thecodrr, line 40: You can just use `console.log` instead of importing it out of `nx` if required.

thecodrr (changes requested):
Hey! Thanks for the PR. There are some changes needed before I can merge it into master.

packages/editor/src/extensions/table/component.tsx
thecodrr: We can add a check to handle the `getBoundingClientRect` issue.

Currently `getPosition` function expects an `HTMLElement`; instead we can change the type to `Node` which includes all sorts of nodes. This will force us to check and make sure the element is actually an `HTMLElement`.

packages/editor/src/extensions/table/component.tsx
dylhack: resolved by 3d59b6e8de9ff09374f26fd5f8cadce99eb50ca5

packages/editor/src/extensions/table/component.tsx
dylhack: resolved by 3075afa6ae6e8f061ae0f8601d1c915e14510895

packages/editor/src/utils/position.ts
```diff
@@ -46,9 +46,13 @@ export type PositionOptions = {
...
 };
 export function getPosition(
-  element: HTMLElement,
+  node: Node,
   options: PositionOptions
 ): { top: number; left: number } {
+  const element = node.parentElement;
+  if (element === null) {
+    throw new Error(`Unsupported node type ${node.nodeType}`);
+  }
```
thecodrr, line 55: Awesome! However, we aren't always looking for the parentElement or need that so the check should me just on the element. Or wait.

I think we can do a check like this:

1. Check if the `node` is an instance of `HTMLElement`
2. If it is, just use that.
3. If it isn't:
    1. We can either just return
    2. Or use `node.parentElement` (in cases where it's a TextNode).

thecodrr (changes requested):

packages/editor/src/utils/position.ts
dylhack: Sounds good, adding now.

dylhack: 6b32274fc93e621e31cc7ada2d9fd94b0a09649a

thecodrr (approved):

thecodrr:
Everything looks good! Merging asap!

PR #1266, opened 2022-11-01
Preview notes in trash

alihamuh:

### DIFF
--- a/apps/web/src/components/trash-item/index.js
+++ b/apps/web/src/components/trash-item/index.js
@@ -26,11 +26,16 @@ import TimeAgo from "../time-ago";
 import { pluralize, toTitleCase } from "../../utils/string";
 import { showUndoableToast } from "../../common/toasts";
 import { showToast } from "../../utils/toast";
+import { hashNavigate } from "../../navigation";
+import { useStore } from "../../stores/note-store";
 
 function TrashItem({ item, index, date }) {
+  const isOpened = useStore((store) => store.selectedNote === item.id);
+
   return (
     <ListItem
       selectable
+      isFocused={isOpened}
       item={item}
       title={item.title}
       body={item.headline || item.description}
@@ -45,6 +50,9 @@ function TrashItem({ item, index, date }) {
         </Flex>
       }
       menu={{ items: menuItems, extraData: { item } }}
+      onClick={() => {
+        hashNavigate(`/notes/${item.id}/edit`, { replace: true });
+      }}
     />
   );
 }
--- a/apps/web/src/components/trash-item/index.js
+++ b/apps/web/src/components/trash-item/index.js
@@ -51,7 +51,8 @@ function TrashItem({ item, index, date }) {
       }
       menu={{ items: menuItems, extraData: { item } }}
       onClick={() => {
-        hashNavigate(`/notes/${item.id}/edit`, { replace: true });
+        if (item.itemType === "note")
+          hashNavigate(`/notes/${item.id}/edit`, { replace: true });
       }}
     />
   );
--- a/apps/web/src/stores/editor-store.js
+++ b/apps/web/src/stores/editor-store.js
@@ -55,7 +55,8 @@ export const getDefaultSession = (sessionId = Date.now()) => {
     context: undefined,
     color: undefined,
     dateEdited: 0,
-    attachmentsLength: 0
+    attachmentsLength: 0,
+    isDeleted: false
   };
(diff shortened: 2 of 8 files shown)


apps/web/src/components/trash-item/index.js
```diff
@@ -45,6 +50,15 @@ function TrashItem({ item, index, date }) {
         </Flex>
       }
       menu={{ items: menuItems, extraData: { item } }}
+      onClick={() => {
+        if (item.conflicted) {
+          hashNavigate(`/notes/${item.id}/conflict`, { replace: true });
+        } else if (item.locked) {
```
thecodrr, line 56: This can also be removed since notes and unlocked before deletion.

apps/web/src/stores/editor-store.js
```diff
@@ -123,9 +123,11 @@ class EditorStore extends BaseStore {
       return;
     }
 
-    let note = db.notes.note(noteId);
+    let note = db.notes.note(noteId)
+      ? db.notes.note(noteId).data
+      : db.notes.trashed(noteId);
+
```
thecodrr, line 129:
```suggestion
    const note = db.notes.note(noteId)?.data || db.notes.trashed(noteId);
```

apps/web/src/stores/editor-store.js
```diff
@@ -145,6 +147,10 @@ class EditorStore extends BaseStore {
         attachmentsLength: db.attachments.ofNote(note.id, "all")?.length || 0
       };
     });
+    if (!db.notes.note(noteId))
```
thecodrr, line 150: This will be extremely fragile. It's better to create a constant `isDeleted` that holds the value. Trashed notes have a `dateDeleted` property & their type is also `trash` so that will be more reliable.

apps/web/src/stores/editor-store.js
```diff
@@ -145,6 +147,10 @@ class EditorStore extends BaseStore {
...
+      this.set((state) => {
```
thecodrr, line 151: There's no need to add a separate call here. You can just update the state in a single `this.set` call which is being made in the previous lines.

apps/web/src/stores/editor-store.js
```diff
@@ -145,6 +147,10 @@ class EditorStore extends BaseStore {
...
+        state.session.readonly = true;
```
thecodrr, line 152: This approach won't work with deleted notes. We want to make sure of a few things:

1. User should not be able to change anything about the note. The `readonly` property can be changed from properties.
2. We need to show a notice that clearly indicates that this is a deleted note. This notice should also have a restore button. Pressing this button will close the note & restore it from trash.
3. The only information the user should be able to see is the metadata of a note. They shouldn't be able to change any options, see the session history etc.

To that end, it might be a better idea to add a `isDeleted` property in the session which can then be used later on for all the purposes mentioned above.

thecodrr (changes requested):

apps/web/src/stores/editor-store.js
```diff
@@ -281,11 +287,11 @@ class EditorStore extends BaseStore {
...
-    this.set(
-      (state) =>
-        (state.arePropertiesVisible =
-          toggleState !== undefined ? toggleState : !state.arePropertiesVisible)
-    );
+    this.set((state) => {
+      if (!state.session.isDeleted)
+        state.arePropertiesVisible =
+          toggleState !== undefined ? toggleState : !state.arePropertiesVisible;
+    });
```
thecodrr, line 294:
```suggestion
    if (this.get().session.isDeleted) return;
```

thecodrr (approved):

apps/web/src/stores/editor-store.js
```diff
@@ -281,11 +287,7 @@ class EditorStore extends BaseStore {
   };
 
   toggleProperties = (toggleState) => {
-    this.set(
-      (state) =>
-        (state.arePropertiesVisible =
-          toggleState !== undefined ? toggleState : !state.arePropertiesVisible)
-    );
+    if (this.get().session.isDeleted) return;
```
thecodrr, line 290:
```suggestion
    if (this.get().session.isDeleted) return;
    this.set(
      (state) =>
        (state.arePropertiesVisible =
          toggleState !== undefined ? toggleState : !state.arePropertiesVisible)
    );
```

thecodrr:
This also needs to be added on your end. @ammarahm-ed

PR #1303, opened 2022-11-08
Fix issue where latest backups are not restored

alihamuh:

### DIFF
--- a/packages/core/database/backup.js
+++ b/packages/core/database/backup.js
@@ -200,8 +200,7 @@ export default class Backup {
       await this._migrator.migrate(
         this._db,
         collections,
-        (id, version, type) =>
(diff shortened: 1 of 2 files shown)


packages/core/database/backup.js
```diff
@@ -200,7 +200,8 @@ export default class Backup {
       await this._migrator.migrate(
         this._db,
         collections,
-        (id) => data[id],
+        (id, version, type) =>
+          version < 5.8 ? data[id] : data[`${id}_${type}`],
```
thecodrr, line 204: `version` is already available in this scope so there is no need to take it as an argument.

thecodrr (changes requested):

PR #1312, opened 2022-11-10
Add support for saving custom colors in editor color picker

alihamuh:

### DIFF
--- a/packages/editor/src/toolbar/popups/color-picker.tsx
+++ b/packages/editor/src/toolbar/popups/color-picker.tsx
@@ -26,6 +26,7 @@ import tinycolor from "tinycolor2";
 import { HexColorPicker } from "react-colorful";
 import { Button } from "../../components/button";
 import { debounce } from "../../utils/debounce";
+import { config } from "../../utils/config";
 export const DEFAULT_COLORS = [
   "#e91e63",
   "#9c27b0",
@@ -44,30 +45,37 @@ export const DEFAULT_COLORS = [
 ];
 
 type ColorPickerProps = {
-  colors?: string[];
   color?: string;
   onClear: () => void;
   expanded?: boolean;
   onChange: (color: string) => void;
   onClose?: () => void;
   title?: string;
+  onSave?: (color: string) => void;
+  cacheKey?: string;
+  colors?: string[];
+  onColorDelete?: (color: string) => void;
 };
 const PALETTE_SIZE = [35, 35, 25];
 export function ColorPicker(props: ColorPickerProps) {
   const {
-    colors = DEFAULT_COLORS,
     color,
     onClear,
     onChange,
     title,
     onClose,
-    expanded
+    expanded,
+    onSave,
+    colors,
+    onColorDelete
   } = props;
+  console.log("color-picker", colors);
   const ref = useRef<HTMLDivElement>(null);
   const [isPickerOpen, setIsPickerOpen] = useState(expanded || false);
   const [currentColor, setCurrentColor] = useState<string>(
-    tinycolor(color || colors[0]).toHexString()
+    tinycolor(color || colors?.[0]).toHexString()
   );
+  const [trashMode, setTrashMode] = useState(false);
   const tColor = tinycolor(currentColor);
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


packages/editor/src/toolbar/tools/colors.tsx
```diff
@@ -120,11 +148,11 @@ export function TextColor(props: ToolProps) {
...
       activeColor={editor.getAttributes("textStyle").color}
       title="Text color"
-      onColorChange={(color) =>
+      onColorChange={(color) => {
         color
           ? editor.current?.chain().focus().setColor(color).run()
-          : editor.current?.chain().focus().unsetColor().run()
-      }
+          : editor.current?.chain().focus().unsetColor().run();
+      }}
```
thecodrr, line 155: This change can be undo-ed.

packages/editor/src/toolbar/tools/colors.tsx
```diff
@@ -60,7 +76,9 @@ export function ColorTool(props: ColorToolProps) {
       }}
       onOpen={() => setIsOpen((s) => !s)}
       toggled={isOpen}
-      onClick={() => onColorChange(activeColor)}
+      onClick={() => {
+        onColorChange(activeColor);
+      }}
```
thecodrr, line 81: This change can be undo-ed.

packages/editor/src/toolbar/tools/colors.tsx
```diff
@@ -34,6 +34,23 @@ type ColorToolProps = ToolProps & {
...
+  "#00bcd4",
+  "#009688",
+  "#4caf50",
+  "#8bc34a",
+  "#cddc39",
+  "#ffeb3b",
+  "#ffc107",
+  "#f44336"
+];
+
```
thecodrr, line 53: Why duplicate this instead of importing it from `color-picker.tsx`?

packages/editor/src/toolbar/popups/color-picker.tsx
```diff
@@ -51,23 +52,30 @@ type ColorPickerProps = {
...
+  cacheKey?: string;
 };
 const PALETTE_SIZE = [35, 35, 25];
 export function ColorPicker(props: ColorPickerProps) {
   const {
-    colors = DEFAULT_COLORS,
+    colors = [
+      ...DEFAULT_COLORS,
+      ...(config.get<string[]>(props.cacheKey!, []) || [])
+    ],
```
thecodrr, line 64: `colors` is a prop which means that it can be used by each instance to specify it's own set of custom colors.

The current approach uses the `cacheKey` to get the custom colors on every render. Aside from being inefficient, it also makes it super hard to update the custom colors when a color is added/deleted.

Here's what we should do instead:
1. Take the custom colors as a prop
2. Create a new state for storing all the colors (default & custom).
3. Use a `useEffect` hook to `push` the custom colors into the state.

packages/editor/src/toolbar/popups/color-picker.tsx
```diff
@@ -130,30 +138,47 @@ export function ColorPicker(props: ColorPickerProps) {
...
+              }}
+            />
+            <Button
+              sx={{
+                width: "50%",
+                flexShrink: 0,
+                p: 0,
+                fontSize: ["title", "title", "body"],
+                fontWeight: 800,
+                color: `${currentColor}`,
```
thecodrr, line 173: Using `currentColor` as the foreground is undesired since it'll become unreadable for low contrast colors (e.g. white/yellow). We should use a fixed color instead.

packages/editor/src/toolbar/tools/colors.tsx
```diff
@@ -82,6 +100,16 @@ export function ColorTool(props: ColorToolProps) {
               onColorChange();
               config.set(cacheKey, null);
             }}
+            onSave={() => {
+              if (!activeColor) return;
+              const colors = config.get<string[]>(`custom_${cacheKey}`, []);
+              if (colors?.includes(config.get(cacheKey)!)) return;
+              if (DEFAULT_COLORS.includes(config.get(cacheKey)!)) return;
+
+              colors?.push(config.get(cacheKey)!);
```
thecodrr, line 109: Instead of getting the current color from `config` why not take it as a parameter in the `onSave` function?

packages/editor/src/toolbar/popups/color-picker.tsx
```diff
@@ -220,6 +268,17 @@ export function ColorPicker(props: ColorPickerProps) {
                 }
               }}
               onClick={() => {
+                if (trashMode) {
```
thecodrr, line 271: Deleting custom colors should be done on the same level as saving custom colors. How do we know for certain that custom colors are always stored in `config`? We don't know what the `onSave` function did. Hence, it is better to leave it up to the upper layer to decide what to do.

packages/editor/src/toolbar/popups/color-picker.tsx
```diff
@@ -220,6 +268,17 @@ export function ColorPicker(props: ColorPickerProps) {
...
+                  if (colors.indexOf(color) > 13) {
```
thecodrr, line 272: What does `> 13` mean? If this is used to detect whether the color is a default one or not...how about just checking the `DEFAULT_COLORS` instead?

packages/editor/src/toolbar/popups/color-picker.tsx
```diff
@@ -220,6 +268,17 @@ export function ColorPicker(props: ColorPickerProps) {
...
+                    const customColors = config.get<string[]>(
+                      props.cacheKey!,
+                      []
+                    );
+                    customColors?.splice(colors.indexOf(color) - 14, 1);
```
thecodrr, line 277: What is `14`?

packages/editor/src/toolbar/popups/color-picker.tsx
```diff
@@ -204,6 +229,29 @@ export function ColorPicker(props: ColorPickerProps) {
               size={18}
             />
           </Button>
+          <Button
+            variant={"icon"}
+            sx={{
+              flexShrink: 0,
+              width: PALETTE_SIZE,
+              height: PALETTE_SIZE,
+              bg: trashMode ? "#ff0000" : "transparent",
```
thecodrr, line 238: Instead of `red` use a less angry color like `bgSecondary` or `fontTertiary`.

thecodrr (changes requested):
Kindly go through the requested changes one by one.

PR #1320, opened 2022-11-11
Download attachments on real-time sync

thecodrr:


thecodrr:
@ammarahm-ed you'll need to fix this on your end too.

PR #1471, opened 2022-12-08
Add support for moving topics between notebooks

alihamuh:

### DIFF
--- a/apps/web/src/components/dialogs/move-topics-dialog.tsx
+++ b/apps/web/src/components/dialogs/move-topics-dialog.tsx
@@ -17,21 +17,16 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { ChangeEvent, useCallback, useEffect, useRef, useState } from "react";
-import { Button, Flex, Text } from "@theme-ui/components";
+import { useCallback, useState } from "react";
+import { Flex, Text } from "@theme-ui/components";
 import * as Icon from "../icons";
 import { db } from "../../common/db";
 import Dialog from "./dialog";
-import Field from "../field";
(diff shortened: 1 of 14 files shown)


packages/core/collections/notebooks.js
```diff
@@ -155,4 +155,28 @@ export default class Notebooks extends Collection {
...
+    let noteIds = [];
+
+    for (let index = 0; index < items.length; index++) {
+      topicIds.push(items[index].id);
+      items[index].notebookId = id;
+      await this._db.notes.addToNotebook(
+        { id, topic: items[index].id },
+        ...this._db.notes.topicReferences.get(items[index].id)
+      );
+      noteIds.push(...this._db.notes.topicReferences.get(items[index].id));
```
thecodrr, line 174: This can be greatly simplified. Right now it is a confusing mess.

1. use `for..of` loop
2. use variables to improve readability
3. No idea why `noteIds` is being used here?
4. Delete topic one-by-one instead of at the end. This will also take care of `lastNotebookId`.

apps/web/src/components/dialogs/move-topics-dialog.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+type FilteredListProps<T extends Item> = {
+  placeholders: { filter: string; empty: string };
+  items: () => T[];
+  filter: (items: T[], query: string) => T[];
+  onCreateNewItem: (title: string) => Promise<void>;
+  renderItem: (item: T, index: number, refresh: () => void) => JSX.Element;
+};
+
+function FilteredList<T extends Item>(props: FilteredListProps<T>) {
```
thecodrr, line 135: This should be moved to its own component instead of copying it all over the place.

thecodrr:
This is no longer relevant as we have added drag/drop support for moving notebooks in v3.

PR #1490, opened 2022-12-14
web: sorting tests for notes, notebooks, tags & topics

alihamuh:


apps/web/__e2e__/sorting-notebooks.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    orderBy: "ascendingOrder",
+    sortBy: "dateEdited"
+  });
+  expect(await isLengthCorrect(page)).toBeTruthy();
+  await notebooks.sort({
+    groupBy: "year",
+    orderBy: "descendingOrder",
+    sortBy: "dateEdited"
+  });
+  expect(await isLengthCorrect(page)).toBeTruthy();
```
thecodrr, line 198: Ideally tests should follow the AAA rule (Arrange, Act, Assert). Here's what you should do:

1. Move all possible Group By, Sort by & Order by options into one array each.
2. Use 3 nested loops to iterate over all possible sorting combinations.
3. Inside the nested loops should be 1 test that will follow the AAA pattern (create notes, do the sorting, ensure items are there).

apps/web/__e2e__/sorting-notebooks.test.ts
```diff
@@ -0,0 +1,205 @@
...
+});
+
+async function isLengthCorrect(page: Page) {
+  let itemCount = await page.locator(getTestId("list-item")).count();
+  if (itemCount !== 5) return false;
+  return true;
+}
```
thecodrr, line 205: This is not the right place for this. If you just want the item count, move the method to `BaseViewModel` & name it appropriately.

thecodrr (changes requested):
This PR requires some other file-level refactoring:

1. You don't need a separate file for sorting of each item type (e.g. `sorting-notebooks.test.ts`). In fact, why separate them out at all? The notes related tests should be `notes.test.ts`, notebooks related tests should go in `notebooks.test.ts` and so on.

PR #1498, opened 2022-12-15
editor: disable deleting completd tasks in read mode

alihamuh:


thecodrr (changes requested):
Instead of manipulating the styles & actions of both buttons in readonly mode, let's just hide them altogether.

PR #1542, opened 2022-12-31
Backup tests

alihamuh:


apps/web/__e2e__/models/settings-view.model.ts
```diff
@@ -59,4 +72,19 @@ export class SettingsViewModel {
   async isLoggedIn() {
     return await this.accountStatusContainer.isVisible();
   }
+
+  async createBackup(encryptBackups = false, password = "") {
```
thecodrr, line 76: no need to specify 2 parameters. Just use `password` to see if we should encrypt or not.

apps/web/__e2e__/models/settings-view.model.ts
```diff
@@ -59,4 +72,19 @@ export class SettingsViewModel {
...
+    await this.backupRestoreContainer.click();
+    if (encryptBackups) await this.encyptBackups.click();
+    await this.backupData.click();
+    if (encryptBackups) await fillPasswordDialog(this.page, password);
```
thecodrr, line 80: Password dialog needs to be filled regardless of whether encryption is on or not.

apps/web/__e2e__/models/settings-view.model.ts
```diff
@@ -59,4 +72,19 @@ export class SettingsViewModel {
...
+    return await downloadAndReadFile(this.page, this.backupData, "utf-8");
+  }
+
+  async restoreData(filename: string, encrypted = false) {
+    await this.backupRestoreContainer.click();
+    await this.restoreBackup.click();
+    await uploadAndReadFile(this.page, this.restoreBackup, filename);
+    if (encrypted) await fillPasswordDialog(this.page, USER.CURRENT.password!);
```
thecodrr, line 88: Don't use the password directly. Take is as a parameter instead.

apps/web/__e2e__/utils/index.ts
```diff
@@ -104,6 +104,19 @@ async function downloadAndReadFile(
   return fs.readFileSync(path, { encoding });
 }
 
+async function uploadAndReadFile(
```
thecodrr, line 107: You are not reading the file after upload. Avoid misleading names.

apps/web/src/components/group-header/index.js
```diff
@@ -247,6 +247,7 @@ function GroupHeader(props) {
       {index === 0 && (
         <Flex mr={1}>
           <IconButton
+            testId={"sort-icon-button"}
```
thecodrr, line 250: Why are we making changes here?

apps/web/src/components/group-header/index.js
```diff
@@ -292,6 +293,7 @@ function IconButton(props) {
   const isMobile = useMobile();
   return (
     <Button
+      data-test-id={props.testId}
```
thecodrr, line 296: Why are we making changes here?

thecodrr (changes requested):

apps/web/__e2e__/models/settings-view.model.ts
```diff
@@ -59,4 +72,19 @@ export class SettingsViewModel {
   async isLoggedIn() {
     return await this.accountStatusContainer.isVisible();
   }
+
+  async createBackup(password: string | undefined = undefined) {
```
thecodrr, line 76:
```suggestion
  async createBackup(password?: string) {
```

PR #1636, opened 2023-01-11
editor: added fix for closing prompt on list selection

alihamuh:


packages/editor/src/toolbar/tools/lists.tsx
```diff
@@ -59,7 +59,9 @@ function _ListTool<TListStyleTypes extends string>(
       onClick={onClick}
       toggled={isOpen}
       sx={{ mr: 0 }}
-      onOpen={() => setIsOpen((s) => !s)}
+      onOpen={() => {
+        setIsOpen((s) => !s);
+      }}
```
thecodrr, line 64: No need for this change.

thecodrr (changes requested):

PR #1655, opened 2023-01-14
Add delete key shortcut for deleting notebooks

alihamuh:


thecodrr:
We should add this shortcut for everything that can be deleted.

alihamuh:
okay

PR #1721, opened 2023-01-18
Add option to change default font size & font family in editor

alihamuh:


thecodrr (changes requested):
Both the settings are not stateful i.e. changing their values will require restarting the app.

alihamuh:
okay

PR #1738, opened 2023-01-20
editor: added text direction support for all lists

alihamuh:


Yolakalemowa:
is this resolved? sorry noob here and wanted to report the same RTL misalignment issue.

PR #1883, opened 2023-02-07
Better separation between notes in the list

gitstart:
Added a border for better separation between notes in the Lists, Notebooks, Topics, and Tags.

Fixes [#474](https://github.com/streetwriters/notesnook/issues/474)
### Dark mode:
image

### Light mode:
image

### DIFF
--- a/apps/web/src/components/list-item/index.js
+++ b/apps/web/src/components/list-item/index.js
@@ -17,7 +17,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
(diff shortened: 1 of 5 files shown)


thecodrr:
Hey, @gitstart thanks for the PR.

I am not 100% sure this is necessary as the spacing between items is more than enough to separate them. Having a bottom border makes the whole list look too boxy.

_Sidenote: before you start working on a PR, it would be better that you discuss the issue you are trying to fix and get a confirmation. That will avoid any wasted time on your end. This is also mentioned in the [contributing guidelines](https://github.com/streetwriters/notesnook/blob/master/CONTRIBUTING.md)._

PR #1940, opened 2023-02-16
Add setting to control trash clear interval

alihamuh:


thecodrr:
@ammarahm-ed left on your end.

PR #2057, opened 2023-03-04
Search autocomplete text color

Abdulrehman-Jafer:
Fixed Search bar autocomplete text color

### DIFF
--- a/apps/web/src/components/search/index.js
+++ b/apps/web/src/components/search/index.js
@@ -21,9 +21,11 @@ import * as Icon from "../icons";
 import "./search.css";
 import Field from "../field";
 import { useStore } from "../../stores/theme-store";
+import { getDarkScheme } from "@notesnook/theme/dist/theme/colorscheme/dark";
 
 function SearchBox(props) {
   const { theme } = useStore();
+  const {text} = getDarkScheme()
   return (
     <Field
       data-test-id="search-input"
@@ -37,7 +39,7 @@ function SearchBox(props) {
         mt: 1,
         ...(theme === "dark" && {
           "input:-webkit-autofill": {
-            WebkitTextFillColor: "rgba(231, 225, 225, 0.842)",
+            WebkitTextFillColor: text,
             caretColor: "white",
             fontSize: "inherit"
           }
--- a/apps/web/src/components/search/index.js
+++ b/apps/web/src/components/search/index.js
@@ -24,13 +24,6 @@ import { useStore } from "../../stores/theme-store";
 
 function SearchBox(props) {
   const { theme } = useStore();
-  const autoCompleteTextStyle = {
-    "input:-webkit-autofill": {
-      WebkitTextFillColor: "rgba(231, 225, 225, 0.842) !important",
-      caretColor: "white",
-      fontSize: "inherit !important"
-    }
-  };
   return (
     <Field
       data-test-id="search-input"
@@ -38,7 +31,18 @@ function SearchBox(props) {
       id="search"
       name="search"
       type="text"
-      sx={{ m: 0, mx: 1, mt: 1, ...(theme === "dark" && autoCompleteTextStyle) }}
+      sx={{
+        m: 0,
+        mx: 1,
+        mt: 1,
+        ...(theme === "dark" && {
+          "input:-webkit-autofill": {
+            WebkitTextFillColor: "rgba(231, 225, 225, 0.842)",
+            caretColor: "white",
+            fontSize: "inherit"
+          }
+        })
+      }}
       placeholder="Type your query here"
       onKeyDown={(e) => {
         if (e.key === "Enter") props.onSearch(e.target.value);
--- a/apps/web/src/components/search/index.js
+++ b/apps/web/src/components/search/index.js
@@ -16,21 +16,21 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
+
 import * as Icon from "../icons";
 import "./search.css";
 import Field from "../field";
-import { useEffect } from "react";
 import { useStore } from "../../stores/theme-store";
+
 function SearchBox(props) {
   const { theme } = useStore();
-  useEffect(() => {
-    const search = document.getElementById("search");
-    if (theme === "dark") {
-      search.classList.add("darkModeSearch");
-    } else {
-      search.classList.remove("darkModeSearch");
+  const autoCompleteTextStyle = {
+    "input:-webkit-autofill": {
+      WebkitTextFillColor: "rgba(231, 225, 225, 0.842) !important",
+      caretColor: "white",
+      fontSize: "inherit !important"
     }
-  }, [theme]);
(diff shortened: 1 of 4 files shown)


apps/web/src/components/search/index.js
```diff
@@ -16,12 +16,21 @@ GNU General Public License for more details.
...
 function SearchBox(props) {
+  const { theme } = useStore();
+  useEffect(() => {
+    const search = document.getElementById("search");
+    if (theme === "dark") {
+      search.classList.add("darkModeSearch");
+    } else {
+      search.classList.remove("darkModeSearch");
+    }
+  }, [theme]);
```
thecodrr, line 33: There's no need to go to all this trouble. Just do inline styling via the sx prop. More details here: https://theme-ui.com/sx-prop

apps/web/src/components/search/search.css
```diff
@@ -1,3 +1,8 @@
 #searchInput:focus + #searchIcon {
   color: var(--primary);
 }
+.darkModeSearch:-webkit-autofill{
+  -webkit-text-fill-color: rgba(231, 225, 225, 0.842) !important;
```
thecodrr, line 5: What is this color?

thecodrr (changes requested):

apps/web/src/components/search/search.css
Abdulrehman-Jafer: It matches the original search box color

Abdulrehman-Jafer: rgba(231, 225, 225, 0.842)

thecodrr: According to https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color this isn't needed. The browser will automatically fall back to using the `color` property.

apps/web/src/components/search/index.js
```diff
@@ -20,16 +20,25 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 
 function SearchBox(props) {
+  const { theme } = useStore();
+  const autoCompleteTextStyle = {
+    "input:-webkit-autofill": {
+      WebkitTextFillColor: "rgba(231, 225, 225, 0.842) !important",
+      caretColor: "white",
+      fontSize: "inherit !important"
+    }
+  };
```
thecodrr, line 33: Either move this to a global constant or inline it as well.

apps/web/src/components/search/search.css
Abdulrehman-Jafer: Without using the webkitFillColor the text colors are not changing..

apps/web/src/components/search/index.js
Abdulrehman-Jafer: No It isn't necessary I added it to avoid any conflict we can remove it no problem

apps/web/src/components/search/index.js
```diff
@@ -20,16 +20,29 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+        m: 0,
+        mx: 1,
+        mt: 1,
+        ...(theme === "dark" && {
+          "input:-webkit-autofill": {
+            WebkitTextFillColor: "rgba(231, 225, 225, 0.842)",
+            caretColor: "white",
+            fontSize: "inherit"
+          }
+        })
```
thecodrr, line 44: Now let's get rid of applying this only for dark theme by using colors from the current theme.

You can see all theme colors here: https://github.com/streetwriters/notesnook/tree/master/packages/theme/src/theme/colorscheme

Abdulrehman-Jafer: OK

apps/web/src/components/search/index.js
```diff
@@ -20,16 +20,31 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       type="text"
-      sx={{ m: 0, mx: 1, mt: 1 }}
+      sx={{
+        m: 0,
+        mx: 1,
+        mt: 1,
+        ...(theme === "dark" && {
+          "input:-webkit-autofill": {
+            WebkitTextFillColor: text,
+            caretColor: "white",
```
thecodrr, line 43: Instead of `white` use `text` here as well.

apps/web/src/components/search/index.js
```diff
@@ -20,16 +20,31 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       autoFocus
       id="search"
       name="search"
       type="text"
-      sx={{ m: 0, mx: 1, mt: 1 }}
+      sx={{
+        m: 0,
+        mx: 1,
+        mt: 1,
+        ...(theme === "dark" && {
```
thecodrr, line 40: This is no longer necessary since we are using colors from the current theme.

thecodrr (changes requested):

apps/web/src/components/search/index.js
Abdulrehman-Jafer: Ok

apps/web/src/components/search/index.js
Abdulrehman-Jafer: ok

apps/web/src/components/search/index.js
```diff
@@ -20,16 +20,29 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import * as Icon from "../icons";
 import "./search.css";
 import Field from "../field";
+import { useStore } from "../../stores/theme-store";
+import { getColors } from "@notesnook/theme/dist/theme/colorscheme";
 
 function SearchBox(props) {
+  const { theme } = useStore();
+  const { text } = getColors(theme);
```
thecodrr, line 28: You don't need to use `getColors` to get theme colors. We are using `theme-ui` internally so stuff like colors, dimensions, margins etc are automatically resolved by it.

apps/web/src/components/search/index.js
```diff
@@ -20,16 +20,29 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
       name="search"
       type="text"
-      sx={{ m: 0, mx: 1, mt: 1 }}
+      sx={{
+        m: 0,
+        mx: 1,
+        mt: 1,
+        "input:-webkit-autofill": {
+          WebkitTextFillColor: text,
+          caretColor: text,
```
thecodrr, line 42:
```suggestion
          WebkitTextFillColor: "text",
          caretColor: "text",
```

thecodrr (changes requested):

apps/web/src/components/search/index.js
```diff
@@ -22,14 +22,24 @@ import "./search.css";
 import Field from "../field";
 
 function SearchBox(props) {
+
```
thecodrr, line 25:
```suggestion
```

thecodrr (approved):

thecodrr:
@Abdulrehman-Jafer now that I think about it, this change should be applied to every input instead of just on the search input. Let's move this change to the `Field` component.

thecodrr (approved):

PR #2123, opened 2023-03-17
feat: image preview & download

ammarahm-ed:
This adds support for basic attachment preview support. Currently only image previews are supported on mobile.

- Fixed image downloading not working on mobile
- Added image preview support on mobile with support for loading full quality images in full screen.
- Added Attachment preview logic in editor, so in future we can support preview for other files such as audio & video

### DIFF
--- a/apps/mobile/app/common/filesystem/download-attachment.js
+++ b/apps/mobile/app/common/filesystem/download-attachment.js
@@ -29,19 +29,29 @@ import { db } from "../database";
 import Storage from "../database/storage";
 import { cacheDir } from "./utils";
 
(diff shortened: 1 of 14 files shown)


packages/editor/src/toolbar/tools/image.tsx
thecodrr, line 1: These changes should be reverted.

packages/editor/src/extensions/image/image.ts
thecodrr, line 1: This change should be reverted.

thecodrr (changes requested):

Issue #2109, opened 2023-03-13
Title of the note in the browser


livejamie:
### What problem are you facing?

It's excellent context when using a note program in the browser to have the title of what you're working with displayed on the tab.

### What's the solution you'd like to see in Notesnook?

n/a

### What alternatives have you considered?

n/a

### Additional context

Here's the same note in Evernote, Supernotes and Notesnook showing the difference it makes.

![image](https://user-images.githubusercontent.com/49513/224619515-f21d4695-4ef6-4358-ab8a-2cab71cdc0ca.png)

thecodrr:
I agree! Will add this ASAP.

ammarahm-ed:
This has been added in #2127

PR #2127, opened 2023-03-17
Note title in browser Issue: Title of the note in the browser #2109

Abdulrehman-Jafer:


apps/web/src/utils/updateWebTitle.ts
thecodrr, line 1: A few issues:

1. There's no license header on this file.
2. Each file in the `utils/` folder can have multiple utilities so instead of naming it `updateWebTitle`, you should name it `document` or `dom` or whatever makes more sense.
3. I recommend installing prettier & eslint extensions in VSCode. If you can't do that, please run them manually on the code you change.

apps/web/src/utils/updateWebTitle.ts
```diff
@@ -0,0 +1,4 @@
+
+export const updateWebTitle = (title:string) => {
```
thecodrr, line 2: `updateWebTitle` is ambiguous & might be confusing. Can you change it to `setDocumentTitle`?

apps/web/src/utils/updateWebTitle.ts
```diff
@@ -0,0 +1,4 @@
...
+    document.title = `${title}-Notesnook`
```
thecodrr, line 3:
```suggestion
    document.title = `${title} - Notesnook`;
```

thecodrr (changes requested):
While this change works for newly opened notes, it doesn't set the document title when the note is saved. The result is that the document title won't be updated when a user changes the note's title.

Abdulrehman-Jafer (commented):
Check it out

thecodrr (approved):

PR #2144, opened 2023-03-20
converted color.js file into typescript

Abdulrehman-Jafer:


thecodrr:
@Abdulrehman-Jafer did you delete `color.js`?

PR #2153, opened 2023-03-20
web: migrated dimensions.js to typescript

Abdulrehman-Jafer:


thecodrr (changes requested):
You are not running prettier on these files. Have you enabled `Format on Save` in VSCode?

PR #2156, opened 2023-03-20
web: migrated filename.js to typescript

Abdulrehman-Jafer:


thecodrr (changes requested):
Please run prettier on this file.

thecodrr (approved):

PR #2161, opened 2023-03-21
web: migrated object.js into typescript

Abdulrehman-Jafer:


thecodrr (approved):
Really good work here with generics.

PR #2162, opened 2023-03-21
web: migrated parse.js into typescript

Abdulrehman-Jafer:


apps/web/src/utils/parse.ts
```diff
@@ -17,12 +17,13 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-export function tryParse(val) {
-  if (val === "undefined" || val === "null") return;
-
+export function tryParse(val: string | null | undefined) {
+  if (!val) {
```
thecodrr, line 21: Notice how we were comparing with strings (`"undefined"` & `"null"` instead of primitive types `undefined` & `null`) so keep those comparisons. This function is used to turn a string into a JSON object.

Secondly, the `val` parameter must not be `falsy`.

thecodrr (changes requested):
Check the above comment.

apps/web/src/utils/parse.ts
Abdulrehman-Jafer: Ok

apps/web/src/utils/parse.ts
```diff
@@ -17,12 +17,13 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-export function tryParse(val) {
-  if (val === "undefined" || val === "null") return;
-
+export function tryParse(val: string) {
+  if (!val || val == "null" || val == "undefined") {
```
thecodrr, line 21: No need for `!val` check.

Abdulrehman-Jafer: Ok

apps/web/src/utils/config.ts
```diff
@@ -27,7 +27,7 @@ function get<T>(key: string, def?: T): T {
   const value = window.localStorage.getItem(key);
   if (!value && def !== undefined) return def;
 
-  return tryParse(value);
+  return tryParse(value!);
```
thecodrr, line 30: Can you check the `value` here instead and return `def` if it is undefined?

Abdulrehman-Jafer: OK

thecodrr (approved):

PR #2163, opened 2023-03-21
web: migrated platform.js into typescript and updated global window type

Abdulrehman-Jafer:


apps/web/src/global.d.ts
```diff
@@ -32,6 +32,7 @@ type DesktopIntegrationSettings = {
 };
 
 declare interface Window {
+  os: string | undefined;
```
thecodrr, line 35: this `os` property is basically the same as `process.platform` in Node.JS so we should use the typings from that.

Abdulrehman-Jafer: Ok

thecodrr (approved):

PR #2165, opened 2023-03-21
web: converted css.js into typescript

Abdulrehman-Jafer:


apps/web/src/utils/css.ts
```diff
@@ -17,14 +17,14 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-export function removeCss(id) {
-  var link = document.getElementById(id);
-  link.remove();
+export function removeCss(id: string) {
+  const link = document.getElementById(id);
+  link ? link.remove() : undefined;
```
thecodrr, line 22: A simple if statement will suffice here.

apps/web/src/utils/css.ts
```diff
@@ -34,32 +34,33 @@ export function injectCssSrc(id, src) {
...
   if (variableCss) {
     head.removeChild(variableCss);
   }
-  let css = document.createElement("style");
+  const css = document.createElement("style");
   css.type = "text/css";
   css.id = "variables";
   // Support for IE
-  if (css.styleSheet) css.styleSheet.cssText = rule;
+  if ((css as any).styleSheet) (css as any).styleSheet.cssText = rule;
```
thecodrr, line 47: We don't support IE so you can remove this.

thecodrr (changes requested):

apps/web/src/utils/css.ts
Abdulrehman-Jafer: Ok

apps/web/src/utils/css.ts
thecodrr: @Abdulrehman-Jafer

Abdulrehman-Jafer: Ok I'll remove the ternary..

thecodrr (approved):

PR #2166, opened 2023-03-21
converted page visibility.js into typescript

Abdulrehman-Jafer:
Used any type at line 41 please review that..

### DIFF
--- a/apps/web/src/utils/page-visibility.ts
+++ b/apps/web/src/utils/page-visibility.ts
@@ -18,23 +18,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 const getVisibilityChangeParams = () => {
-  let hidden: "hidden" | "msHidden" | "webkitHidden" = "hidden";
-  let visibilityChange:
-    | "visibilitychange"
-    | "msvisibilitychange"
-    | "webkitvisibilitychange" = "visibilitychange";
   if ("hidden" in document) {
     // Opera 12.10 and Firefox 18 and later support
-    hidden = "hidden";
-    visibilityChange = "visibilitychange";
+    return ["hidden", "visibilityChange"] as const;
   } else if ("msHidden" in document) {
-    hidden = "msHidden";
-    visibilityChange = "msvisibilitychange";
+    return ["msHidden", "msvisibilityChange"] as const;
   } else if ("webkitHidden" in document) {
-    hidden = "webkitHidden";
-    visibilityChange = "webkitvisibilitychange";
+    return ["webkitHidden", "webkitvisibilityChange"] as const;
   }
-  return { hidden, visibilityChange };
+  return ["hidden", "visibilityChange"] as const;
 };
 
 export function onPageVisibilityChanged(
--- a/apps/web/src/utils/page-visibility.ts
+++ b/apps/web/src/utils/page-visibility.ts
@@ -17,7 +17,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-const getVisibilityChange = () => {
+const getVisibilityChangeParams = () => {
   let hidden: "hidden" | "msHidden" | "webkitHidden" = "hidden";
   let visibilityChange:
     | "visibilitychange"
@@ -47,7 +47,7 @@ export function onPageVisibilityChanged(
   onDeviceOffline(() => handler("offline", false));
 
   // Handle page visibility change
-  const { hidden, visibilityChange } = getVisibilityChange();
+  const { hidden, visibilityChange } = getVisibilityChangeParams();
   document.addEventListener(visibilityChange, () =>
     handler("visibilitychange", (document as any)[hidden])
   );
--- a/apps/web/src/utils/page-visibility.ts
+++ b/apps/web/src/utils/page-visibility.ts
@@ -18,20 +18,20 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 let hidden: string, visibilityChange: string;
-if (typeof document.hidden !== "undefined") {
+if ("hidden" in document) {
   // Opera 12.10 and Firefox 18 and later support
   hidden = "hidden";
   visibilityChange = "visibilitychange";
-} else if (typeof (document as any).msHidden !== "undefined") {
+} else if ("msHidden" in document) {
   hidden = "msHidden";
   visibilityChange = "msvisibilitychange";
-} else if (typeof (document as any).webkitHidden !== "undefined") {
+} else if ("webkitHidden" in document) {
   hidden = "webkitHidden";
   visibilityChange = "webkitvisibilitychange";
 }
 
 export function onPageVisibilityChanged(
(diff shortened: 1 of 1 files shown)


apps/web/src/utils/page-visibility.ts
```diff
@@ -17,36 +17,38 @@ You should have received a copy of the GNU General Public License
...
 
-function onDeviceOnline(handler) {
+function onDeviceOnline(handler: () => void | undefined) {
   window.addEventListener("online", function () {
     handler && handler();
   });
 }
 
-function onDeviceOffline(handler) {
+function onDeviceOffline(handler: () => void | undefined) {
```
thecodrr, line 51: `handler` shouldn't be optional in both of these.

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,36 +17,38 @@ You should have received a copy of the GNU General Public License
...
   visibilityChange = "msvisibilitychange";
-} else if (typeof document.webkitHidden !== "undefined") {
+} else if ("webkitHidden" in document) {
   hidden = "webkitHidden";
   visibilityChange = "webkitvisibilitychange";
 }
 
-export function onPageVisibilityChanged(handler) {
+export function onPageVisibilityChanged(
+  handler: (status: string, bool: boolean) => void
```
thecodrr, line 34: Use TypeScript union for specific `status` types here. Currently we have `online`, `offline` & `visibilitychange`

thecodrr (changes requested):

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,36 +17,38 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-var hidden, visibilityChange;
-if (typeof document.hidden !== "undefined") {
+let hidden: string, visibilityChange: string;
```
thecodrr, line 20: Use strict typing here. Actually turn this into a function that returns `hidden` & `visibilityChange` with strict types.

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,36 +17,38 @@ You should have received a copy of the GNU General Public License
...
+export function onPageVisibilityChanged(
+  handler: (status: string, bool: boolean) => void
+) {
   onDeviceOnline(() => handler("online", false));
   onDeviceOffline(() => handler("offline", false));
 
   // Handle page visibility change
   document.addEventListener(visibilityChange, () =>
-    handler("visibilitychange", document[hidden])
+    handler("visibilitychange", (document as any)[hidden])
```
thecodrr, line 41: Removing this `any` will take multiple steps. For now keep it here.

thecodrr (changes requested):

apps/web/src/utils/page-visibility.ts
Abdulrehman-Jafer: I adjusted according to line 47

apps/web/src/utils/page-visibility.ts
Abdulrehman-Jafer: Ok

apps/web/src/utils/page-visibility.ts
Abdulrehman-Jafer: Ok

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,37 +17,50 @@ You should have received a copy of the GNU General Public License
...
-} else if (typeof document.webkitHidden !== "undefined") {
-  hidden = "webkitHidden";
-  visibilityChange = "webkitvisibilitychange";
-}
+const getVisibilityChangeParams = () => {
+  let hidden: "hidden" | "msHidden" | "webkitHidden" = "hidden";
+  let visibilityChange:
+    | "visibilitychange"
+    | "msvisibilitychange"
+    | "webkitvisibilitychange" = "visibilitychange";
```
thecodrr, line 25: No need for statically typing these. It'll be automatically inferred by Typescript. That's why we are using a function.

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,37 +17,50 @@ You should have received a copy of the GNU General Public License
...
+  if ("hidden" in document) {
+    // Opera 12.10 and Firefox 18 and later support
+    hidden = "hidden";
+    visibilityChange = "visibilitychange";
```
thecodrr, line 29: Do a direct return instead of assignment.

thecodrr (changes requested):

apps/web/src/utils/page-visibility.ts
Abdulrehman-Jafer: Ok

apps/web/src/utils/page-visibility.ts
Abdulrehman-Jafer: Ok

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,37 +17,42 @@ You should have received a copy of the GNU General Public License
...
-} else if (typeof document.webkitHidden !== "undefined") {
-  hidden = "webkitHidden";
-  visibilityChange = "webkitvisibilitychange";
-}
-
-export function onPageVisibilityChanged(handler) {
+const getVisibilityChangeParams = () => {
+  if ("hidden" in document) {
+    // Opera 12.10 and Firefox 18 and later support
+    return ["hidden", "visibilityChange"] as const;
```
thecodrr, line 23: You can remove this rule or move it to an `else` statement because it is the default.

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,37 +17,42 @@ You should have received a copy of the GNU General Public License
...
-} else if (typeof document.msHidden !== "undefined") {
-  hidden = "msHidden";
-  visibilityChange = "msvisibilitychange";
-} else if (typeof document.webkitHidden !== "undefined") {
-  hidden = "webkitHidden";
-  visibilityChange = "webkitvisibilitychange";
-}
-
-export function onPageVisibilityChanged(handler) {
+const getVisibilityChangeParams = () => {
```
thecodrr, line 20: Why are we creating an arrow function here? Just use the normal `function` statement.

apps/web/src/utils/page-visibility.ts
```diff
@@ -17,37 +17,42 @@ You should have received a copy of the GNU General Public License
...
+  handler: (
+    status: "online" | "offline" | "visibilitychange",
+    bool: boolean
+  ) => void
+) {
   onDeviceOnline(() => handler("online", false));
   onDeviceOffline(() => handler("offline", false));
 
   // Handle page visibility change
+  const { hidden, visibilityChange } = getVisibilityChangeParams();
```
thecodrr, line 42: You are returning an array with 2 elements. It's not an object. So use `[hidden, ...]` instead.

thecodrr (changes requested):

PR #2175, opened 2023-03-22
Fix RTL support for outline lists

alihamuh:


packages/editor/src/extensions/outline-list-item/outline-list-item.ts
```diff
@@ -140,17 +146,34 @@ export const OutlineListItem = Node.create<ListItemOptions>({
...
+
+        let xlowerLimit = clientX >= x - hitArea.width;
+        let xHigherLimit = clientX <= x;
+
+        if (parent) {
+          if (parent.node.attrs.textDirection === "rtl") {
+            xHigherLimit = clientX <= right + hitArea.width;
+            xlowerLimit = clientX >= right;
+          }
+        }
```
thecodrr, line 172: No need for these nested if conditions. You can merge both the conditions.

packages/editor/src/extensions/outline-list-item/outline-list-item.ts
```diff
@@ -140,17 +146,34 @@ export const OutlineListItem = Node.create<ListItemOptions>({
...
-
         const hitArea = { width: 26, height: 24 };
+
+        const selection = editor.state.selection;
+        const parent = findParentNode((node) => !!node.attrs.textDirection)(
+          selection
+        );
+
+        let xlowerLimit = clientX >= x - hitArea.width;
+        let xHigherLimit = clientX <= x;
```
thecodrr, line 165: 1. Can we rename these to `xStart` & `xEnd`?
2. It'd be better for readability to put the `y` coordinates here as well.

packages/editor/src/extensions/outline-list/outline-list.ts
```diff
@@ -98,9 +98,10 @@ export const OutlineList = Node.create<OutlineListOptions>({
   },
 
   addNodeView() {
-    return () => {
+    return ({ HTMLAttributes }) => {
       const ul = document.createElement("ul");
       ul.classList.add("outline-list");
+      if (HTMLAttributes.dir) ul.dir = HTMLAttributes.dir;
```
thecodrr, line 104: `HTMLAttributes` might not always contain `dir` so you should add a fallback here like `node.attrs.textDirection`.

packages/editor/styles/styles.css
```diff
@@ -520,6 +519,14 @@ p > *::selection {
   transition: transform 250ms ease;
 }
 
+*:not([dir="rtl"]).outline-list li > p::before {
+  left: -22px;
+}
```
thecodrr, line 524: No need for this. You can reset the `left` property in the `rtl` CSS rule below.

packages/editor/styles/styles.css
```diff
@@ -530,30 +537,52 @@ p > *::selection {
...
+
 .outline-list li .outline-list::before {
   content: " ";
   position: absolute;
   height: 100%;
+  transition: border 200ms ease-in-out;
+}
+
+.outline-list:not([dir="rtl"]) li .outline-list::before {
```
thecodrr, line 551: These `:not` css rules are not required. You can move the changed styles to the main CSS rule for the element.

The `[dir="rtl"]` should always be below the non-rtl rules. CSS Rules lower in the file take precedence.

packages/editor/styles/styles.css
```diff
@@ -530,30 +537,52 @@ p > *::selection {
...
-  transition: border 200ms ease-in-out;
 }
 
-.outline-list li .outline-list:hover:before {
+[dir="rtl"].outline-list li .outline-list::before {
+  right: -14px;
+  border-right: 1px solid var(--border);
+}
+
+*:not([dir="rtl"]).outline-list li .outline-list:hover:before {
```
thecodrr, line 561: These `:not` css rules are not required. You can move the changed styles to the main CSS rule for the element.

The `[dir="rtl"]` should always be below the non-rtl rules. CSS Rules lower in the file take precedence.

thecodrr (changes requested):

thecodrr:
@alihamuh good work! There are a few changes required though.

PR #2193, opened 2023-03-24
web: added condition to handle "none" key in disabled tools

alihamuh:


apps/web/src/components/dialogs/toolbar-config-dialog.tsx
```diff
@@ -714,8 +714,9 @@ export function getDisabledTools(tools: TreeNode[]) {
   for (const key in allTools) {
     const tool = allTools[key as ToolId];
     if (tool.conditional) continue;
-    if (items.findIndex((t) => t.toolId === key) <= -1)
-      disabled.push(key as ToolId);
+
+    const isToolTrashed = items.findIndex((t) => t.toolId === key);
+    if (isToolTrashed <= -1 && key !== "none") disabled.push(key as ToolId);
```
thecodrr, line 719:
```suggestion
    const isDisabled = items.findIndex((t) => t.toolId === key);
    if (isDisabled === -1 && key !== "none") disabled.push(key as ToolId);
```

PR #2205, opened 2023-03-26
editor: improved use-editor.ts hook 

### LINKED ISSUE #2086: Changing editor settings requires app restart to take effect
Turning off double-spaced lines does not hold til next app start. Very frustrating!

**Device information:**
App version: 2.4.4-2c6f205-web
OS: Windows 10 64-bit
Browser: Microsoft Edge 110.0.1587.63
Pro: true

Abdulrehman-Jafer:
Issue: Changing editor settings requires app restart to take effect #2086
web:  Also added doubelSpacedLines in the dependencies of useTipTap  and also in the settings edited restart editor message to New settings are applied!

Closes #2086

### DIFF
--- a/apps/web/src/components/editor/tiptap.tsx
+++ b/apps/web/src/components/editor/tiptap.tsx
@@ -107,7 +107,7 @@ function TipTap(props: TipTapProps) {
     onDownloadAttachment,
     onAttachFile,
     onContentChange,
-    onFocus = () => {},
+    onFocus = () => { },
     content,
     toolbarContainerId,
     editorContainer,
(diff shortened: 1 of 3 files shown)


thecodrr:
@Abdulrehman-Jafer after some testing I realized that this behavior is web/desktop specific i.e. mobile doesn't have this issue so can we fix it on the web side instead?

Abdulrehman-Jafer:
Ok I'll look into it but this problem is also in mob

Abdulrehman-Jafer:
And secondly this new settings is not only for double space I mean it prevent any loss of data if we add anything in editor dependency to reset the editor what do you think about it?

thecodrr:
@Abdulrehman-Jafer I tested it on mobile and it is working without these changes. Just had to add `doubleSpacedLines` in the dependency array.

Let's move this to the web side of the app instead of in the editor.

PR #2215, opened 2023-03-28
core: fixed exports line spacings

Abdulrehman-Jafer:
Issue: Font size, line spacing and export #2111
Fixed export as Pdf, text and html.

### DIFF
--- a/packages/core/content-types/tiptap.js
+++ b/packages/core/content-types/tiptap.js
@@ -42,7 +42,15 @@ export class Tiptap {
   }
 
   toHTML() {
-    return this.data;
+    let data = this.data.replaceAll(
+      'data-spacing="single"',
+      'style = "margin:0px"'
+    );
+    data = data.replaceAll(
+      'data-spacing="double"',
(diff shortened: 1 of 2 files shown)


packages/core/content-types/tiptap.js
```diff
@@ -42,7 +42,15 @@ export class Tiptap {
...
-    return this.data;
+    let data = this.data.replaceAll(
+      'data-spacing="single"',
+      'style = "margin:0px"'
+    );
+    data = data.replaceAll(
+      'data-spacing="double"',
+      'style = "margin-bottom:0px"'
+    );
+    return data;
```
thecodrr, line 53: this will be unnecessary because you are already changing the html template.

thecodrr (changes requested):

packages/core/content-types/tiptap.js
Abdulrehman-Jafer: yes we can do that.

thecodrr:
Good work!

PR #2250, opened 2023-04-01
web: converted use-auto-updater hook into typescript

Abdulrehman-Jafer:


thecodrr:
This is no longer needed.

PR #2267, opened 2023-04-04
editor: fixed-code-highlights not applying

### LINKED ISSUE #2222: Syntax highlighting for code blocks?
### What happened?

You can select a type/language in the code blocks.

Assuming thats for syntax highlighting then its not working. Text stays the same in any case.

Add a code block, try different types.

### Version

2.4.9

### Platform/OS

Android

### Relevant log output

_No response_

Abdulrehman-Jafer:
Fixes #2222


thecodrr:
@Abdulrehman-Jafer editor tests are failing.

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -198,7 +205,20 @@ export function HighlighterPlugin({
...
+              const isRegistered = () => {
+                for (const key in registeredLanguage) {
+                  const currentLang = registeredLanguage[key];
+                  if (currentLang == language) {
+                    return true;
+                  }
+                }
+              };
+
+              if (!isRegistered()) {
```
thecodrr, line 219: Why not just use `refractor.registered`?

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -221,8 +241,6 @@ export function HighlighterPlugin({
                   decorations = decorations.remove(toRemove);
                 if (toAdd.length > 0)
                   decorations = decorations.add(tr.doc, toAdd);
-              } else {
-                languages[id] = language;
```
thecodrr, line 225: Why are you flipping the `if` statement? We don't do negative checks in the if statement if possible.

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -121,10 +121,17 @@ export function HighlighterPlugin({
...
+          const registeredLanguage = refractor.listLanguages();
           if (!pluginState) return;
 
           const changedBlocks: Set<string> = new Set();
           for (const blockKey in pluginState.languages) {
+            if (pluginLanguages) {
+              if (!registeredLanguage.includes(pluginLanguages[blockKey])) {
+                HIGHLIGHTED_BLOCKS.delete(blockKey);
+              }
+            }
```
thecodrr, line 134: Simplify this and merge it with the if statement below. You might need to move the `if..statement` a few downs lower.

thecodrr (changes requested):
This solution might work but needs fixes. Also please add tests for this.

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: Ok I will flip it back and it will make more sense.

Abdulrehman-Jafer:
I can pass the editor test by changing line 205 from  return "No need to register"; to return false this will pass the test but it will call the update method useless until there is no language selected for the block.

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -198,7 +196,16 @@ export function HighlighterPlugin({
               updated.add(block.pos);
 
               const { id, language } = block.node.attrs;
-              if (languages[id]) {
+
+              const isRegistered = () => {
```
thecodrr, line 200: No need for this function.

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -198,7 +196,16 @@ export function HighlighterPlugin({
...
+                if (language) {
+                  return refractor.registered(language) ? true : false;
+                }
+                return "No need to register";
+                // If I return false It will uselessly calls the update method until the language in null(plaintext).
+              };
+
+              if (isRegistered()) {
```
thecodrr, line 208: I am not sure this is correct.

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -124,24 +124,22 @@ export function HighlighterPlugin({
           if (!pluginState) return;
 
           const changedBlocks: Set<string> = new Set();
-          for (const blockKey in pluginState.languages) {
-            if (HIGHLIGHTED_BLOCKS.has(blockKey)) continue;
 
+          for (const blockKey in pluginState.languages) {
```
thecodrr, line 128: Why this change?

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -124,24 +124,22 @@ export function HighlighterPlugin({
...
             if (!languageDefinition) continue;
 
-            if (refractor.registered(language)) {
-              if (!HIGHLIGHTED_BLOCKS.has(blockKey)) {
-                HIGHLIGHTED_BLOCKS.add(blockKey);
-                changedBlocks.add(blockKey);
-              }
-              continue;
-            }
-
```
thecodrr, line 144: Why is this removed?

thecodrr (changes requested):
Not ready yet.

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: Ok

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: I did not change this line. Its the same.

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: This is unnecessary now as we are already checking if the language is registered in the top if statement.

packages/editor/src/extensions/code-block/highlighter.ts
thecodrr: Why is it showing in the diff?

packages/editor/src/extensions/code-block/highlighter.ts
thecodrr: What about everything inside that if block?

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: It is some kind of glitch probab it is not showing in the diff in my view.

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -121,18 +121,18 @@ export function HighlighterPlugin({
         },
         async update(view) {
           const pluginState = HIGHLIGHTER_PLUGIN_KEY.getState(view.state);
-          const pluginLanguages = pluginState?.languages;
-          const registeredLanguage = refractor.listLanguages();
           if (!pluginState) return;
 
           const changedBlocks: Set<string> = new Set();
+
           for (const blockKey in pluginState.languages) {
```
Abdulrehman-Jafer, line 128: same

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: OK.

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: I will add it back my bad It was a mistake.

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: I am sure sir.

Abdulrehman-Jafer:
Still returning "No need to register" which I think is enhancing performance. But if this not feel right to you I can remove it,

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -198,6 +203,15 @@ export function HighlighterPlugin({
               updated.add(block.pos);
 
               const { id, language } = block.node.attrs;
+
+              const isRegistered =
+                language ? refractor.registered(language) ? true : false : "No need to register"
```
thecodrr, line 208: You are still pointlessly doing this instead of just `refractor.registered`.

thecodrr (changes requested):

packages/editor/src/extensions/code-block/highlighter.ts
Abdulrehman-Jafer: OK I am removing it right now.

Abdulrehman-Jafer: Actually no I checked this check is necessary because when new code block is created language is null so that is why I am doing this check.

thecodrr: If language is null, `refractor.registered` will return false, no?

Abdulrehman-Jafer: It gives error.
type error

thecodrr: Then just do `language && refractor.registered(language)`

Abdulrehman-Jafer: Ok

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -198,6 +203,15 @@ export function HighlighterPlugin({
               updated.add(block.pos);
 
               const { id, language } = block.node.attrs;
+
+              const isRegistered =
+                language && refractor.registered(language) ? true : false;
+
+              if (!isRegistered) {
```
thecodrr, line 210:
```suggestion
              if (language && !refractor.registered(language)) {
```

thecodrr:
@Abdulrehman-Jafer you need to add tests for this.

packages/editor/src/extensions/code-block/tests/code-block.test.ts
```diff
@@ -191,3 +192,28 @@ test("editing code in a highlighted code block should not be too slow", async ()
...
+    defaultLanguage: "plaintext"
+  });
+  editor.commands.updateAttributes(CodeBlock.name, { language: "java" });
+  await new Promise((resolve) => setTimeout(resolve, 100));
+
+  const highlighterState = plugin.spec.state?.init(editor.state, editor.state);
+  const highlighterLanguages = highlighterState?.languages;
+  for (const key in highlighterLanguages) {
+    const isTrue = refractor.registered(highlighterLanguages[key]);
+    expect(isTrue).toBe(true);
```
thecodrr, line 217: This should be a single line or rename the `isTrue` to `isRegistered`.

packages/editor/src/extensions/code-block/tests/code-block.test.ts
```diff
@@ -191,3 +192,28 @@ test("editing code in a highlighted code block should not be too slow", async ()
...
+
+  const plugin = HighlighterPlugin({
+    name: CodeBlock.name,
+    defaultLanguage: "plaintext"
+  });
+  editor.commands.updateAttributes(CodeBlock.name, { language: "java" });
+  await new Promise((resolve) => setTimeout(resolve, 100));
+
+  const highlighterState = plugin.spec.state?.init(editor.state, editor.state);
+  const highlighterLanguages = highlighterState?.languages;
```
thecodrr, line 214: This is not necessary. You can just check that `java` or whatever the language you switched to is registered.

packages/editor/src/extensions/code-block/tests/code-block.test.ts
```diff
@@ -191,3 +192,28 @@ test("editing code in a highlighted code block should not be too slow", async ()
...
+    initialContent: CODEBLOCKS_HTML,
+    extensions: {
+      codeblock: CodeBlock
+    }
+  });
+
+  const plugin = HighlighterPlugin({
+    name: CodeBlock.name,
+    defaultLanguage: "plaintext"
+  });
```
thecodrr, line 209: Not needed in this test.

packages/editor/src/extensions/code-block/tests/code-block.test.ts
```diff
@@ -191,3 +192,28 @@ test("editing code in a highlighted code block should not be too slow", async ()
   expect(timings.reduce((a, b) => a + b) / timings.length).toBeLessThan(16);
   expect(editorElement.outerHTML).toMatchSnapshot();
 });
+
+test("Switching codeblock language should register the new language", async () => {
```
thecodrr, line 196: Are we just checking that language got registered? Shouldn't we also check that the code also got highlighted?

thecodrr (changes requested):

packages/editor/src/extensions/code-block/tests/code-block.test.ts
Abdulrehman-Jafer: OK

packages/editor/src/extensions/code-block/tests/code-block.test.ts
Abdulrehman-Jafer: The problem we solved here was the languages were not being registered which was the root cause of the code not being hghlighted so that is why I made the test to only check for new language registeration

packages/editor/src/extensions/code-block/tests/code-block.test.ts
Abdulrehman-Jafer: Ok understood

packages/editor/src/extensions/code-block/tests/code-block.test.ts
Abdulrehman-Jafer: Ok

thecodrr (approved):

Abdulrehman-Jafer (commented):
This will regenerate the other bug that I solved which was language was not applying when user create a new Block and select a language for the first time and user had to select the language for second time to select the language. After debugging I found out that the reason of it was the first time codeblock id was undefined.

thecodrr:
@Abdulrehman-Jafer 

> This will regenerate the other bug that I solved which was language was not applying when user create a new Block and select a language for the first time and user had to select the language for second time to select the language. After debugging I found out that the reason of it was the first time codeblock id was undefined.

That's why tests are important.

thecodrr (approved):

PR #2274, opened 2023-04-05
core: disable color-name-sanitization

### LINKED ISSUE #1823: Rename color text always getting set to all lowercase without spaces
### What happened?

When I try to right click on a specific color, purple in my example, and then click "Rename color" and type my desired name (for example, Official Notes) and click on Save. 

It ignores any formatting, becomes case insensitive, ignores spaces and gets saved as "officialnotes". I tested and it appears to accept special characters but not spaces.

1. Right click on any color from left pane
2. Click "Rename color"
3. Type any desired name (with a mix of uppercase, lowercase and spaces)
4. Click on Save
5. You will see the effect

### Version

v2.4.0

### Platform/OS

Windows

### Relevant log output

_No response_

Abdulrehman-Jafer:
Fixes #1823

### DIFF
--- a/apps/web/src/common/dialog-controller.tsx
+++ b/apps/web/src/common/dialog-controller.tsx
@@ -564,7 +564,7 @@ export function showRenameColorDialog(colorId: string) {
       onClose={() => perform(false)}
       onAction={async (title: string) => {
         if (!title) return;
-        await db.colors?.rename(colorId, title, true);
+        await db.colors?.rename(colorId, title);
         showToast("success", "Color renamed!");
         appStore.refreshNavItems();
         perform(true);
--- a/apps/web/src/common/dialog-controller.tsx
+++ b/apps/web/src/common/dialog-controller.tsx
@@ -209,9 +209,8 @@ export function showError(title: string, message: string) {
 export function showMultiDeleteConfirmation(length: number) {
   return confirm({
     title: `Delete ${length} items?`,
-    message: `These items will be **kept in your Trash for ${
-      db.settings?.getTrashCleanupInterval() || 7
-    } days** after which they will be permanently deleted.`,
+    message: `These items will be **kept in your Trash for ${db.settings?.getTrashCleanupInterval() || 7
+      } days** after which they will be permanently deleted.`,
     positiveButtonText: "Yes",
     negativeButtonText: "No"
   });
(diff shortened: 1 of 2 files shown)


packages/core/collections/tags.js
```diff
@@ -72,15 +72,19 @@ export default class Tags extends Collection {
...
     }
+    let sanitizedName = this.sanitize(newName);
+    if (!sanitizedName)
+      throw new Error(`${isColorTag ? "Color" : "Tag"} title cannot be empty.`);
 
-    newName = this.sanitize(newName);
-    if (!newName) throw new Error("Tag title cannot be empty.");
+    if (!isColorTag) {
+      newName = sanitizedName;
+    }
```
thecodrr, line 87: Just don't do any sanitization on rename. There's no need to introduce a new parameter.

Abdulrehman-Jafer: Sanitization is also required for renaming color as if we don't the user can save a color name with only spaces like "        " which will not be good.

thecodrr: > which will not be good

Yeah I don't think that's an issue. If a user does that they can rename again. Renaming should allow full freedom.

Abdulrehman-Jafer: Ok

packages/core/collections/tags.js
```diff
@@ -72,15 +72,15 @@ export default class Tags extends Collection {
...
+  async rename(tagId, newName, isColorTag) {
     let tag = this.tag(tagId);
     if (!tag) {
       console.error(`No tag found. Tag id:`, tagId);
       return;
     }
 
-    newName = this.sanitize(newName);
-    if (!newName) throw new Error("Tag title cannot be empty.");
+    if (!isColorTag) newName = this.sanitize(newName);
```
thecodrr, line 82: There is no need to do any sanitization in any case during renaming.

apps/web/src/common/dialog-controller.tsx
```diff
@@ -209,9 +209,8 @@ export function showError(title: string, message: string) {
 export function showMultiDeleteConfirmation(length: number) {
   return confirm({
     title: `Delete ${length} items?`,
-    message: `These items will be **kept in your Trash for ${
-      db.settings?.getTrashCleanupInterval() || 7
-    } days** after which they will be permanently deleted.`,
+    message: `These items will be **kept in your Trash for ${db.settings?.getTrashCleanupInterval() || 7
+      } days** after which they will be permanently deleted.`,
```
thecodrr, line 213: What's this change?

Abdulrehman-Jafer: Just auto formatted by prettier.

packages/core/collections/tags.js
Abdulrehman-Jafer: Ok

Abdulrehman-Jafer:
Now the sanitization is disabled for both the tags and colors.

thecodrr (approved):

PR #2287, opened 2023-04-06
Migrated use-database.js to TypeScript

Abdulrehman-Jafer:


Abdulrehman-Jafer:
Forgot to staged the deleted .js file in first commit

apps/web/src/hooks/use-database.ts
```diff
@@ -28,14 +28,14 @@ if (process.env.NODE_ENV === "production") {
...
-export default function useDatabase(persistence) {
+export default function useDatabase(persistence?: "db" | "memory") {
   const [isAppLoaded, setIsAppLoaded] = useState(memory.isAppLoaded);
 
   useEffect(() => {
     if (memory.isAppLoaded) return;
 
     (async () => {
-      await import("../app.css");
+      await import("../app.css" as string);
```
thecodrr, line 38: Why is this needed?

Abdulrehman-Jafer: This problem I faced and found out Typescript won't let us import the css file it will only allowed .ts file that is why as string is necessary to import the .css file.

Abdulrehman-Jafer: Otherwise there was an error module not found.

thecodrr (approved):

PR #2328, opened 2023-04-11
web: added dialog to handle adding tags

alihamuh:


thecodrr:
The UI currently for assigning tags is not right. We should use `FilteredList` to show the list of tags.

PR #2387, opened 2023-04-19
web: added functionality and ui

alihamuh:


ammarahm-ed:
Reimplemented in #9167

PR #2465, opened 2023-04-29
web: Fixed Issue #2450

RohanJacob23:
web: Issue #2450

Fix positioning of text in checklists #2450 when font size is increased

### Before
Screenshot (211)

### After
Screenshot (212)

Signed-off-by: Rohan Koshy Jacob <rohankoshyjacob@gmail.com>


thecodrr:
@RohanJacob23 thanks for the PR. I think a better way to fix this issue would be to increase the size of the checkbox with the font. We are already doing this in the case of changing default font size from app settings.

RohanJacob23:
Thank you, @thecodrr, for your feedback and suggestion. So, should I work on increasing the size of the checkbox with the font as you suggested?

PR #2466, opened 2023-04-29
Fix issue where notes directly in a notebook could not be unlinked

alihamuh:
note can be unlinked.

### DIFF
--- a/packages/core/collections/notes.js
+++ b/packages/core/collections/notes.js
@@ -347,7 +347,6 @@ export default class Notes extends Collection {
 
     for (const noteId of noteIds) {
       const note = this.note(noteId);
-
       if (!note || note.deleted) {
         continue;
       }
--- a/packages/core/collections/notes.js
+++ b/packages/core/collections/notes.js
@@ -348,8 +348,12 @@ export default class Notes extends Collection {
     for (const noteId of noteIds) {
       const note = this.note(noteId);
 
+      if (!note || note.deleted) {
+        continue;
+      }
+
       if (topicId) {
-        if (!note || note.deleted || !note.notebooks) {
+        if (!note.notebooks) {
           continue;
         }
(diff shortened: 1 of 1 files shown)


packages/core/collections/notes.js
```diff
@@ -347,11 +347,11 @@ export default class Notes extends Collection {
 
     for (const noteId of noteIds) {
       const note = this.note(noteId);
-      if (!note || note.deleted || !note.notebooks) {
-        continue;
-      }
 
       if (topicId) {
+        if (!note || note.deleted || !note.notebooks) {
```
thecodrr, line 352: We should only check for `!note.notebooks` here. All other checks should remain where they were.

thecodrr (changes requested):

packages/core/collections/notes.js
alihamuh: okay

packages/core/collections/notes.js
```diff
@@ -347,11 +347,15 @@ export default class Notes extends Collection {
 
     for (const noteId of noteIds) {
       const note = this.note(noteId);
-      if (!note || note.deleted || !note.notebooks) {
+
+      if (!note || note.deleted) {
```
thecodrr, line 351:
```suggestion
      if (!note || note.deleted) {
```

packages/core/collections/notes.js
```diff
@@ -347,11 +347,15 @@ export default class Notes extends Collection {
...
         continue;
       }
 
       if (topicId) {
+        if (!note.notebooks) {
+          continue;
+        }
```
thecodrr, line 358:
```suggestion
        if (!note.notebooks) continue;
```

thecodrr (changes requested):

PR #2621, opened 2023-05-30
editor: attachment download on Readonly

RamiroPruis:
Issue: #2609 

Item solved
- Can not download attachments.

### DIFF
--- a/packages/editor/src/extensions/attachment/component.tsx
+++ b/packages/editor/src/extensions/attachment/component.tsx
@@ -32,7 +32,7 @@ export function AttachmentComponent(
   const { editor, node, selected } = props;
   const { filename, size, progress } = node.attrs;
   const elementRef = useRef<HTMLSpanElement>();
-
+  
   return (
     <>
       <Box
@@ -87,11 +87,14 @@ export function AttachmentComponent(
           {selected && (
             <ToolbarGroup
               editor={editor}
-              tools={[
+              tools={ editor.isEditable ? [
                 "removeAttachment",
                 "downloadAttachment",
                 "previewAttachment"
-              ]}
+              ] : [
+                "downloadAttachment"
+              ]
+            }
               sx={{
                 boxShadow: "menu",
                 borderRadius: "default",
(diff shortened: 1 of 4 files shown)


packages/editor/src/extensions/attachment/component.tsx
```diff
@@ -87,11 +87,14 @@ export function AttachmentComponent(
...
               editor={editor}
-              tools={[
+              tools={ editor.isEditable ? [
                 "removeAttachment",
                 "downloadAttachment",
                 "previewAttachment"
-              ]}
+              ] : [
+                "downloadAttachment"
+              ]
```
thecodrr, line 96: Add `previewAttachment` here as well.

packages/editor/src/extensions/react/selection-based-react-node-view.tsx
```diff
@@ -90,9 +90,7 @@ export class SelectionBasedNodeView<
     forwardRef?: ForwardRef
   ): React.ReactElement<unknown> | null {
     if (!this.options.component) return null;
-    const isSelected =
-      this.editor.isEditable &&
-      this.isSelectedNode(this.editor.view.state.selection);
+    const isSelected = this.isSelectedNode(this.editor.view.state.selection);
```
thecodrr, line 93: Instead of enabling selection for all node views, we should make it an option. That way we can specifically enable selection in readonly mode for each node view.

thecodrr (changes requested):

packages/editor/src/extensions/react/types.ts
```diff
@@ -62,4 +62,5 @@ export type ReactNodeViewOptions<P> = {
   shouldUpdate?: ShouldUpdate;
   contentDOMFactory?: (() => ContentDOM) | boolean;
   wrapperFactory?: () => HTMLElement;
+  enableSelection?: boolean;
```
thecodrr, line 65: Can we name it `forceEnableSelection` since we are overriding the defaults?

thecodrr (changes requested):

thecodrr (approved):

PR #2977, opened 2023-07-19
Feat slash commands

alihamuh:


mufeedali:
Hi, any reason why this was never merged? Currently, Notesnook feels very unintuitive for keyboard-centric users and this would be a huge step forward.

rodnheli:
Hi,
new to notesnook and i really tried most of the notetaking apps on the market - all have their pros and cons :-) 

the only real big thing i'm missing in notesnook are / slash commands what brings note taking to another level instead using the mouse.

i think user (like me) which want to switch from notion, outline, standard notes, docmost, confluence and so on - will consider slash command feature on top of their decision... 

i've seen slash command is on the roadmap  and planned - _my question is there still developement on it and if yes - do you have a timeline?_

Thank you very much!

devguy23857848467:
This feature would be wonderful to have merged.

thecodrr:
Closing in favor of #8059

PR #3014, opened 2023-07-25
editor: added KeyboardShortcuts and InputRules

alihamuh:


packages/editor/src/extensions/heading/heading.ts
```diff
@@ -40,5 +41,32 @@ export const Heading = TiptapHeading.extend({
...
+  addInputRules() {
+    return this.options.levels.map((level) => {
+      return textblockTypeInputRule({
+        find: new RegExp(`^(#{1,${level}})\\s$`),
+        type: this.type,
+        getAttributes: {
+          level,
+          textAlign: this.editor.state.selection.$from.parent.attrs.textAlign,
+          textDirection:
+            this.editor.state.selection.$from.parent.attrs.textDirection
```
thecodrr, line 67: This is not a good idea because it'll only be evaluated once when creating the input rules. We need to infer the attributes every time users uses `#### ` syntax. In order to do that, you have to convert `getAttributes` into a function (it supports receiving a function already). Like so:

```
getAttributes: () => ({ level })
```

Secondly, `this.editor.state.selection.$from.parent.attrs.textAlign` you need to add several null checks here so it doesn't crash.

thecodrr (changes requested):

PR #3252, opened 2023-09-02
Fix issue where task list parent item remains unchecked after moving the last unchecked sub-item outside via drag drop

alihamuh:


packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -193,6 +193,74 @@ export const TaskListNode = TaskList.extend({
...
+            for (const taskList of taskLists) {
+              const range = {
+                start: taskList.pos,
+                end: taskList.pos + taskList.node.nodeSize
+              };
+
+              const _isWithInList = range.start < _from && range.end > _to;
+              const isWithInList = range.start < from && range.end > to;
+
+              if (_isWithInList && isWithInList && from < _from) {
```
thecodrr, line 228: This will be too expensive as it iterates on all nodes to find all the task lists and then iterates on all task lists again. Why not get only the affected task list from `newState` based on the selection value? Then you won't need that last `if statement`.

packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -193,6 +193,74 @@ export const TaskListNode = TaskList.extend({
...
+                      checked: true
+                    }
+                  );
+
+                  return transaction;
+                }
+              }
+            }
+          }
+          return newState.tr;
```
thecodrr, line 262: Just return `null` instead.

packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -193,6 +193,74 @@ export const TaskListNode = TaskList.extend({
...
+              };
+
+              const _isWithInList = range.start < _from && range.end > _to;
+              const isWithInList = range.start < from && range.end > to;
+
+              if (_isWithInList && isWithInList && from < _from) {
+                const _itemIndex = _$from.index(_$from.depth - 2) + 1;
+                const _itemSubList = $from.parent.child(_itemIndex).lastChild;
+                const _isItemSubList =
+                  _itemSubList?.type.name === TaskList.name;
```
thecodrr, line 232: This whole logic is finicky and will probably break in various ways. Use `findParentNodeClosestToPos` instead to get the sub-tasklist.

packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -193,6 +193,74 @@ export const TaskListNode = TaskList.extend({
...
+                  _itemIndex,
+                  _$from.depth - 2
+                );
+
+                let isChecked = true;
+
+                for (let index = 0; index < _itemSubList.childCount; index++) {
+                  if (!_itemSubList.child(index).attrs.checked)
+                    isChecked = false;
+                }
```
thecodrr, line 246: There's no need to create your own `for` loop. Just use `node.forEach`.

packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -193,6 +193,74 @@ export const TaskListNode = TaskList.extend({
...
+                const _itemSubList = $from.parent.child(_itemIndex).lastChild;
+                const _isItemSubList =
+                  _itemSubList?.type.name === TaskList.name;
+
+                if (!_isItemSubList) return newState.tr;
+
+                const _itemPosition = $from.posAtIndex(
+                  _itemIndex,
+                  _$from.depth - 2
+                );
```
thecodrr, line 239: Why do this here instead of where it is needed?

thecodrr (changes requested):
The overall logic is correct but the code is not as performant as it could be.

thecodrr:
This is done but needs extensive tests before merging.

PR #3341, opened 2023-09-15
core: added missing equality sign

alihamuh:


thecodrr:
@alihamuh please add a test case for this.

thecodrr (approved):

PR #3480, opened 2023-09-28
Feat implement hide title setting

### LINKED ISSUE #2733: Allow hiding the title of note in browser/app window titlebar
I think it's a bad idea to have the title of the page be title of the note I'm writing, because then it shows up in the search auto complete history if I type something like notes. This is especially concerning since my note titles are sometimes extremely private and I do not want them to just appear in the browser suggestions for everyone to see. Maybe the page title when I write my notes can be something generic like "Notesnook"?

**Device information:**
App version: 2.5.2-3e95960-web
OS: Linux 64-bit
Browser: Chrome 114.0.0.0
Pro: false

johnny-2123:
## User can set "Hide Title" privacy setting to true
Notesnook1

## Page title will not update with the user's note title
Notesnook2

## User can toggle "Hide Title" privacy setting to false
Notesnook3

## Page title will update with the user's note title
Notesnook4

## closes #2733


apps/web/src/utils/hideTitle.ts
thecodrr, line 1: I don't think this abstraction is necessary.

apps/web/src/stores/editor-store.js
```diff
@@ -151,7 +152,11 @@ class EditorStore extends BaseStore {
     if (!note) return;
 
     noteStore.setSelectedNote(note.id);
-    setDocumentTitle(note.title);
+    if (!settingStore.get().hideTitle) {
+      setDocumentTitle(note.title);
+    } else {
+      setDocumentTitle();
+    }
```
thecodrr, line 159: Can we use a ternary instead of a full blown if-else block?

apps/web/src/stores/editor-store.js
```diff
@@ -247,7 +252,11 @@ class EditorStore extends BaseStore {
         state.session.dateEdited = note.dateEdited;
         state.session.attachmentsLength = attachments.length;
       });
-      setDocumentTitle(note.title);
+      if (!settingStore.get().hideTitle) {
+        setDocumentTitle(note.title);
+      } else {
+        setDocumentTitle();
+      }
```
thecodrr, line 259: Can we use a ternary instead of a full blown if-else block?

thecodrr (changes requested):

PR #3620, opened 2023-10-24
Disallow links in inline code

### LINKED ISSUE #2793: Web addresses are converted to links in inline code and are breaking inline code appearance
### What happened?

The appearance of links in continuous inline code is segmented from the rest of the code. I think it would be best to keep addresses as code rather than converting them into links. This is already the case for the code blocks and is considered standard for Markdown.

Notesnook:
image

GitHub markdown for example: 
image

- Type any inline code with a web address in it.
	- Eg: `sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc`
- See that the address is converted into a link and is segmented off from the rest of the inline code.

### Version

v2.5.3

### Platform/OS

Windows, Linux

### Relevant log output

_No response_

thecodrr:
Fixes #2793

### DIFF
--- a/packages/editor/src/index.ts
+++ b/packages/editor/src/index.ts
@@ -263,7 +263,7 @@ const useTiptap = (
         OutlineListItem,
         OutlineList.configure({ keepAttributes: true, keepMarks: true }),
         ListItem,
-        Code.extend({ excludes: "" }),
+        Code.extend({ excludes: "link" }),
         Codemark,
         MathInline,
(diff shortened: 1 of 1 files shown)


mufeedali:
image

This breaks copy pasting links (from a browser URL bar for example) into inline code

mufeedali:
Sorry, that isn't a new bug apparently, just tested in the latest stable

PR #3621, opened 2023-10-24
Change pwa theme color according to current theme

### LINKED ISSUE #2906: Chrome window color does not change with theme color
Reproduce the bug:

1 - Use google chrome

2 - Use color theme by chrome (e.g. https://chrome.google.com/webstore/detail/just-black/aghfnjkcakhmadgdomlmlhhaocbkloab)

3 - Download notesnook as chrome app (while on notesnook page - chrome upper right three dots - more tools - create shortcut - check "open as window" )

4 - Open notesnook chrome app as stand-alone window 

5 - Usually, the upper bar should match the accent color of the app, but it is always green. In other apps, when the accent color is changes so is the color of the upper bar.

**Device information:**
App version: 2.5.2-3e95960-web
OS: Linux 64-bit
Browser: Chrome 113.0.0.0
Pro: true

### LINKED ISSUE #3663: PWA theme color according to current theme

In PR #3621 (which is in response of #2906 ),  the PWA color changes according to "paragraph" or "accent" instead of "background".

I think if it will change according to "background" then it will be more immersive. Also "-" "X" signs for window operations could be in "accent" or "paragraph" color.

I have attached some screenshots. Please take a look into these. Thanks.

What it looks like as of now⤵️

What it will look like after changing it according the recommendations⤵️

_No response_

### Additional context

Please refer to #3621 and #2906

thecodrr:
Closes #2906
Closes #3663


Starship-Headlight:
Is it possible to add a separator between the upper bar and the rest of the page. It looks a little odd when the separators end before reaching the top.

PR #3678, opened 2023-10-31
Make duplicate filenames in attachments unique on download

thecodrr:


apps/web/src/utils/streams/attachment-stream.ts
```diff
@@ -63,9 +64,10 @@ export class AttachmentStream extends ReadableStream<ZipFile> {
         });
 
         if (file) {
-          const filePath = attachment.metadata.filename;
+          const filePath: string = attachment.metadata.filename;
+
```
thecodrr, line 68:
```suggestion
```

PR #3826, opened 2023-11-20
Fix weird janky animation when opening dialogs

alihamuh:


apps/web/src/components/dialog/index.tsx
```diff
@@ -212,6 +211,7 @@ function onAfterOpen(
   e: ReactModal.OnAfterOpenCallbackOptions | undefined,
   props: DialogProps
 ) {
+  console.log("OnAfterOpen");
```
thecodrr, line 214:
```suggestion
```

Issue #3856, opened 2023-11-23
links on login/sign-up screen can trap user on the twitter.com


dym-sh:
### What happened?

opened the desktop app for the first time and clicked on the link

![notesnook desktop-app noted 20231123](https://github.com/streetwriters/notesnook/assets/140316503/fdae6b6a-2a4e-4a93-ab3a-4526c3df55b8)

after which i could not escape twitter.com until closing the app entirely

![notesnook after clicking the link 20231123](https://github.com/streetwriters/notesnook/assets/140316503/4d3e0717-ba36-4ea2-b8e1-7b237a290cb9)

![notesnook the top most level within twitter 20231123](https://github.com/streetwriters/notesnook/assets/140316503/098f5fb1-eb00-433c-8175-11e38f5bf52a)

### Steps to reproduce the problem

1. open the desktop-app
2. click on one of the «Source» link provided with the quotation
3. get trapped on twitter.com

### Version

notesnook_linux_x86_64-v2.6.11.AppImage

### Platform/OS

Linux

Dharmavineta:
Hey, @thecodrr , how do you wish to go along with this feature?
Adding a button that'll route back to the main app is ok?

thecodrr:
The links should open in the default browser instead of inside the app.

alihamuh:
bug solved

PR #3863, opened 2023-11-24
fix: (#3856) link issue fixed

Dharmavineta:
Added the _target and rel fields which will open the link in a new tab. 

Fixes issue #3856


thecodrr:
Thank you!

PR #3963, opened 2023-12-14
editor: add emoji selector to editor (wip)

wendevlin:
I used https://tiptap.dev/api/nodes/mention since tiptap is already integrated. 

It actually recognizes the `:` but it doesn't trigger the render function.
I am not sure what the issue is.


wendevlin:
Sorry I wasn't able to finish it.

PR #4137, opened 2024-01-17
web: Fix incorrect title on scrolling note

Parthiv-M:


apps/web/src/components/editor/toolbar.jsx
```diff
@@ -92,6 +92,10 @@ function Toolbar() {
     };
   }, [isTitleVisible]);
 
+  useEffect(() => {
+    document.querySelector("#onScrollTitle").defaultValue = title;
```
thecodrr, line 96: Use a ref instead of `document.querySelector`.

thecodrr (changes requested):

thecodrr:
@Parthiv-M thanks for fixing this!

Parthiv-M:
I've made the change

apps/web/src/components/editor/toolbar.jsx
```diff
@@ -93,7 +94,7 @@ function Toolbar() {
   }, [isTitleVisible]);
 
   useEffect(() => {
-    document.querySelector("#onScrollTitle").defaultValue = title;
+    onScrollTitleRef.current.defaultValue = title;
```
thecodrr, line 97:
```suggestion
    if (onScrollTitleRef.current)
        onScrollTitleRef.current.defaultValue = title;
```

thecodrr (approved):

PR #4223, opened 2024-02-03
desktop: added command for autstart appimage

alihamuh:


apps/desktop/src/utils/autolaunch.ts
```diff
@@ -25,7 +25,11 @@ Type=Application
 Version=${app.getVersion()}
 Name=${app.getName()}
 Comment=${app.getName()} startup script
-Exec=${process.execPath}${hidden ? " --hidden" : ""}
+Exec=${
+  process.env.APPIMAGE
+    ? process.env.APPIMAGE
```
thecodrr, line 30:
```suggestion
    ? `${process.env.APPIMAGE}${hidden ? " --hidden" : ""}`
```

PR #4353, opened 2024-02-26
editor: excluded katex class from span styling

alihamuh:


packages/editor/styles/styles.css
```diff
@@ -1,6 +1,6 @@
 
 
-.ProseMirror span * {
+.ProseMirror span *:not(.katex) {
```
thecodrr, line 3:
```suggestion
.ProseMirror p span {
```

Issue #5081, opened 2024-05-01
Confusing Nested Notebook Display


ckdanny:
### What problem are you facing?

UI of the notebook display

### What's the solution you'd like to see in Notesnook?

The display of the nested notebook can be simplified as below, 
<img width="823" alt="image" src="https://github.com/streetwriters/notesnook/assets/11786201/cba98766-b0c9-4c29-882c-4aaf4a023e10">

Sub-notebook does not need to nest into the current notebook (i.e. inbox) and the expand there is not necessary. 
Suggest to improve this UI to show only nested notebook there and move the total count to next to the `NOTEBOOKS`.

ammarahm-ed:
We are changing how notebooks appear in notesnook in v3.1

PR #6011, opened 2024-06-26
web: fix confusing nested display

Adityakale12834:
**This pull request makes the following changes.**
* fixes issue  #5081 
 
* The display of notebook is simplified 
Screenshot 2024-06-26 230932

* Sub-notebook does not need to nest into the current notebook (i.e. inbox) and the expand there is not necessary.
Screenshot 2024-06-26 231245


thecodrr:
This is already done in #5543

PR #6255, opened 2024-07-28
global: clarify explanations for force push and force pull

martinfrancois:
- I noticed the explanation of force push is misdescribed in the mobile version, it has the same explanation as for force pull
- There's a spelling mistake I noticed being made `with whatever is one` that I fixed => `with whatever is on`
- I rephrased the wording to clarify when the buttons need to be pressed and what happens:
  from: `Use this if some changes from this device are not appearing on other devices.`
  to: `Use this if changes made on this device are not appearing on other devices.`
  and:
  from: `This will pull everything from the server and overwrite with whatever is one this device.`
  to: `This will overwrite the data on this device with the latest data from the server.`
- I noticed on mobile it only shows `These must only be used for troubleshooting...` between pull sync and push sync, which I thought could be a bit confusing - as only pull sync was explained so far, it isn't 100% clear what "these" refers to. Also, it could result in some people not knowing that this also relates to push sync, as it appears only after pull sync.
  - I considered only adding it after push sync, as then both pull and push sync already appeared, so "these" would make more sense, but as both operations are quite critical, I thought it would be better to stay on the safe side and decided to explain it with force push and force pull once each.

Let me know if you disagree with my decisions or feel something could be improved further!

### DIFF
--- a/apps/mobile/app/screens/settings/settings-data.tsx
+++ b/apps/mobile/app/screens/settings/settings-data.tsx
@@ -475,7 +475,7 @@ export const settingsGroups: SettingSection[] = [
           {
             id: "pull-sync",
             name: "Force pull changes",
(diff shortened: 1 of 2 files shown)


thecodrr (approved):

thecodrr:
Thanks!

PR #6435, opened 2024-08-26
desktop: fix issues with codeblock toggling

seanwynwins:
Addresses streetwriters/notesnook#6321


packages/editor/src/extensions/code-block/code-block.ts
```diff
@@ -234,11 +234,34 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
...
+        ({ commands, state }) => {
+          const { from, to } = state.selection;
+
+          let hasParagraph = false;
+
+          state.doc.nodesBetween(from, to, (node) => {
+            if (node.type.name === "paragraph") {
+              hasParagraph = true;
+            }
           });
```
thecodrr, line 246: Instead of checking whether selection has paragraph, we can just check whether a codeblock node is active using `isNodeActive` function.

packages/editor/src/extensions/code-block/code-block.ts
```diff
@@ -234,11 +234,34 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
...
+                ...attributes,
+                id: createCodeblockId()
+              },
+              content: [
+                {
+                  type: "text",
+                  text: state.doc.textBetween(from, to, "\n")
+                }
+              ]
+            });
```
thecodrr, line 261: This can be simplified like so:

```ts
return commands.insertContent(
  this.type.create(
    { ...attributes, id: createCodeblockId() },
    state.schema.text(state.doc.textBetween(from, to, "\n"))
  )
);
```

However, I don't recommend depending on `insertContent` for replacing the selection. Instead we can directly use the `replaceSelectionWith` method on `Transaction`:

```ts
          const { from, to } = tr.selection;
          const text = tr.doc.textBetween(from, to, "\n");
          tr.replaceSelectionWith(
            this.type.create(
              { ...attributes, id: createCodeblockId() },
              state.schema.text(text)
            )
          );
```

Not to mention, that after replacing the node we also need to set the selection like so:

```ts
          commands.setTextSelection({ from, to: tr.mapping.map(to) });
```

---

This is all very rough so you might need to make adjustments.

thecodrr (changes requested):
Thank you so much for fixing this!

I noticed a bug as well where toggling the code block added an empty line above it. Not sure why that's happening.

packages/editor/src/extensions/code-block/code-block.ts
seanwynwins: Are you saying that the selected text should only be converted if a codeblock node is not active?

I wonder if the behavior should be to convert all selected text to a codeblock, if any one of the selected nodes is of a type other than codeblock?

packages/editor/src/extensions/code-block/code-block.ts
seanwynwins: Sounds good, just changed this

seanwynwins:
> Thank you so much for fixing this!
> 
> I noticed a bug as well where toggling the code block added an empty line above it. Not sure why that's happening.

No problem! I think using the transactions directly solved this issue

thecodrr:
@seanwynwins I think you didn't push the changes.

seanwynwins:
> @seanwynwins I think you didn't push the changes.

Oh I didn't push yet because I wanted to check on this:

https://github.com/streetwriters/notesnook/pull/6435#discussion_r1735496221

packages/editor/src/extensions/code-block/code-block.ts
thecodrr: > Are you saying that the selected text should only be converted if there are no active codeblock nodes?

Yes. Because if the cursor is inside a codeblock we'll just convert it into a paragraph. Otherwise, everything gets converted into a codeblock.

seanwynwins:
> @seanwynwins I think you didn't push the changes.

Just got back from vacation, will push tomorrow

seanwynwins:
Sorry for the delay, can you review?

thecodrr:
Thanks!

PR #6703, opened 2024-10-15
Add docs for app lock

alihamuh:


docs/help/contents/app-lock.md
```diff
@@ -0,0 +1,43 @@
+# App lock
+
+> error Not available on web
+>
+> App lock is not available on Web.
```
thecodrr, line 5: App lock is available on web & desktop.

docs/help/contents/app-lock.md
```diff
@@ -0,0 +1,43 @@
...
+
+<img src="/desktop-enable-app-lock.png" alt="drawing" height="500"/>
+
+2. You can set the time out for your App Lock
+
+<img src="/desktop-lock-app-after.png" alt="drawing" height="500"/>
+
+3. You can also change the pin or password or set a security key.
+
+<img src="/desktop-password-key.png" alt="drawing" height="500"/>
```
thecodrr, line 21: We should split these into their own headings so its easier to refer a specific App Lock feature to a user. That way we'll also be able to add more information to each topic easily.

thecodrr (changes requested):

alihamuh:
new commit made 2.0

PR #6762, opened 2024-10-24
editor: added style to handle katex math render

alihamuh:


packages/editor/styles/styles.css
```diff
@@ -2,6 +2,10 @@
   font-family: inherit;
 }
 
+.ProseMirror span.math-inline {
+  font-family: KaTeX_Main;
```
thecodrr, line 6:
```suggestion
  font-family: KaTeX_Main, Times New Roman, serif; 
```

thecodrr (approved):

PR #6787, opened 2024-10-29
desktop: fix the package dependencies error

alicezou:
(
    ✘ [ERROR] Could not resolve "@notesnook/intl/dist/locales/$pseudo-LOCALE.json"
    ✘ [ERROR] Could not resolve "@notesnook/intl/dist/locales/$en.json"
)
Signed-off-by: Jiayun Zou <alicezou26@gmail.com>

### DIFF
--- a/packages/intl/package.json
+++ b/packages/intl/package.json
@@ -16,7 +16,7 @@
         "default": "./dist/index.mjs"
       }
     },
-    "./locales/*.json": "./dist/locales/*.json"
+    "./dist/locales/*.json": "./dist/locales/*.json"
   },
   "repository": {
     "type": "git",


packages/intl/package.json
```diff
@@ -16,7 +16,7 @@
         "default": "./dist/index.mjs"
       }
     },
-    "./locales/*.json": "./dist/locales/*.json"
+    "./dist/locales/*.json": "./dist/locales/*.json"
```
thecodrr, line 19: Ah no, this was correct the way it was before.

thecodrr (changes requested):

PR #6818, opened 2024-11-03
web: changes related to table of contents (#4977)

### LINKED ISSUE #4977: Table of contents feature improvements for web editor

I'm so happy to have an outline / table of contents. I heavily rely on headings for hierarchy of my notes so this feature is very well appreciated. Being able to easily navigate large notes is nice. However, the implementation isn't the most useful. 

 I think these three changes would make this feature much more useful for everyone. 

- [ ] A single button below the top-right menu to expand and collapse the ToC
The ToC should be below this menu so that the menu is still accessible.

- [ ] Visibility should be statefull, so it will always be in the state last used
Because I use this feature a lot, I would prefer if the ToC was always visible. Currently I need to open it every time I start Notesnook. Then if I want to access the menu in the top right, I need to close it. 

- [ ] Increase the indentation so the hierarchy is easier to see
I think the indent should be two or three times as much.

Here's what these changes might look like:
collapsed
expanded

_No response_

### Additional context

Additional interactions: Focus Mode should not hide the ToC. Disable Editor Margins should remove margins, but only up to the ToC (if it is expanded). 

- [ ] Bonus points for making the ToC width adjustable. I love how much flexibility the sidebar and notes frames have. 

The iOS ToC has appropriate indentation.

### LINKED ISSUE #6711: Table of Contents covers note contents
### What happened?

If you open the table of contents, the side panel covers the note contents. Instead, the contents should re-wrap to the smaller open editing window so that you can still edit the note with the table of contents open.

1. Create a new note.
2. Insert some random long lines of text.
3. Open the table of contents.
4. Create a new line and keep typing - the line will go under the table of contents and you cannot see what you are typing.

### Version

3.0.18-5f0d14f-desktop

### Platform/OS

Linux

### Relevant log output

_No response_

01zulfi:
* move toc button out of action bar into fixed position
* adjust editor width based on toc state
* save toc state in local storage
* increase indentation in toc

Signed-off-by: Zulfiqar Ali <85733202+01zulfi@users.noreply.github.com>

Closes #4977 
Closes #6711

### DIFF
--- a/apps/web/src/components/editor/action-bar.tsx
+++ b/apps/web/src/components/editor/action-bar.tsx
@@ -169,8 +169,7 @@ export function EditorActionBar() {
         activeSession.type !== "locked" &&
         activeSession.type !== "diff" &&
         activeSession.type !== "conflicted",
-      onClick: () => useEditorStore.getState().toggleTableOfContents(),
-      toc: true
+      onClick: () => useEditorStore.getState().toggleTableOfContents()
     },
     {
       title: strings.search(),
@@ -225,7 +224,7 @@ export function EditorActionBar() {
           title={tool.title}
           key={tool.title}
           sx={{
-            height: tool.toc ? 32 : "100%",
+            height: "100%",
             alignItems: "center",
             bg: "transparent",
             display: [
@@ -239,11 +238,7 @@ export function EditorActionBar() {
                 tool.title === "Close"
                   ? "var(--accentForeground-error) !important"
                   : "var(--icon)"
-            },
-            position: tool.toc ? "absolute" : "static",
-            right: tool.toc ? 10 : "auto",
-            top: tool.toc ? 32 : "auto",
-            zIndex: tool.toc ? 1999 : "auto"
+            }
           }}
           onClick={tool.onClick}
         >
--- a/apps/web/src/components/editor/action-bar.tsx
+++ b/apps/web/src/components/editor/action-bar.tsx
@@ -169,7 +169,8 @@ export function EditorActionBar() {
         activeSession.type !== "locked" &&
         activeSession.type !== "diff" &&
         activeSession.type !== "conflicted",
-      onClick: () => useEditorStore.getState().toggleTableOfContents()
+      onClick: () => useEditorStore.getState().toggleTableOfContents(),
+      toc: true
     },
     {
       title: strings.search(),
@@ -224,7 +225,7 @@ export function EditorActionBar() {
           title={tool.title}
           key={tool.title}
           sx={{
-            height: "100%",
+            height: tool.toc ? 32 : "100%",
             alignItems: "center",
             bg: "transparent",
             display: [
@@ -238,7 +239,11 @@ export function EditorActionBar() {
                 tool.title === "Close"
                   ? "var(--accentForeground-error) !important"
                   : "var(--icon)"
-            }
+            },
+            position: tool.toc ? "absolute" : "static",
+            right: tool.toc ? 10 : "auto",
+            top: tool.toc ? 32 : "auto",
+            zIndex: tool.toc ? 1999 : "auto"
           }}
           onClick={tool.onClick}
         >
--- a/apps/web/src/components/editor/index.tsx
+++ b/apps/web/src/components/editor/index.tsx
@@ -63,7 +63,7 @@ import { debounce, debounceWithId } from "@notesnook/common";
 import { Freeze } from "react-freeze";
 import { UnlockView } from "../unlock";
 import DiffViewer from "../diff-viewer";
-import TableOfContents from "./table-of-contents";
+import TableOfContents, { TABLE_OF_CONTENTS_WIDTH } from "./table-of-contents";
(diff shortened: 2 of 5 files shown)


apps/web/src/components/editor/action-bar.tsx
thecodrr, line 1: I think moving the TOC button is unnecessary. Let's undo that.

apps/web/src/components/editor/table-of-contents.tsx
```diff
@@ -102,10 +104,10 @@ function TableOfContents(props: TableOfContentsProps) {
         display: "flex",
         position: "absolute",
         right: 0,
-        top: TITLE_BAR_HEIGHT,
+        top: 32,
```
thecodrr, line 107: Why this change?

apps/web/src/components/editor/index.tsx
```diff
@@ -624,9 +625,15 @@ function EditorChrome(props: PropsWithChildren<EditorProps>) {
...
+            alignSelf: [
+              "stretch",
+              focusMode ? "center" : "stretch",
+              !editorMargins ? (isTOCVisible ? "stretch" : "center") : "center"
+            ],
             maxWidth: editorMargins ? "min(100%, 850px)" : "auto",
-            width: "100%"
+            width: isTOCVisible
+              ? `calc(100% - ${TABLE_OF_CONTENTS_WIDTH}px)`
+              : "100%"
```
thecodrr, line 636: I don't think this is the best way to do this. Instead of altering the width of the editor, we can render the TableOfContents to the right of the editor. That way it'll automatically resize when TOC becomes visible.

thecodrr (changes requested):

apps/web/src/components/editor/table-of-contents.tsx
01zulfi: `TITLE_BAR_HEIGHT` is 0 for web and since I moved the TOC a little below, I had to give some value to `top`. But now TOC isn't in absolute position but rather rendered to the right of editor. Hence, I've reverted this change as not needed anymore.

apps/web/src/components/editor/table-of-contents.tsx
```diff
@@ -49,6 +49,8 @@ import { useEditorManager } from "./manager";
 import { TITLE_BAR_HEIGHT } from "../title-bar";
 import { strings } from "@notesnook/intl";
 
+export const TABLE_OF_CONTENTS_WIDTH = 300;
+
```
thecodrr, line 53: This isn't needed anymore.

thecodrr (changes requested):

apps/web/src/components/editor/index.tsx
```diff
@@ -152,7 +152,12 @@ export default function TabsView() {
                 ) : session.type === "conflicted" || session.type === "diff" ? (
                   <DiffViewer session={session} />
                 ) : (
-                  <MemoizedEditorView session={session} />
+                  <Flex sx={{ overflow: "hidden" }}>
+                    <MemoizedEditorView session={session} />
+                    {isTOCVisible && activeSessionId && (
+                      <TableOfContents sessionId={activeSessionId} />
+                    )}
+                  </Flex>
```
thecodrr, line 160: Instead of wrapping this in its own Flex, we should move the TOC to its own Panel. That way it'll become automatically resizeable.

apps/web/src/components/editor/table-of-contents.tsx
thecodrr, line 1: 2 things:

1. Let's get rid of the opening/closing animation on TableOfContents.
2. Instead of showing the Back arrow icon for closing the TOC, we should show a Cross button on the right side.

thecodrr (changes requested):
Just some final changes. Everything else looks good.

thecodrr (approved):

apps/web/src/components/editor/index.tsx
```diff
@@ -204,14 +204,17 @@ export default function TabsView() {
               </Panel>
             </>
           )}
+
+          {isTOCVisible && activeSessionId && (
+            <Panel order={3} defaultSize={30}>
```
thecodrr, line 209: You need to add `PanelResizeHandle` otherwise it won't resize. Also set `defaultSize` to 25 and minSize to 15. It looks too big. You'll also need to add an `id` prop to the handle otherwise the `order` doesn't work.

thecodrr (changes requested):

apps/web/src/components/editor/index.tsx
01zulfi: Add `id` prop to `Panel` or `PanelResizeHandle`. I've added `id` to `Panel` as it is in other panels in the component. Let me know if we want to add to `PanelResizeHandle` as well

thecodrr: I don't think adding it to `PanelResizeHandle` is necessary.

thecodrr (approved):

PR #6830, opened 2024-11-05
web: add vault in status bar

### LINKED ISSUE #6626: Show indication when vault is unlocked in status bar
Hey there

I just found that even if a note is in lock mode, we can export the notes in different formats, here is the irony all the contents can be seen without any passcode when exported the locked notes. Why do we have locked notes if can see the notes contents clearly if exported.  this happened when the note is lockdown mode, I checked clearly. please fix this issues ASAP.

**Device information:**
App version: 3.0.18-5f0d14f-web
OS: Windows 10 64-bit
Browser: Chrome 129.0.0.0
Pro: true

01zulfi:
* add vault lock functionality for user
* add new vault EVENTS

Closes #6626

### DIFF
--- a/apps/web/src/common/vault.ts
+++ b/apps/web/src/common/vault.ts
@@ -184,5 +184,10 @@ class Vault {
       }
     });
   }
+
+  static async lockVault() {
+    await db.vault.lock();
+    showToast("success", strings.vaultLocked());
+  }
 }
 export default Vault;
--- a/apps/web/src/components/status-bar/index.tsx
+++ b/apps/web/src/components/status-bar/index.tsx
@@ -27,7 +27,9 @@ import {
   SyncError,
   Alert,
   SyncOff,
-  Icon
+  Icon,
+  Lock,
+  Unlock
 } from "../icons";
 import { useStore as useUserStore } from "../../stores/user-store";
 import { useStore as useAppStore } from "../../stores/app-store";
@@ -40,6 +42,7 @@ import { getTimeAgo, toTitleCase } from "@notesnook/common";
 import { User } from "@notesnook/core";
 import { showUpdateAvailableNotice } from "../../dialogs/confirm";
 import { strings } from "@notesnook/intl";
+import { useVault } from "../../hooks/use-vault";
 
 function StatusBar() {
   const user = useUserStore((state) => state.user);
@@ -47,6 +50,7 @@ function StatusBar() {
   const statuses = useStatus();
   const updateStatus = useAutoUpdater();
   const isFocusMode = useAppStore((state) => state.isFocusMode);
+  const { isVaultLocked, lockVault } = useVault();
 
   return (
     <ScopedThemeProvider
@@ -157,6 +161,22 @@ function StatusBar() {
               />
               <Text variant="subBody" ml={1} sx={{ color: "paragraph" }}>
                 {statusToInfoText(updateStatus)}
+              </Text>
+            </Button>
+          )}
+          {!isVaultLocked && (
+            <Button
+              variant="statusitem"
+              onClick={lockVault}
+              sx={{
+                alignItems: "center",
+                justifyContent: "center",
+                display: "flex"
+              }}
+            >
+              <Unlock size={10} />
+              <Text variant="subBody" ml={1} sx={{ color: "paragraph" }}>
+                {strings.vaultUnlocked()}
               </Text>
             </Button>
           )}
--- a/packages/intl/src/strings.ts
+++ b/packages/intl/src/strings.ts
@@ -1982,7 +1982,6 @@ All attachments will be downloaded & cached again on access.
       ]()}`
     }),
   backupReadyToDownload: () => t`Your backup is ready to download`,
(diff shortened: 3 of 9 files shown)


apps/web/src/common/vault.ts
```diff
@@ -184,5 +184,18 @@ class Vault {
...
   }
+
+  static async lockVault() {
+    await db.vault.lock();
+    showToast("success", strings.vaultLockedSuccessfully());
+  }
+
+  static async doesExist() {
+    return db.vault.exists();
+  }
```
thecodrr, line 195: This is unnecessary. Use `db.vault.exists()` directly.

apps/web/src/hooks/use-vault-lock.ts
```diff
@@ -0,0 +1,23 @@
+import { EV, EVENTS } from "@notesnook/core";
+import { useEffect, useState } from "react";
+import Vault from "../common/vault";
+
+export default function useVaultLock() {
```
thecodrr, line 5: Let's rename this to `useVault` and by convention we no longer do `default` exports.

apps/web/src/hooks/use-vault-lock.ts
```diff
@@ -0,0 +1,23 @@
...
+  const [isLocked, setIsLocked] = useState(Vault.isLocked());
+  const [doesExist, setDoesExist] = useState(false);
```
thecodrr, line 7: I think a better name for this would be `vaultExists` and `setVaultExists`

packages/intl/src/strings.ts
```diff
@@ -1982,6 +1982,9 @@ All attachments will be downloaded & cached again on access.
       ]()}`
     }),
   backupReadyToDownload: () => t`Your backup is ready to download`,
+  vaultLocked: () => t`Vault locked`,
+  vaultLockedSuccessfully: () => t`Vault locked successfully`,
```
thecodrr, line 1986: This is unnecessary. The word "successfully" is not meaningful. We can just use `vaultLocked` instead.

apps/web/src/components/status-bar/index.tsx
```diff
@@ -160,6 +165,28 @@ function StatusBar() {
...
+              sx={{
+                alignItems: "center",
+                justifyContent: "center",
+                display: "flex"
+              }}
+            >
+              {isVaultLocked ? (
+                <Lock size={10} color="var(--icon-error)" />
+              ) : (
+                <Unlock size={10} color="var(--icon-success)" />
```
thecodrr, line 181: Just use the default icon color for both. It'll be distracting otherwise. Also, I don't think `var(--` is necessary. Direct color names will work too.

apps/web/src/common/vault.ts
```diff
@@ -184,5 +184,18 @@ class Vault {
...
+
+  static isLocked() {
+    return !db.vault.unlocked;
+  }
```
thecodrr, line 199: This is unnecessary. Use `db.vault.unlocked` directly. No need for this abstraction.

thecodrr (changes requested):

apps/web/src/components/status-bar/index.tsx
01zulfi: I'm not fond of the color as well. I put it because at that icon size it's hard to tell between the lock and unlock icon. I've removed the colors, if you have any suggestion here, please let me know

apps/web/src/hooks/use-vault-lock.ts
```diff
@@ -0,0 +1,24 @@
...
+  useEffect(() => {
+    db.vault.exists().then(setVaultExists);
+    EV.subscribe(EVENTS.vaultCreated, () => setVaultExists(true));
+    EV.subscribe(EVENTS.vaultDeleted, () => setVaultExists(false));
+    EV.subscribe(EVENTS.vaultLocked, () => setIsLocked(true));
+    EV.subscribe(EVENTS.vaultUnlocked, () => setIsLocked(false));
+  }, []);
+
+  return {
+    doesVaultExist: vaultExists,
```
thecodrr, line 19:
```suggestion
    vaultExists,
```

thecodrr (approved):

thecodrr (approved):

packages/intl/src/strings.ts
```diff
@@ -1982,6 +1982,8 @@ All attachments will be downloaded & cached again on access.
       ]()}`
     }),
   backupReadyToDownload: () => t`Your backup is ready to download`,
+  vaultLocked: () => t`Vault locked`,
```
thecodrr, line 1985:
```suggestion
```

thecodrr (approved):

packages/intl/src/strings.ts
01zulfi: This string is used here:

https://github.com/streetwriters/notesnook/pull/6830/files#diff-45c664ba81d530ce7480cf6bf8ad6d3680aa845598589257607979adc7b3c169R190

thecodrr: Oops. Undoing...

PR #6837, opened 2024-11-06
web: fix restore notebooks over limit for free users (#6812)

### LINKED ISSUE #6812: Notebooks & Subnotebooks
Hi, has the number of notebooks a free user can make increased from 3? And does the number of subnotebooks a free user can make have a limit? Sorry, I'm genuinely not sure if this is a bug I'm experiencing.

Currently, I have 3 notebooks and 17 subnotebooks. When I tried making a new subnotebook, it tells me that I already need to subscribe to the Pro plan/version to add more notebooks. I deleted some subnotebooks and tried making new notebooks and I was able to have more than 3 notebooks. I was able to add more notebooks as long as I deleted some subnotebooks. Is the limit for subnotebooks and notebooks combined to 20?

Also, I checked the trash and tried to see what would happen if I restored the deleted notebooks and subnotebooks despite the limit and it all got restored. It seems like the limit of 3 notebooks (based on what was stated) was bypassed because I was able to restore more than 5 notebooks despite the limit that's been written. The only thing I couldn't do was add new notebooks and/or subnotebooks. I even tested what would happen if I kept making new notebooks and/or subnotebooks and kept deleting them then restoring them. I figured that all of those notebooks and subnotebooks could be restored despite the limit. It's just that I can't make new ones until I repeat the process.

_______________
**Device information:**
App version: 3.0.20
Platform: android
Device: samsung-SM-A225F-13
Pro: false
Logged in: yes

01zulfi:
Closes #6812

### DIFF
--- a/packages/core/src/collections/trash.ts
+++ b/packages/core/src/collections/trash.ts
@@ -29,6 +29,11 @@ import {
 } from "../utils/grouping.js";
 import { sql } from "@streetwriters/kysely";
 import { MAX_SQL_PARAMETERS } from "../database/sql-collection.js";
+import {
+  CHECK_IDS,
+  checkIsUserPremium,
+  FREE_NOTEBOOKS_LIMIT
+} from "../common.js";
 
 export default class Trash {
   collections = ["notes", "notebooks"] as const;
@@ -192,6 +197,13 @@ export default class Trash {
     }
 
     if (notebookIds.length > 0) {
+      const notebooksLimitReached =
+        (await this.db.notebooks.all.count()) + notebookIds.length >
+        FREE_NOTEBOOKS_LIMIT;
+      const isUserPremium = await checkIsUserPremium(CHECK_IDS.notebookAdd);
+      if (notebooksLimitReached && !isUserPremium) {
+        return false;
+      }
       const ids = [...notebookIds, ...(await this.subNotebooks(notebookIds))];
       await this.db.notebooks.collection.update(ids, {
         type: "notebook",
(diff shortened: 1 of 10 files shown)


packages/core/src/collections/trash.ts
```diff
@@ -192,6 +197,12 @@ export default class Trash {
     }
 
     if (notebookIds.length > 0) {
+      const canRestoreNotebooks =
+        (await this.db.notebooks.all.count()) >= FREE_NOTEBOOKS_LIMIT &&
```
thecodrr, line 201: This'll only work prevent restoration if the notebooks length exactly matches `FREE_NOTEBOOKS_LIMIT` but what if user has 10 deleted notebooks in trash and 12 notebooks otherwise? After restoration, the user will have a total of 22 notebooks (2 more than allowed).

thecodrr (changes requested):

packages/core/src/collections/trash.ts
01zulfi: Nice catch! I've included `notebookIds.length` in the check, hope that caters to this case

packages/core/src/collections/trash.ts
```diff
@@ -192,6 +197,13 @@ export default class Trash {
     }
 
     if (notebookIds.length > 0) {
+      const cannotRestoreNotebooks =
+        (await this.db.notebooks.all.count()) + notebookIds.length >
+          FREE_NOTEBOOKS_LIMIT &&
+        !(await checkIsUserPremium(CHECK_IDS.notebookAdd));
+      if (cannotRestoreNotebooks) {
```
thecodrr, line 204: I think this would be more readable as:

```
if (notebooksLimitReached and !checkIsUserPremium)
```

thecodrr (changes requested):

thecodrr (approved):

PR #6844, opened 2024-11-07
Fix duplicate window controls showing on Linux desktop app

alihamuh:


apps/desktop/src/main.ts
```diff
@@ -101,7 +101,7 @@ async function createWindow() {
     ...(config.desktopSettings.nativeTitlebar
       ? {}
       : {
-          titleBarStyle: "hidden",
+          titleBarStyle: process.platform === "linux" ? "default" : "hidden",
```
thecodrr, line 104:
```suggestion
          titleBarStyle: process.platform === "win32" || process.platform === "darwin" ? "hidden" : "default",
```

PR #6846, opened 2024-11-07
web: add text zoom functionality in editor (#6633)

### LINKED ISSUE #6633: Add text zoom functionality with "Ctrl + Scroll"

The current desktop version of Notesnook does not allow users to zoom in or out on the text of a note using a simple "Ctrl + Scroll" combination. This makes it inconvenient for users who want to quickly adjust the text size for better readability without having to change global font settings or go through menus. This is especially an issue for users with visual impairments or those who prefer different text sizes for different contexts. 

I would like to propose the implementation of a text zoom function within the notes using the "Ctrl + Scroll" combination in the desktop version of Notesnook, similar to what is available in many modern text editors and applications, such as the default text editor in the GNOME graphical environment. 

The only alternative currently is manually adjusting the font size in the settings menu, but this is a slower, less convenient process and applies to all notes globally. I also considered using browser-level zoom, but this affects the entire interface and not just the note content, which can lead to an awkward experience.

### Additional context

_No response_

01zulfi:

Closes #6633

### DIFF
--- a/apps/web/src/components/editor/common.ts
+++ b/apps/web/src/components/editor/common.ts
@@ -1,0 +1,25 @@
+/*
+This file is part of the Notesnook project (https://notesnook.com/)
+
+Copyright (C) 2023 Streetwriters (Private) Limited
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+export const EDITOR_ZOOM = {
+  DEFAULT: 100,
+  MAX: 200,
+  MIN: 50,
+  STEP: 10
+};
--- a/apps/web/src/components/editor/common.ts
+++ b/apps/web/src/components/editor/common.ts
@@ -19,7 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 export const EDITOR_ZOOM = {
   DEFAULT: 100,
-  MAX: 200,
-  MIN: 50,
+  MAX: 500,
+  MIN: 30,
   STEP: 10
 };
--- a/apps/web/src/components/editor/footer.tsx
+++ b/apps/web/src/components/editor/footer.tsx
@@ -20,10 +20,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import { Button, Flex, Text } from "@theme-ui/components";
 import { SaveState, useEditorStore } from "../../stores/editor-store";
 import { Loading, Saved, NotSaved } from "../icons";
-import { useNoteStatistics } from "./manager";
+import { useEditorConfig, useNoteStatistics } from "./manager";
 import { getFormattedDate } from "@notesnook/common";
 import { MAX_AUTO_SAVEABLE_WORDS } from "./types";
 import { strings } from "@notesnook/intl";
+import { EDITOR_ZOOM } from "./common";
 
 const SAVE_STATE_ICON_MAP = {
   "-1": NotSaved,
@@ -34,6 +35,7 @@ const SAVE_STATE_ICON_MAP = {
 function EditorFooter() {
   const { words } = useNoteStatistics();
   const session = useEditorStore((store) => store.getActiveSession());
+  const { editorConfig, setEditorConfig } = useEditorConfig();
   if (!session) return null;
 
   const saveState =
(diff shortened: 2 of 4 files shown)


apps/web/src/components/editor/tiptap.tsx
```diff
@@ -443,14 +443,43 @@ function TiptapWrapper(
     }
   }, [isHydrating]);
 
+  useEffect(() => {
```
thecodrr, line 446: We should be doing this in `EditorChrome` in `components/editor/index.tsx`.

apps/web/src/components/editor/footer.tsx
```diff
@@ -43,6 +44,23 @@ function EditorFooter() {
 
   return (
     <Flex sx={{ alignItems: "center", justifyContent: "center", gap: 2 }}>
+      <button
```
thecodrr, line 47: Use `Button` from `@theme-ui/components`

apps/web/src/components/editor/tiptap.tsx
01zulfi: `editorContainerRef` in `EditorChrome` component is the entire editor, so the zoom property gets applied to the properties bar and the note title as well. I've kept this `useEffect` in `tiptap` for now, please let me know if you have any suggestion

thecodrr: Okay, that makes sense but is there a way to not do this inside a useEffect and directly use the `onWheel` event on the `Flex` below?

01zulfi: I get these errors in the browser console when I use the `onWheel` prop on `Flex`: 

```
Unable to preventDefault inside passive event listener invocation.
```

Currently investigating them...

Although, I've removed the unnecessary `editorConfig.zoom` dependency in the `useEffect`.

apps/web/src/components/editor/common.ts
thecodrr, line 1: Can you add the license header here?

apps/web/src/components/editor/tiptap.tsx
thecodrr: Got it. I think React uses passive event listeners by default so we are forced to use refs.

apps/web/src/components/editor/footer.tsx
```diff
@@ -43,6 +45,42 @@ function EditorFooter() {
 
   return (
     <Flex sx={{ alignItems: "center", justifyContent: "center", gap: 2 }}>
+      <Flex sx={{ alignItems: "center", justifyContent: "center", gap: 1 }}>
+        <Button
```
thecodrr, line 49: Use `variant="icon"`. The button is too unresponsive.

apps/web/src/components/editor/footer.tsx
```diff
@@ -43,6 +45,42 @@ function EditorFooter() {
 
   return (
     <Flex sx={{ alignItems: "center", justifyContent: "center", gap: 2 }}>
+      <Flex sx={{ alignItems: "center", justifyContent: "center", gap: 1 }}>
```
thecodrr, line 48: No need for the `gap`.

apps/web/src/components/editor/footer.tsx
```diff
@@ -43,6 +45,42 @@ function EditorFooter() {
...
+        >
+          <b>-</b>
+        </Button>
+        <Text
+          className="selectable"
+          data-test-id="editor-word-count"
+          variant="subBody"
+          sx={{ color: "paragraph" }}
+        >
+          {editorConfig.zoom} %
```
thecodrr, line 68:
```suggestion
          {editorConfig.zoom}%
```

thecodrr (changes requested):

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -443,14 +444,42 @@ function TiptapWrapper(
...
+        const delta = e.deltaY;
+        if (delta > 0) {
+          editorConfig.zoom += EDITOR_ZOOM.STEP;
+        } else {
+          editorConfig.zoom -= EDITOR_ZOOM.STEP;
+        }
+        const zoom = Math.min(
+          EDITOR_ZOOM.MAX,
+          Math.max(EDITOR_ZOOM.MIN, editorConfig.zoom)
+        );
```
thecodrr, line 462: You shouldn't directly change `editorConfig.zoom`. Also can we slow down the zoom in/out a bit on wheel scroll? It's too fast.

thecodrr (changes requested):

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -443,14 +444,36 @@ function TiptapWrapper(
...
 
+  useEffect(() => {
+    if (!editorContainerRef.current) return;
+
+    const handleWheel = (e: WheelEvent) => {
+      if (e.ctrlKey) {
+        e.preventDefault();
+        const delta = e.deltaY;
+        let zoom = editorConfig.zoom;
+        zoom += Math.round(delta);
```
01zulfi, line 455: @thecodrr I think for the mouse scroll event, we should give user the control on what zoom level they want to be instead of increasing/decreasing by 10 ourselves.

thecodrr: Agreed

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -443,14 +444,36 @@ function TiptapWrapper(
...
+        zoom = Math.min(EDITOR_ZOOM.MAX, Math.max(EDITOR_ZOOM.MIN, zoom));
+        setEditorConfig({ zoom });
```
thecodrr, line 457: I think this can be shortened to something like:
```suggestion
        const zoom = Math.min(EDITOR_ZOOM.MAX, Math.max(EDITOR_ZOOM.MIN, editorConfig.zoom + Math.round(delta));
        setEditorConfig({ zoom });
```

thecodrr (approved):

thecodrr:
Will do some final testing on this locally. Everything looks good though so no more changes will be needed probably.

PR #6849, opened 2024-11-08
web: (wip) better ux when autosave is off

01zulfi:

### DIFF
--- a/apps/web/src/components/editor/action-bar.tsx
+++ b/apps/web/src/components/editor/action-bar.tsx
@@ -44,6 +44,7 @@ import {
 } from "../icons";
 import { ScrollContainer } from "@notesnook/ui";
 import {
+  SaveState,
   SessionType,
   isLockedSession,
   useEditorStore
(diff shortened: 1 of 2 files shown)


apps/web/src/components/editor/tiptap.tsx
```diff
@@ -242,7 +244,23 @@ function TipTap(props: TipTapProps) {
...
+            [
+              {
+                text: "Dismiss",
+                onClick: () => {
+                  hide();
+                }
+              }
+            ],
+            Infinity
+          );
```
01zulfi, line 261: not sure if this is the best place for this toast

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -242,7 +244,23 @@ function TipTap(props: TipTapProps) {
         const ignoreEdit = transaction.getMeta("ignoreEdit") as boolean;
         if (preventSave || !editor.isEditable || !onChange) return;
 
-        if (!autoSave.current) return;
+        if (!autoSave.current) {
+          setEditorSaveState(id, SaveState.NotSaved);
```
thecodrr, line 248: We should call something like `onAutoSaveDisabled` or something and let the upper layer handle this.

apps/web/src/components/editor/manager.ts
```diff
@@ -94,6 +94,8 @@ const [useEditorManager] = createStore<EditorManager>(
 
 export { useEditorManager };
 
+export const editorManager = useEditorManager.getState();
```
thecodrr, line 97: Bad idea. `useEditorManager.getState()` will only give you the current snapshot of the state. Just use `useEditorManager.getState` directly.

apps/web/src/stores/editor-store.ts
```diff
@@ -888,6 +911,25 @@ class EditorStore extends BaseStore<EditorStore> {
...
+            content
+              ? {
+                  content: {
+                    data: content,
+                    type: "tiptap"
+                  }
+                }
+              : {}
+          );
+        }
```
thecodrr, line 931: Good idea to extract this into a function and reuse it.

thecodrr (changes requested):

apps/web/src/components/editor/manager.ts
01zulfi: When I use `useEditorManager.getState` directly, I was getting invalid hook call errors. Also, during my testing, using `editorManager` as I've done worked correctly. Let me explore more on this...

thecodrr: Not sure how you were using it because that error has nothing to do with `useEditorManager.getState`.

01zulfi: My bad, I was calling it like `useEditorManager()`. Fixed.

thecodrr (approved):
Looks good. Will do some testing on this.

apps/web/src/components/editor/index.tsx
```diff
@@ -559,6 +560,23 @@ export function Editor(props: EditorProps) {
...
+        onAutoSaveDisabled={() => {
+          setEditorSaveState(id, SaveState.NotSaved);
+          const { hide } = showToast(
+            "error",
+            "Auto-save is disabled for large notes. Press Ctrl + S to save.",
+            [
+              {
+                text: "Dismiss",
+                onClick: () => {
+                  hide();
```
thecodrr, line 572: The auto save toast appears again on change after dismissing it once. We should store it in memory (not persistent) if user dismisses it.

Secondly, switching notes doesn't get rid of the toast.

apps/web/src/components/editor/action-bar.tsx
```diff
@@ -302,89 +303,95 @@ function TabStrip() {
...
+          renderItem={({ item: session, index: i }) => {
+            const isUnsaved =
+              session.type === "default" &&
+              session.saveState === SaveState.NotSaved;
+            return (
+              <Tab
+                id={session.id}
+                key={session.id}
+                title={
+                  (isUnsaved ? "* " : "") +
```
thecodrr, line 315: Instead of this, let's change the tab icon as that'll be more prominent.

apps/web/src/stores/editor-store.ts
```diff
@@ -869,6 +875,25 @@ class EditorStore extends BaseStore<EditorStore> {
     });
   };
 
+  saveSessionContentIfNotSaved = (sessionId: string) => {
```
thecodrr, line 878: Can we also call this when user moves the app into the background?

thecodrr (changes requested):

thecodrr:
Since we are doing this, let's also add the following tests:

- [ ] Auto save should turn off after user crosses 100K words
- [ ] Ctrl + S should trigger save
- [ ] Switching notes should trigger save
- [ ] Creating a new note should trigger save
- [ ] Closing the note should trigger save

And any other conditions that we can test for.

apps/web/src/components/editor/index.tsx
```diff
@@ -458,6 +468,20 @@ export function Editor(props: EditorProps) {
...
+      if (document.hidden) {
+        saveSessionContentIfNotSaved(id);
+      }
+    }
+
+    window.addEventListener("visibilitychange", handle);
+
+    return () => {
+      window.removeEventListener("visibilitychange", handle);
+    };
```
thecodrr, line 482: Use `onPageVisibilityChanged` from `utils/page-visibility.ts`

thecodrr (changes requested):

apps/web/__e2e__/editor.test.ts
```diff
@@ -84,6 +84,87 @@ test("creating a new note should clear the editor contents & title", async ({
...
+    title: NOTE.title,
+    content
+  });
+
+  await page.locator(getTestId("editor-save-state-notsaved")).waitFor();
+  await page.locator(getTestId("editor-save-state-saved")).waitFor({
+    state: "hidden"
+  });
+
+  page.keyboard.press("Control+s");
```
thecodrr, line 104:
```suggestion
  await page.keyboard.press("Control+s");
```

apps/web/__e2e__/editor.test.ts
```diff
@@ -84,6 +84,87 @@ test("creating a new note should clear the editor contents & title", async ({
...
+  const app = new AppModel(page);
+  await app.goto();
+  const notes = await app.goToNotes();
+  const content = "a ".repeat(100);
+  await notes.createNote({
+    title: NOTE.title,
+    content
+  });
+
+  await page.locator(getTestId("editor-save-state-notsaved")).waitFor();
```
thecodrr, line 99: Let's make getting the save state of the editor a method in `editor-model.ts` since it'll be reused.

apps/web/__e2e__/editor.test.ts
```diff
@@ -84,6 +84,87 @@ test("creating a new note should clear the editor contents & title", async ({
...
+  );
+
+  await page.locator(note1UnsavedTestId).waitFor();
+  await page.locator(note1SavedTestId).waitFor({ state: "hidden" });
+
+  const note2 = await notes.createNote({
+    title: "Test note 2",
+    content: "Test note 2 content"
+  });
+  note2?.openNote();
```
thecodrr, line 137:
```suggestion
  await note2?.openNote();
```

apps/web/__e2e__/editor.test.ts
```diff
@@ -84,6 +84,87 @@ test("creating a new note should clear the editor contents & title", async ({
...
+
+  page.keyboard.press("Control+s");
+
+  await page.locator(getTestId("editor-save-state-notsaved")).waitFor({
+    state: "hidden"
+  });
+  await page.locator(getTestId("editor-save-state-saved")).waitFor();
+});
+
+test("when autosave is disabled, switching to another note should save the note", async ({
```
thecodrr, line 112: This test should go like this:

1. Open app
2. Create 2 notes
3. Switch to note 1 and write until autosave is disabled
4. Switch to note 2
5. See if note 1 gets saved or not

apps/web/__e2e__/notes.test.ts
```diff
@@ -37,6 +38,67 @@ test("create a note", async ({ page }) => {
...
+  const notes = await app.goToNotes();
+  const note = await notes.createNote(NOTE);
+
+  expect(await notes.editor.getContent("text")).toBe(NOTE.content);
+  const noteCloseButton = page.locator(
+    getTestId("note-close-button-" + (await note?.getTitle()))
+  );
+  await noteCloseButton.click();
+  expect(await notes.editor.getContent("text")).toBe("");
+});
```
thecodrr, line 53: What's this? We need to keep the tests related to the `auto saving` functionality. General tests related to app do not belong here.

apps/web/__e2e__/notes.test.ts
```diff
@@ -37,6 +38,67 @@ test("create a note", async ({ page }) => {
...
+
+test("note crosses MAX_AUTO_SAVEABLE_WORDS", async ({ page }) => {
```
thecodrr, line 55:
```suggestion
test("disable auto save when note crosses MAX_AUTO_SAVEABLE_WORDS", async ({ page }) => {
```

apps/web/__e2e__/notes.test.ts
```diff
@@ -37,6 +38,67 @@ test("create a note", async ({ page }) => {
...
+      "Auto-save is disabled for large notes. Press Ctrl + S to save."
+    )
+  ).toBe(true);
+
+  const noteCloseButton = page.locator(
+    getTestId("note-close-button-" + (await note?.getTitle()))
+  );
+  await noteCloseButton.click();
+
+  note?.openNote();
```
thecodrr, line 94:
```suggestion
  await note?.openNote();
```

apps/web/src/components/editor/action-bar.tsx
```diff
@@ -544,7 +556,17 @@ function Tab(props: TabProps) {
           if (e.button == 0) onFocus();
         }}
       >
-        <Icon size={16} color={isActive ? "accent-selected" : "icon"} />
+        <Icon
+          data-test-id={
+            isUnsaved
+              ? "note-icon-unsaved-" + title
+              : "note-icon-saved-" + title
```
thecodrr, line 563: Use id here as title can be the same for notes causing issues.

Also it should be `tab-icon` not `note-icon`

apps/web/src/components/editor/action-bar.tsx
```diff
@@ -590,6 +612,7 @@ function Tab(props: TabProps) {
             }
           }}
           className="closeTabButton"
+          data-test-id={"note-close-button-" + title}
```
thecodrr, line 615: Use `id` here. And change to `tab-close-button`

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -242,7 +246,9 @@ function TipTap(props: TipTapProps) {
         const ignoreEdit = transaction.getMeta("ignoreEdit") as boolean;
         if (preventSave || !editor.isEditable || !onChange) return;
 
-        if (!autoSave.current) return;
+        if (!autoSave.current) {
+          return;
+        }
```
thecodrr, line 251: Revert this change.

thecodrr (changes requested):

PR #6857, opened 2024-11-11
web: fix close tab opening again

01zulfi:
* introduced in #6817

### DIFF
--- a/apps/web/src/components/editor/action-bar.tsx
+++ b/apps/web/src/components/editor/action-bar.tsx
@@ -535,13 +535,15 @@ function Tab(props: TabProps) {
       onAuxClick={(e) => {
         if (e.button == 1) onClose();
       }}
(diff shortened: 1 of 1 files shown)


apps/web/src/components/editor/action-bar.tsx
```diff
@@ -314,6 +314,7 @@ function TabStrip() {
               isActive={session.id === activeSessionId}
               isPinned={!!session.pinned}
               isLocked={isLockedSession(session)}
+              isOnly={sessions.length === 1}
```
thecodrr, line 317: Instead of all this wizardry, let's just move the listeners from the parent `Flex` to the sibling `Flex`

thecodrr (changes requested):

thecodrr (approved):

PR #6866, opened 2024-11-13
editor: fix markdown link pasting (#6639)

### LINKED ISSUE #6639: Pasting in some checklist/link Markdown getting messed up
Hi, I’m a new user, currently using Bear and Standard Notes and evaluating Notesnook.

I’m copying this Markdown text into a note and it is getting mangled:

- [ ] Have Proton apps as web apps (Firefox?): [Installing and uninstalling web apps](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Installing) from [this thread](https://www.reddit.com/r/ProtonMail/comments/1f7mz16/whats_the_purpose_of_the_desktop_apps/)

I’ve tried on the Mac app and iPad app with the same result. It seems to be getting the square brackets missed up, perhaps due to the fact it’s a checkbox?

Thanks.

_______________
**Device information:**
App version: 3.0.18
Platform: ios
Device: Apple-iPad-17.6.1
Pro: false
Logged in: yes

01zulfi:
* don't include empty square brackets in link
* handle multiple markdown links during pasting

Closes #6639

### DIFF
--- a/packages/editor/src/extensions/link/tests/link.test.ts
+++ b/packages/editor/src/extensions/link/tests/link.test.ts
@@ -52,7 +52,7 @@ const input = [
 ];
 
 input.forEach(({ text, expected }, i) => {
-  test(`link regex ${i + 1}`, () => {
+  test(`link regex ${text}`, () => {
     const result = text.match(linkRegex);
     expect(result).toEqual(expected);
   });
--- a/packages/editor/src/extensions/link/tests/link.test.ts
+++ b/packages/editor/src/extensions/link/tests/link.test.ts
@@ -1,0 +1,117 @@
+/*
+This file is part of the Notesnook project (https://notesnook.com/)
+
+Copyright (C) 2023 Streetwriters (Private) Limited
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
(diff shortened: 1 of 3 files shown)


packages/editor/src/extensions/link/tests/link.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    {
+      text: "[NoLink]",
+      expected: null
+    }
+  ];
+
+  input.forEach(({ text, expected }) => {
+    const result = text.match(linkRegex);
+    expect(result).toEqual(expected);
+  });
```
thecodrr, line 58: Instead of looping inside the `test`, move the loop outside and the `test` in. That way we'll have multiple tests and if any one fails, we'll quickly know which one exactly failed.

packages/editor/src/extensions/link/tests/link.test.ts
```diff
@@ -0,0 +1,117 @@
...
+});
+
+describe("paste text", () => {
+  it("with markdown link", async () => {
```
thecodrr, line 62: We always use `test` instead of `it`.

thecodrr (changes requested):

packages/editor/src/extensions/link/tests/link.test.ts
01zulfi: ohh, nice suggestion

packages/editor/src/extensions/link/tests/link.test.ts
01zulfi: I opted to use `it` because in a lot of places where `describe` was used, the test used `it` instead

changing to `test`...

packages/editor/src/extensions/link/tests/link.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    expected: ["[Link](http://example.com)"]
+  },
+  {
+    text: "[NoLink]",
+    expected: null
+  }
+];
+
+input.forEach(({ text, expected }, i) => {
+  test(`link regex ${i + 1}`, () => {
```
thecodrr, line 55:
```suggestion
  test(`link regex ${text}`, () => {
```

thecodrr (approved):

PR #6878, opened 2024-11-15
editor: fix backspace deleting formatting in list item (#6712)

### LINKED ISSUE #6712: Pressing BACKSPACE after SHIFT-ENTER deletes the formatting of the previous line
### What happened?

If you create a bullet or checklist, write some text, press SHIFT-ENTER and then BACKSPACE, the bullet or checklist is removed.

In the following `|` represents the cursor.

If you have:

```
- Random text
  |
```

and you press BACKSPACE, you get:

```
Random text|
```

Rather than:

```
- Random text|
```

1. Create a new note.
2. Type `- ` to create a bullet.
3. Enter some random text.
4. Press SHIFT-ENTER.
5. Press BACKSPACE

### Version

3.0.18-5f0d14f-desktop

### Platform/OS

Linux

### Relevant log output

_No response_

01zulfi:
* fix hitting backspace inside the second (or next) p in list item deleted the formatting of the entire list item

Closes #6712

### DIFF
--- a/packages/editor/patches/@tiptap+extension-list-keymap+2.6.6.patch
+++ b/packages/editor/patches/@tiptap+extension-list-keymap+2.6.6.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@tiptap/extension-list-keymap/dist/index.cjs b/node_modules/@tiptap/extension-list-keymap/dist/index.cjs
-index 0f8ae36..2f715b5 100644
+index 0f8ae36..adeee92 100644
 --- a/node_modules/@tiptap/extension-list-keymap/dist/index.cjs
 +++ b/node_modules/@tiptap/extension-list-keymap/dist/index.cjs
 @@ -82,7 +82,9 @@ const handleBackspace = (editor, name, parentListTypes) => {
@@ -13,8 +13,20 @@ index 0f8ae36..2f715b5 100644
          const $listPos = editor.state.doc.resolve($anchor.before() - 1);
          const listDescendants = [];
          $listPos.node().descendants((node, pos) => {
+@@ -111,6 +113,11 @@ const handleBackspace = (editor, name, parentListTypes) => {
+     if (!listItemPos) {
+         return false;
+     }
++    // if the current position is not at the start of the list item 
++    // then join backward i.e. join within the list item
++    if (listItemPos.$pos.parentOffset !== 0) {
++        return editor.commands.joinBackward();
++    }
+     const $prev = editor.state.doc.resolve(listItemPos.$pos.pos - 2);
+     const prevNode = $prev.node(listItemPos.depth);
+     const previousListItemHasSubList = listItemHasSubList(name, editor.state, prevNode);
 diff --git a/node_modules/@tiptap/extension-list-keymap/dist/index.js b/node_modules/@tiptap/extension-list-keymap/dist/index.js
-index f7ab1e4..8645c44 100644
+index f7ab1e4..6ea03d5 100644
 --- a/node_modules/@tiptap/extension-list-keymap/dist/index.js
 +++ b/node_modules/@tiptap/extension-list-keymap/dist/index.js
 @@ -78,7 +78,9 @@ const handleBackspace = (editor, name, parentListTypes) => {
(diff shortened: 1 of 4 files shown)


packages/editor/src/extensions/list-item/tests/list-tem.test.ts
thecodrr, line 1: Please fix the filename.

packages/editor/patches/@tiptap+extension-list-keymap+2.6.6.patch
```diff
@@ -14,7 +14,7 @@ index 0f8ae36..2f715b5 100644
          const listDescendants = [];
          $listPos.node().descendants((node, pos) => {
 diff --git a/node_modules/@tiptap/extension-list-keymap/dist/index.js b/node_modules/@tiptap/extension-list-keymap/dist/index.js
-index f7ab1e4..8645c44 100644
+index f7ab1e4..f6b996f 100644
 --- a/node_modules/@tiptap/extension-list-keymap/dist/index.js
 +++ b/node_modules/@tiptap/extension-list-keymap/dist/index.js
```
thecodrr, line 19: You also need to make the same changes in `dist/index.cjs`

packages/editor/patches/@tiptap+extension-list-keymap+2.6.6.patch
```diff
@@ -28,3 +28,13 @@ index f7ab1e4..8645c44 100644
          const $listPos = editor.state.doc.resolve($anchor.before() - 1);
          const listDescendants = [];
          $listPos.node().descendants((node, pos) => {
+@@ -103,6 +105,9 @@ const handleBackspace = (editor, name, parentListTypes) => {
+     if (!isAtStartOfNode(editor.state)) {
+         return false;
+     }
++    if (isNodeActive(editor.state, name) && !hasListItemBefore(name, editor.state)) {
```
thecodrr, line 35: I don't think the `isNodeActive` check is needed since it is already being checked in the previous lines.

Also, please add a comment that provides info as to why this change is necessary.

packages/editor/patches/@tiptap+extension-list-keymap+2.6.6.patch
```diff
@@ -13,8 +13,20 @@ index 0f8ae36..2f715b5 100644
          const $listPos = editor.state.doc.resolve($anchor.before() - 1);
          const listDescendants = [];
          $listPos.node().descendants((node, pos) => {
+@@ -107,6 +109,11 @@ const handleBackspace = (editor, name, parentListTypes) => {
+     if (!core.isAtStartOfNode(editor.state)) {
+         return false;
+     }
++    // if previous item is not a list (could be a paragraph in the current list item)
++    // join to the same list item
++    if (!hasListItemBefore(name, editor.state)) {
```
thecodrr, line 22: After some more testing, this is not going to work. It breaks in cases where you are backspacing in the first item of the list. This logic should be more robust to handle only the cases where we are backspacing in the second or later child inside a list item.

I also recommend adding tests for other cases:

1. Backspacing at the start of the first list item
2. Backspacing when there is a list right before the list item

It should handle all these cases correctly.

thecodrr (changes requested):

packages/editor/src/extensions/list-item/tests/list-item.test.ts
```diff
@@ -0,0 +1,138 @@
...
+            wrapperNames: [TaskListNode.name]
+          },
+          {
+            itemName: OutlineListItem.name,
+            wrapperNames: [OutlineList.name]
+          },
+          {
+            itemName: CheckListItem.name,
+            wrapperNames: [CheckList.name]
+          }
```
thecodrr, line 58: These are unnecessary.

packages/editor/src/extensions/list-item/tests/list-item.test.ts
```diff
@@ -0,0 +1,138 @@
...
```
thecodrr, line 93: These are unnecessary.

packages/editor/src/extensions/list-item/tests/list-item.test.ts
```diff
@@ -0,0 +1,138 @@
...
```
thecodrr, line 129: These are unnecessary.

packages/editor/src/extensions/list-item/tests/list-item.test.ts
```diff
@@ -0,0 +1,138 @@
...
+        ]
+      })
+    }
+  });
+  const event = new KeyboardEvent("keydown", { key: "Backspace" });
+  editor.editor.view.dom.dispatchEvent(event);
```
thecodrr, line 64: Add a 100ms sleep after `dispatchEvent` to reduce flakiness in tests.

thecodrr (approved):
Good work!

thecodrr (approved):

PR #6922, opened 2024-11-20
editor: prevent creation of multiple empty notes

luis-411:
This fixes the issue mentioned in #5202 by checking if an empty note already exists and if so, just setting the focus on the existing empty note instead of creating a new one when clicking the add note button.

### DIFF
--- a/apps/web/src/common/index.ts
+++ b/apps/web/src/common/index.ts
@@ -50,7 +50,15 @@ import { ZipFile } from "../utils/streams/zip-stream";
 export const CREATE_BUTTON_MAP = {
   notes: {
     title: strings.addItem("note"),
-    onClick: () => useEditorStore.getState().newSession()
+    onClick: () => {
+      const { sessions, newSession } = useEditorStore.getState();
+      if (sessions.some((session) => "note" in session === false)) {
+        const invalidSession = sessions.find((session) => !("note" in session));
+        useEditorStore.getState().activateSession(invalidSession?.id);
+      } else {
+        newSession();
+      }
+    }
   },
   notebooks: {
     title: strings.addItem("notebook"),
--- a/apps/web/src/common/index.ts
+++ b/apps/web/src/common/index.ts
@@ -50,15 +50,7 @@ import { ZipFile } from "../utils/streams/zip-stream";
 export const CREATE_BUTTON_MAP = {
   notes: {
     title: strings.addItem("note"),
-    onClick: () => {
-      const { sessions, newSession } = useEditorStore.getState();
-      if (sessions.some((session) => "note" in session === false)) {
-        const invalidSession = sessions.find((session) => !("note" in session));
-        useEditorStore.getState().activateSession(invalidSession?.id);
-      } else {
-        newSession();
-      }
-    }
+    onClick: () => useEditorStore.getState().newSession()
   },
   notebooks: {
     title: strings.addItem("notebook"),
--- a/apps/web/src/stores/editor-store.ts
+++ b/apps/web/src/stores/editor-store.ts
@@ -871,12 +871,20 @@ class EditorStore extends BaseStore<EditorStore> {
   };
 
   newSession = () => {
-    this.addSession({
-      type: "new",
-      id: getId(),
-      context: useNoteStore.getState().context,
-      saveState: SaveState.NotSaved
-    });
+    const state = useEditorStore.getState();
+    if (state.sessions.some((session) => session.type === "new")) {
+      const invalidSession = state.sessions.find(
+        (session) => session.type === "new"
+      );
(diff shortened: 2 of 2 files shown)


apps/web/src/common/index.ts
```diff
@@ -50,7 +50,15 @@ import { ZipFile } from "../utils/streams/zip-stream";
 export const CREATE_BUTTON_MAP = {
   notes: {
     title: strings.addItem("note"),
-    onClick: () => useEditorStore.getState().newSession()
+    onClick: () => {
+      const { sessions, newSession } = useEditorStore.getState();
+      if (sessions.some((session) => "note" in session === false)) {
+        const invalidSession = sessions.find((session) => !("note" in session));
+        useEditorStore.getState().activateSession(invalidSession?.id);
```
thecodrr, line 57: Let's handle this inside `newSession` in `editor-store`.

thecodrr (changes requested):

apps/web/src/common/index.ts
luis-411: So you mean like this?

```
  newSession = () => {
    const state = useEditorStore.getState()
    if (state.sessions.some((session) => !("note" in session))) {
      const invalidSession = state.sessions.find((session) => !("note" in session));
      this.activateSession(invalidSession?.id);
    } else {
      this.addSession({
        type: "new",
        id: getId(),
        context: useNoteStore.getState().context,
        saveState: SaveState.NotSaved
      });
    }
  };
```

thecodrr: yep. Although you should fix this too:

```diff
- const invalidSession = state.sessions.find((session) => !("note" in session));
+ const session = state.sessions.find((session) => session.type === "new");
```

apps/web/src/stores/editor-store.ts
```diff
@@ -871,12 +871,20 @@ class EditorStore extends BaseStore<EditorStore> {
...
-      type: "new",
-      id: getId(),
-      context: useNoteStore.getState().context,
-      saveState: SaveState.NotSaved
-    });
+    const state = useEditorStore.getState();
+    if (state.sessions.some((session) => session.type === "new")) {
+      const invalidSession = state.sessions.find(
+        (session) => session.type === "new"
+      );
```
thecodrr, line 878: No need to loop twice over the sessions. Move the `.find` out and just check its return value in the if statement.

thecodrr (changes requested):

apps/web/src/stores/editor-store.ts
thecodrr: Also can we not call it `invalidSession`? There's nothing invalid about it.

luis-411: Like this?

```
  newSession = () => {
    const state = useEditorStore.getState();
    const newSession = state.sessions.find((session) => session.type === "new");
    if (newSession) {
      this.activateSession(newSession?.id);
    } else {
      this.addSession({
        type: "new",
        id: getId(),
        context: useNoteStore.getState().context,
        saveState: SaveState.NotSaved
      });
    }
  };
```

luis-411: Like this?

```
  newSession = () => {
    const state = useEditorStore.getState();
    const session = state.sessions.find((session) => session.type === "new");
    if (session) {
      this.activateSession(session?.id);
    } else {
      this.addSession({
        type: "new",
        id: getId(),
        context: useNoteStore.getState().context,
        saveState: SaveState.NotSaved
      });
    }
  };
```

thecodrr: Yes

thecodrr:
Thank you. This looks good.

thecodrr:
@ammarahm-ed can you add this on the mobile side?

luis-411:
I noticed that when adding a note e.g. in notebook1 and not typing anything yet, but switching into notebook2 and then click add again, the new note would be added to notebook1, because this is where it first war created. This could confuse a lot of people. So now when clicking "+" the context is set to the current context.

apps/web/src/stores/editor-store.ts
```diff
@@ -871,12 +871,19 @@ class EditorStore extends BaseStore<EditorStore> {
...
-      type: "new",
-      id: getId(),
-      context: useNoteStore.getState().context,
-      saveState: SaveState.NotSaved
-    });
+    const state = useEditorStore.getState();
+    const session = state.sessions.find((session) => session.type === "new");
+    if (session) {
+      this.activateSession(session?.id);
+      session.context = useNoteStore.getState().context;
```
thecodrr, line 878:
```suggestion
      session.context = useNoteStore.getState().context;
      this.activateSession(session.id);
```

PR #6935, opened 2024-11-21
web: fix setting item width (#6909)

### LINKED ISSUE #6909: Parameters don't take all the width of the page.
When I go in parameters and check for "backup option with attachments", I don't see the button who is probably out of the width...

**Device information:**
App version: 3.0.20-0f84fab-desktop
OS: OS X 10.15.7
Browser: Electron 31.7.4
Pro: true

01zulfi:
Note: apart from the reported issue, this bug was present in other setting items as well for example in the "Authentication" and "Sync" tabs.

Closes #6909

### DIFF
--- a/apps/web/src/dialogs/settings/index.tsx
+++ b/apps/web/src/dialogs/settings/index.tsx
@@ -466,7 +466,7 @@ function SettingItem(props: { item: Setting }) {
           gap: 4
         }}
       >
-        <Flex sx={{ flexDirection: "column", flexShrink: 0 }}>
+        <Flex sx={{ flexDirection: "column", flexBasis: "85%" }}>
           <Text variant={"subtitle"}>{item.title}</Text>
           {item.description && (
             <Text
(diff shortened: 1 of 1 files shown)


apps/web/src/dialogs/settings/index.tsx
```diff
@@ -466,7 +466,7 @@ function SettingItem(props: { item: Setting }) {
           gap: 4
         }}
       >
-        <Flex sx={{ flexDirection: "column", flexShrink: 0 }}>
+        <Flex sx={{ flexDirection: "column", flexBasis: "85%" }}>
```
thecodrr:
```suggestion
        <Flex sx={{ flexDirection: "column", flex: 1 }}>
```

thecodrr (commented):
`flexBasis` fixes the width of the element which is not what we want because in some items the buttons need more than 15% of the space.

In any case, I am merging this since its required for the release.

PR #6953, opened 2024-11-22
web: always show desktop backup directory path (#6867)

### LINKED ISSUE #6867: Current path to the backup directory is not shown
This is only a minor issue, more of a suggestion. At the moment you cannot see where your backups are saved if you simply browse "Backup & export" section of the settings. It takes extra steps if you want to make sure your backups are synced to the cloud, for example. 

As a potential solution I suggest adding current path to the backups directory to the "Backup directory" sub-section of the "Backup & export" section of the Settings. Third line in the section probably. And a notice that a path is not set when it is not set. 

**Device information:**
App version: 3.0.19-53a1e2c-desktop
OS: Linux 64-bit
Browser: Electron 30.5.1
Pro: true

01zulfi:

Closes #6867

### DIFF
--- a/apps/web/src/dialogs/settings/backup-export-settings.ts
+++ b/apps/web/src/dialogs/settings/backup-export-settings.ts
@@ -28,6 +28,9 @@ import { useStore as useUserStore } from "../../stores/user-store";
 import { desktop } from "../../common/desktop-bridge";
 import { PATHS } from "@notesnook/desktop";
 
+const getDesktopBackupsDirectoryPath = () =>
+  useSettingStore.getState().backupStorageLocation || PATHS.backupsDirectory;
+
 export const BackupExportSettings: SettingsGroup[] = [
   {
     key: "backup",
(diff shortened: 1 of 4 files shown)


apps/web/src/dialogs/settings/backup-export-settings.ts
```diff
@@ -164,7 +167,10 @@ export const BackupExportSettings: SettingsGroup[] = [
       {
         key: "backup-directory",
         title: strings.selectBackupDir(),
-        description: strings.selectBackupDirDesc(),
+        description: () =>
+          strings
+            .selectBackupDirDesc(getDesktopBackupsDirectoryPath())
```
thecodrr, line 172: This won't update when user changes their backup directory path.

thecodrr (changes requested):

apps/web/src/dialogs/settings/backup-export-settings.ts
01zulfi: It is updating for me. When I change the backup directory path, the path text shown in the description also changes. Is there something else you're referring to here?

thecodrr: Let me see.

thecodrr: It's working.

PR #6987, opened 2024-11-24
monograph: add monograph response charset header

dyw770:
when no charset is specified, Chinese notes will be garbled. Add default charset.


thecodrr (approved):

thecodrr:
Thanks!

PR #6990, opened 2024-11-25
web: Fix inconsistent zoom amount when adjusting with scroll wheel.

Leviob:
Zoom steps are always in increments of 10 when using a mouse wheel or trackpad.

Running `npm run prettier` as directed in the contributing guidelines modified many files, and I included them in this PR. Should I not run this command?


thecodrr:
@Leviob thanks for this! Yes, you should probably exclude those files. Maybe just run prettier on the file you changed.

Leviob:
@thecodrr, Thank you. I'll do that from now on. This (much more realistic) PR is ready for review!

PR #6993, opened 2024-11-25
monograph: remove callout headings from toc (#6875)

### LINKED ISSUE #6875: Callouts break Monograph Table of Contents
### What happened?

Image

Add any callout and publish monograph.

Example: https://monogr.ph/663a0adf06e91ea603c31037/

### Version

Latest monograph

### Platform/OS

Linux, Chrome/Chromium, Firefox, Android

### Relevant log output

```shell

```

01zulfi:

Closes #6875

### DIFF
--- a/apps/monograph/app/components/monographpost/index.tsx
+++ b/apps/monograph/app/components/monographpost/index.tsx
@@ -73,6 +73,11 @@ function generateTableOfContents() {
   let currentHeading = 0;
 
   for (const heading of headings) {
+    const isCalloutHeading = heading.closest(".callout");
+    if (isCalloutHeading) {
+      continue;
+    }
+
     const text = heading.textContent || "<empty>";
     const nodeName = heading.nodeName;
     const headingLevel = levelsMap[nodeName];


01zulfi:
Are you talking about this specific approach or the code changes? Did you have something else in mind?

thecodrr:
@01zulfi I am talking about the changes _and_ the approach. I don't think I can come up with a more complicated way to do this.

apps/monograph/app/components/monographpost/index.tsx
```diff
@@ -90,6 +90,13 @@ function generateTableOfContents() {
         : level;
     currentHeading = headingLevel;
 
+    const isCalloutHeading =
+      !heading.previousSibling &&
+      heading.parentElement?.classList.contains("callout");
+    if (isCalloutHeading) {
+      continue;
+    }
+
```
01zulfi, line 99: Please let me know if this can be improved

apps/monograph/app/components/monographpost/index.tsx
```diff
@@ -90,6 +90,13 @@ function generateTableOfContents() {
         : level;
     currentHeading = headingLevel;
 
+    const isCalloutHeading =
+      !heading.previousSibling &&
```
thecodrr, line 94: Why this check?

apps/monograph/app/components/monographpost/index.tsx
```diff
@@ -90,6 +90,13 @@ function generateTableOfContents() {
...
+      heading.parentElement?.classList.contains("callout");
+    if (isCalloutHeading) {
```
thecodrr, line 96: You should do this at the very start of the loop.

apps/monograph/app/components/monographpost/index.tsx
```diff
@@ -90,6 +90,13 @@ function generateTableOfContents() {
         : level;
     currentHeading = headingLevel;
 
+    const isCalloutHeading =
+      !heading.previousSibling &&
+      heading.parentElement?.classList.contains("callout");
```
thecodrr, line 95: Why not do `heading.closest('.callout');`?

Read more about `closest` here: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

thecodrr (changes requested):

apps/monograph/app/components/monographpost/index.tsx
01zulfi: If I don't use this check, the user defined headings inside callout aren't added to the toc

thecodrr: We don't want any headings inside a callout to appear in TOC.

apps/monograph/app/components/monographpost/index.tsx
01zulfi: If we move this to the start of the loop, heading levels aren't respected for headings inside callout. 

Check at start of loop: 
image


Otherwise:
image

apps/monograph/app/components/monographpost/index.tsx
01zulfi: ah, got it. I misunderstood then.

apps/monograph/app/components/monographpost/index.tsx
01zulfi: ^ ignore

apps/monograph/app/components/monographpost/index.tsx
```diff
@@ -90,6 +90,11 @@ function generateTableOfContents() {
         : level;
     currentHeading = headingLevel;
 
+    const isCalloutHeading = heading.closest(".callout");
+    if (isCalloutHeading) {
```
thecodrr, line 94: This should be at the start of the loop.

thecodrr (changes requested):

thecodrr (approved):

PR #7009, opened 2024-11-27
web: round off zoom value

01zulfi:

### DIFF
--- a/apps/web/src/components/editor/tiptap.tsx
+++ b/apps/web/src/components/editor/tiptap.tsx
@@ -460,7 +460,7 @@ function TiptapWrapper(
             : EDITOR_ZOOM.STEP;
         const zoom = Math.min(
           EDITOR_ZOOM.MAX,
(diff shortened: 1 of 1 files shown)


apps/web/src/components/editor/tiptap.tsx
```diff
@@ -460,7 +460,7 @@ function TiptapWrapper(
             : EDITOR_ZOOM.STEP;
         const zoom = Math.min(
           EDITOR_ZOOM.MAX,
-          Math.max(EDITOR_ZOOM.MIN, editorConfig.zoom + delta)
+          Math.max(EDITOR_ZOOM.MIN, Math.round(editorConfig.zoom + delta))
```
01zulfi, line 463: I'm including the `editorConfig.zoom` value in `Math.round` instead of just `delta`, because users who have faced this bug, might not be able to fix zoom value unless they clear local storage. Please let me know if this is okay.

thecodrr: Good idea.

thecodrr (approved):

PR #7018, opened 2024-11-28
editor: hide edit button when in read only mode

luis-411:
I noticed that when a note is in read only mode, the `edit link` button is still showing up and the link can be edited.
I added the same line to the `EditLink` function as in `RemoveLink`


thecodrr (approved):

thecodrr:
Thanks!

PR #7060, opened 2024-12-05
editor: added support for default text direction

luis-411:
Addresses: #4654 

Added option in editor settings to enable different text directions by default

### DIFF
--- a/packages/intl/src/strings.ts
+++ b/packages/intl/src/strings.ts
@@ -1110,6 +1110,9 @@ $headline$: Use starting line of the note as title.`,
   titleFormatDesc: () => t`Choose how the new note titles are formatted`,
   mardownShortcuts: () => t`Markdown shortcuts`,
   mardownShortcutsDesc: () => t`Use markdown shortcuts in the editor`,
+  textDirection: () => t`Text direction`,
+  textDirectionDesc: () =>
+    t`Change default text direction to "left-to-right" or "right-to-left"`,
   privacyAndSecurity: () => t`Privacy & security`,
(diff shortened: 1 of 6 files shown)


packages/intl/src/strings.ts
```diff
@@ -1110,6 +1110,9 @@ $headline$: Use starting line of the note as title.`,
   titleFormatDesc: () => t`Choose how the new note titles are formatted`,
   mardownShortcuts: () => t`Markdown shortcuts`,
   mardownShortcutsDesc: () => t`Use markdown shortcuts in the editor`,
+  textDirection: () => t`Text direction`,
+  textDirectionDesc: () =>
+    t`Change default text direction to "left-to-right" or "right-to-left"`,
```
thecodrr, line 1115:
```suggestion
    t`Change default text direction to LTR or RTL`,
```

thecodrr:
This won't work because changing the default text direction also changes the direction of all the paragraphs old and new. This is probably not the expected behavior. We want the text written when default text direction is RTL to remain RTL even if we switch to LTR later on.

luis-411:
This approach seems promising, but it does not work yet, because when performing this:
`editorContainer.style.direction = useEditorStore.getState().getActiveSession()?.textDirection || "ltr";`
the value of `useEditorStore.getState().getActiveSession()?.textDirection` is always undefined.
Other values like `title` are undefined as well. I think I have to maybe use a different function to get the value correctly. Do you have an idea of what is wrong here?

01zulfi:
closing in favor of https://github.com/streetwriters/notesnook/pull/9238

PR #7061, opened 2024-12-05
web: fix tags editor not showing in locked note

01zulfi:

### DIFF
--- a/apps/web/src/components/editor/index.tsx
+++ b/apps/web/src/components/editor/index.tsx
@@ -834,6 +834,7 @@ function UnlockNoteView(props: UnlockNoteViewProps) {
           if (!note || !note.content)
             throw new Error("note with this id does not exist.");
 
+          const tags = await db.notes.tags(note.id);
           useEditorStore.getState().addSession({
             type: session.note.readonly ? "readonly" : "default",
             locked: true,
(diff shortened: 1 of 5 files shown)


apps/web/src/components/editor/index.tsx
```diff
@@ -841,6 +841,7 @@ function UnlockNoteView(props: UnlockNoteViewProps) {
             note: session.note,
             saveState: SaveState.Saved,
             sessionId: `${Date.now()}`,
+            tags: session.tags,
```
thecodrr, line 844: Fetch the note's tags here directly instead of passing it through the locked session.

thecodrr (changes requested):

packages/core/src/collections/notes.ts
```diff
@@ -176,6 +176,15 @@ export class Notes implements ICollection {
...
 
+  async tags(id: string) {
+    return this.db.relations
+      .to({ id, type: "note" }, "tag")
+      .selector.items(undefined, {
+        sortBy: "dateCreated",
+        sortDirection: "asc"
+      });
+  }
+
```
01zulfi, line 187: I've moved the `getTags` function here, do let me know if this is okay or should I revert it

thecodrr: Good idea.

thecodrr (approved):

PR #7078, opened 2024-12-09
editor: add alt arrow keyboard shortcut

### LINKED ISSUE #3908: Keyboard shortcut to move line/paragraph up or down
One of the Editor Shortcuts I love, is to be able to ... 

"MOVE Current Line/Paragraph UP or DOWN"

It is so handy, productive and time efficient.

I realize it would be only for Desktop versions of Notesnook, but that is where bulk of notes are edited, prior to using on Mobile.

Try in following products
(UpArrow or DownArrow)
--------------------------------------------------
Visual Studio     ==>Alt+UpArrow
MS Word            ==>Sift+Alt+UpArrow
MS PowerPoint ==>Sift+Alt+UpArrow
UltraEdit             ==>(Self Define)

Thank You!

**Device information:**
App version: 2.6.12-4d7a8b9-desktop
OS: Windows 10 64-bit
Browser: Electron 25.9.3
Pro: true

01zulfi:
* implementation from https://discuss.prosemirror.net/t/keymap-to-move-a-line/3645/5
* only supports moving paragraphs
* TODO: add tests

Closes #3908

### DIFF
--- a/packages/editor/src/extensions/key-map/key-map.ts
+++ b/packages/editor/src/extensions/key-map/key-map.ts
@@ -17,11 +17,92 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { Extension } from "@tiptap/core";
+import { Editor, Extension } from "@tiptap/core";
 import { isInTable } from "@tiptap/pm/tables";
 import { CodeBlock } from "../code-block/index.js";
 import { showLinkPopup } from "../../toolbar/popups/link-popup.js";
 import { isListActive } from "../../utils/list.js";
+import { findParentNodeOfType } from "../../utils/prosemirror.js";
+import { Fragment, Node, Slice } from "@tiptap/pm/model";
+import { ReplaceStep } from "@tiptap/pm/transform";
+import { Selection } from "@tiptap/pm/state";
+import { Callout } from "../callout/callout.js";
+
+function mapChildren<T>(
+  node: Node | Fragment,
+  callback: (child: Node, index: number, frag: Fragment) => T
+): T[] {
+  const array = [];
+  for (let i = 0; i < node.childCount; i++) {
+    array.push(
+      callback(node.child(i), i, node instanceof Fragment ? node : node.content)
+    );
+  }
+  return array;
+}
+
+/**
+ * implementation from https://discuss.prosemirror.net/t/keymap-to-move-a-line/3645/5
+ */
+function moveNode(editor: Editor, dir: "up" | "down") {
+  const isDown = dir === "down";
+  const { state } = editor;
+  if (!state.selection.empty) {
+    return false;
+  }
+
+  const { $from } = state.selection;
+  const type = $from.node().type;
+  if (!type) {
+    return false;
+  }
+
+  const currentResolved = findParentNodeOfType(type)(state.selection);
+  if (!currentResolved) {
+    return false;
+  }
+
+  const { node: currentNode } = currentResolved;
+  const parentDepth = currentResolved.depth - 1;
+  const parent = $from.node(parentDepth);
+  const parentPos = $from.start(parentDepth);
+
+  if (currentNode.type !== type) {
+    return false;
+  }
+
+  let arr = mapChildren(parent, (node) => node);
+  let index = arr.indexOf(currentNode);
+  let swapWith = isDown ? index + 1 : index - 1;
+  if (swapWith >= arr.length || swapWith < 0) {
+    return false;
+  }
+  if (swapWith === 0 && parent.type.name === Callout.name) {
+    return false;
+  }
+
+  const swapWithNodeSize = arr[swapWith].nodeSize;
+
+  [arr[index], arr[swapWith]] = [arr[swapWith], arr[index]];
+
+  let tr = state.tr;
+  let replaceStart = parentPos;
+  let replaceEnd = $from.end(parentDepth);
+
+  const slice = new Slice(Fragment.fromArray(arr), 0, 0);
+
+  tr = tr.step(new ReplaceStep(replaceStart, replaceEnd, slice, false));
+  tr = tr.setSelection(
+    Selection.near(
+      tr.doc.resolve(
+        isDown ? $from.pos + swapWithNodeSize : $from.pos - swapWithNodeSize
+      )
+    )
+  );
+  tr.scrollIntoView();
+  editor.view.dispatch(tr);
+  return true;
+}
 
 export const KeyMap = Extension.create({
   name: "key-map",
@@ -70,6 +151,12 @@ export const KeyMap = Extension.create({
       "Mod-k": ({ editor }) => {
         showLinkPopup(editor);
         return true;
+      },
+      "Alt-ArrowUp": ({ editor }) => {
+        return moveNode(editor, "up");
+      },
+      "Alt-ArrowDown": ({ editor }) => {
+        return moveNode(editor, "down");
       }
     };
   }


packages/editor/src/extensions/key-map/key-map.ts
```diff
@@ -71,7 +76,122 @@ export const KeyMap = Extension.create({
...
+  }
+
+  const { node: currentNode } = currentResolved;
+  const parentDepth = currentResolved.depth - 1;
+  const parent = $from.node(parentDepth);
+  const parentPos = $from.start(parentDepth);
+
+  if (currentNode.type !== targetType) {
+    return false;
+  }
```
thecodrr: What is all this for? Why not just move the parent node up and down? Why do we need to specifically handle these?

01zulfi: This is for moving individual list items up and down within the list. I thought this would be a nice addition

thecodrr: What's the difference? Aren't they just nodes?

01zulfi: For moving list items, we need to select the list item instead of paragraph. 

`const type = $from.node().type;` this line was selecting the paragraph in the list item instead of the list item

this snippet fixes it:

```ts
  if (isListActive(editor)) {
    targetType = $from.node($from.depth - 1).type;
    currentResolved = findParentNodeOfType(targetType)(state.selection);
  }
  ```

Comment (thecodrr) [reply packages/editor/src/extensions/key-map/key-map.ts]: What if user is inside a callout/blockquote that's in a list item?

File: packages/editor/src/extensions/key-map/key-map.ts [hunk 2]
```diff
@@ -71,7 +86,181 @@ export const KeyMap = Extension.create({
...
+
+/**
+ * implementation inspired from https://discuss.prosemirror.net/t/keymap-to-move-a-line/3645/5
+ * @param mode - "itself" | "parent"  - whether to move the current node itself or its immediate valid parent node
+ */
+function moveNode(
+  editor: Editor,
+  dir: "up" | "down",
+  mode: "itself" | "parent"
+) {
```
thecodrr: This is a clear indication of a code smell. This function should be refactored into parts:

- moveNodeUp
- moveNodeDown
- moveParentUp
- moveParentDown

If there's anything common between these, you can move it to a separate function. Common does not mean passing conditionals as parameters and using `if` conditions all over the function.

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+    // move the entire list item
+    if (isFirstParagraph) {
+      targetType = $from.node($from.depth - 1).type;
+      if (
+        targetType.name === Callout.name ||
+        targetType.name === Blockquote.name
+      ) {
+        targetType = $from.node($from.depth - 2).type;
+      }
+    }
```
thecodrr, line 88: This is very fragile. It just assumes that list item is the first or second parent but that won't always be the case i.e. what if user has nested callouts or blockquotes?

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+  return array;
+}
+
+function resolveNode(editor: Editor) {
+  const { state } = editor;
+  const { $from } = state.selection;
+
+  let targetType = $from.node().type;
+  let currentResolved = findParentNodeOfType(targetType)(state.selection);
```
thecodrr: This looks unnecessary. You don't need to do `findParentNodeOfType` when you are literally standing on the same node.

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  if (isListActive(editor)) {
+    const currentNode = $from.node();
+    const parentNode = $from.node($from.depth - 1);
+    const isFirstParagraph =
+      currentNode.type.name === "paragraph" &&
+      parentNode.firstChild === currentNode;
```
thecodrr, line 83: Why not just check if `parentNode.firstChild.type.name === "paragraph"`?

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    currentResolved = findParentNodeOfType(targetType)(state.selection);
+  }
+
+  if (
+    findParentNodeClosestToPos($from, (node) => node.type.name === Callout.name)
+  ) {
+    const currentNode = $from.node();
+    const parentNode = $from.node($from.depth - 1);
```
thecodrr: The check inside the `if` statement and these 2 statements below it are doing very different things. The `callout` could be any parent from the current selection - not exactly the parent right above it.

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+    const isFirstHeading =
+      currentNode.type.name === "heading" &&
+      parentNode.firstChild === currentNode;
+
+    // move the entire callout
+    if (isFirstHeading) {
+      targetType = $from.node($from.depth - 1).type;
```
thecodrr: Again, we can't assume that the callout is the parent.

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+  const validParents = [
+    Callout.name,
+    Table.name,
+    BulletList.name,
+    OrderedList.name,
+    TaskListNode.name,
+    CheckList.name,
+    OutlineList.name,
+    Blockquote.name
+  ];
```
thecodrr: This should be outside the function.

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  const parent = findParentNodeClosestToPos($from, (node) =>
+    validParents.includes(node.type.name)
+  );
+
+  if (!parent) return undefined;
+
+  const targetType = parent.node.type;
+  return findParentNodeOfType(targetType)(state.selection);
```
thecodrr: What is this? Why not just return `parent` directly?

packages/editor/src/extensions/key-map/move-node.ts
```diff
@@ -0,0 +1,215 @@
...
+): boolean {
+  const { state } = editor;
+  const { $from } = state.selection;
+
+  const parentDepth = resolvedNode.depth - 1;
+  const parent = $from.node(parentDepth);
+  const parentPos = $from.start(parentDepth);
+
+  let arr = mapChildren(parent, (node) => node);
+  let index = arr.indexOf(resolvedNode.node);
```
thecodrr, line 146: Why do you need to map all the children into an array in order to find the index of the child?

packages/editor/src/extensions/key-map/move-node.ts
01zulfi: I'm still trying to remember why I have the callout and blockquote specific checks inside `isFirstParagraph` check. We don't need that. I've removed that, and am also making sure that the `parentNode` is a list item before attempting to move the entire list item

packages/editor/src/extensions/key-map/move-node.ts
01zulfi: We need to determine if  `parentNode.firstChild` is the node we're currently on

packages/editor/src/extensions/key-map/move-node.ts
01zulfi: This `arr` is used for swapping, so we need to find the index and the corresponding swapping index

thecodrr:
@01zulfi did you push the new changes?

01zulfi:
my bad, I accidently pushed into NN's branch instead of my fork

PR #7081, opened 2024-12-10
web: new sidebar ui

01zulfi:


thecodrr:
Just gave this a test run. A few bugs:

1. The middle pane is still navigating when switching to Notebooks/Tags in the sidebar
2. Adding a subnotebook when a root notebook is collapsed doesn't properly expand the root notebook in the tree
3. Adding a new notebook at the root level is not possible after adding the first notebook
4. Adding a new tag is not possible from the sidebar

apps/web/src/components/navigation-menu/index.tsx
```diff
@@ -151,26 +165,55 @@ function NavigationMenu(props: NavigationMenuProps) {
...
+              return { ...route, count: totalMonographs };
+            default:
+              return route;
+          }
+        });
+      });
+    };
+
+    setCounts();
+  }, [notes]);
```
01zulfi, line 205: Don't think this is the best way to get and update counts for the home tab

thecodrr: This should be okay for now.

apps/web/src/components/navigation-menu/index.tsx
```diff
@@ -226,6 +266,69 @@ function NavigationMenu(props: NavigationMenuProps) {
...
+            key={tab.id}
+            id={tab.id}
+            isTablet={isTablet}
+            title={tab.title}
+            icon={tab.icon}
+            selected={currentTab === tab.id}
+            onClick={() => setCurrentTab(tab.id)}
+            showTitle={false}
+          />
+        ))}
```
thecodrr, line 323: I think we should create a separate component for tab items instead of reusing `NavigationItem`.

apps/web/src/components/route-container/index.tsx
```diff
@@ -135,13 +135,15 @@ function Header(props: RouteContainerProps) {
           />
         ) : (
           <Menu
-            onClick={() => toggleSideMenu(true)}
+            onClick={() => {
+              toggleSideMenu(true);
+            }}
```
thecodrr, line 140:
```suggestion
            onClick={() => toggleSideMenu(true)}
```

apps/web/src/components/virtualized-tree/index.tsx
```diff
@@ -85,8 +85,15 @@ export function VirtualizedTree<T>(props: TreeViewProps<T>) {
...
-        setNodes(children);
+        // const children = await getChildNodes(rootId, -1);
+        // setNodes(children);
+        const nodes = await fetchChildren(
+          rootId,
+          -1,
+          expandedIds,
+          getChildNodes
+        );
+        setNodes(nodes);
```
thecodrr, line 96: Why this change?

apps/web/src/navigation/routes.tsx
```diff
@@ -153,6 +149,9 @@ const routes = defineRoutes({
       title: strings.routes.Tags(),
       type: "tags",
       component: Tags,
+      props: {
+        location: "middle-pane"
+      },
```
thecodrr, line 154: Is this used?

apps/web/src/views/tags.tsx
```diff
@@ -24,19 +24,33 @@ import Placeholder from "../components/placeholders";
...
   );
 
+  useEffect(() => {
+    if (location === "sidebar") return;
+    tags?.item(0).then((item) => {
+      if (item && item?.item) {
+        navigate(`/tags/${item.item.id}`);
+      }
+    });
+  }, [tags, location]);
```
thecodrr, line 49: I don't think this is needed.

thecodrr (changes requested):
Good work overall but since this contains _a lot_ of changes, I'll have to review it chunk by chunk.

thecodrr:
You will also have to run a rebase with `master`.

apps/web/src/components/virtualized-tree/index.tsx
01zulfi: Whenever the tree was `refresh`ed, the children of expanded notebooks weren't loaded. This change fixed that

apps/web/src/navigation/routes.tsx
01zulfi: nope, fixed

apps/web/src/views/tags.tsx
01zulfi: This is in case a user manually navigates to `app.notesnook.com/tags`, we can redirect them to the first tag. Plus this also prevents showing the tags listing in the middle pane.

Issue #6932, opened 2024-11-20
The select menu while selecting programming language is deformed


ifajul89:
### What happened?

After writing the code by default, it remains in Plaintext. When I tried to open the selection box of programming languages, it showed me a kind of deformed or the names of programming languages, not having proper padding.
Issue Image: https://drive.proton.me/urls/T61AHEQ3MM#33X2ZpsbB9LC

### Steps to reproduce the problem

Just add paddings to each programming language name while selecting

### Version

3.0.20-0f84fab-desktop

### Platform/OS

macOS

### Relevant log output

```shell

```

ammarahm-ed:
This has been fixed.

PR #7086, opened 2024-12-11
editor: add code block language to local storage

### LINKED ISSUE #7032: Set default programming language for Code Block

Hi! 

I use Code Blocks feature a lot in my notes. Mainly for my Kotlin study.
But it's getting tedious because everytime I insert a code block, I should set the language from "Plaintext" to "Kotlin". 

I reckon it would be seamless if either:
A. Notesnook can remember the last programming language that I used, then use it as the language for the new code block
B. I can set default programming language manually for code block

_No response_

### Additional context

_No response_

01zulfi:
* also fix spacing issues in language selector list

Closes #7032 #6932

### DIFF
--- a/packages/editor/src/extensions/code-block/component.tsx
+++ b/packages/editor/src/extensions/code-block/component.tsx
@@ -18,7 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { Flex, Input, Text } from "@theme-ui/components";
-import { useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Button } from "../../components/button.js";
 import { ResponsivePresenter } from "../../components/responsive/index.js";
 import { useTimer } from "../../hooks/use-timer.js";
@@ -30,6 +30,7 @@ import { CodeBlockAttributes } from "./code-block.js";
 import Languages from "./languages.json";
 import { useThemeEngineStore } from "@notesnook/theme";
 import { strings } from "@notesnook/intl";
+import { config } from "../../utils/config.js";
 
 export function CodeblockComponent(
   props: ReactNodeViewProps<CodeBlockAttributes>
@@ -48,6 +49,23 @@ export function CodeblockComponent(
   const languageDefinition = Languages.find(
     (l) => l.filename === language || l.alias?.some((a) => a === language)
   );
+
+  const cachedLanguage =
+    config.get<(typeof Languages)[number]>("codeBlockLanguage");
+
+  useEffect(() => {
+    if (languageDefinition) {
+      config.set("codeBlockLanguage", languageDefinition);
+      return;
+    }
+
+    if (!languageDefinition && cachedLanguage && cachedLanguage.filename) {
+      updateAttributes(
+        { language: cachedLanguage.filename },
+        { addToHistory: false, preventUpdate: false }
+      );
+    }
+  }, []);
 
   return (
     <>
@@ -155,7 +173,9 @@ export function CodeblockComponent(
             title={strings.changeLanguage()}
           >
             <Text variant={"subBody"} spellCheck={false}>
-              {languageDefinition?.title || "Plaintext"}
+              {languageDefinition?.title ||
+                cachedLanguage?.title ||
+                "Plaintext"}
             </Text>
           </Button>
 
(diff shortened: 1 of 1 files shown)


packages/editor/src/extensions/code-block/component.tsx
```diff
@@ -49,6 +50,23 @@ export function CodeblockComponent(
...
+      return;
+    }
+
+    if (!languageDefinition && cachedLanguage && cachedLanguage.filename) {
+      updateAttributes(
+        { language: cachedLanguage.filename },
+        { addToHistory: false, preventUpdate: false }
+      );
+    }
+  }, []);
```
thecodrr, line 68: We shouldn't do this here. This will run for all codeblocks even though we just want to set the language once when a new codeblock is added. It'll also cause a flash where a codeblock is plaintext and then switches to the last used language.

Instead let's do this in `code-block.ts` in the `parseHTML` function of the `language` attribute. Not sure if it'll work or not though so give it a try.

thecodrr (changes requested):

packages/editor/src/extensions/code-block/component.tsx
01zulfi: Unfortunately, `useEffect` is called for all codeblocks when the page loads. Also, I wasn't able to notice a flash when codeblock is plaintext and then switches to the local storage language. But, there is a minor bug in this PR: user won't ever be able to go back to plaintext codeblock until they delete the local storage entry. I think we should add plaintext as a language in `languages.json`. What do you think?

> Instead let's do this in code-block.ts in the parseHTML function of the language attribute. Not sure if it'll work or not though so give it a try.

I tried in `parseHTML` of the `language` attribute. The issue here is the `parseHTML` doesn't get called when a new code block is added, so we can't save the code block language if user pasted a code block from VSCode for example. I'll continue trying with `parseHTML`.

01zulfi: Thanks for the comemnt, I realized we don't need the `useEffect` at all. Please review the latest changes.

(plus I've added Plaintext in languages, let me know if we should keep this or else I can revert)

packages/editor/src/extensions/code-block/code-block.ts
```diff
@@ -463,6 +464,17 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
...
 
+            if (language) {
+              const languageDefinition = Languages.find(
+                (l) =>
+                  l.filename === language ||
+                  l.alias?.some((a) => a === language)
+              );
+              if (languageDefinition) {
+                config.set("codeBlockLanguage", languageDefinition);
+              }
```
thecodrr, line 475: I don't think we should change the saved code block language when a user pastes some code. It should only be done when the user intentionally selects a different language from the language selector.

thecodrr (changes requested):
I don't think this is working. It changes the language label in the code block status bar but no highlighting actually happens.

packages/editor/src/extensions/code-block/code-block.ts
```diff
@@ -536,6 +537,15 @@ export const CodeBlock = Node.create<CodeBlockOptions>({
...
             l.alias?.some((a) => a === node.attrs.language)
         );
+
+        const cachedLanguage =
+          config.get<(typeof Languages)[number]>("codeBlockLanguage");
+        if (!languageDefinition && cachedLanguage && cachedLanguage.filename) {
+          this.editor.commands.updateAttributes(this.type, {
+            language: cachedLanguage.filename
+          });
+        }
```
01zulfi, line 547: Thanks again for the thorough review!

I've added the language update from local storage logic here and it works. Please let me know if this is okay. Other place this exact logic works is in the `useEffect` inside the code block component.

thecodrr: I just checked again and we can reuse `defaultLanguage` in the `Highlighter` to support this. It can be a function since the value can change.

However, the default language doesn't get stored in the code block attributes which you'll need to add support for in `appendTransaction`.

01zulfi: Done, please review

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -307,6 +314,11 @@ function updateSelection(
       isDocChanged ? toCodeLines(node.textContent, pos) : undefined
     );
     attributes.caretPosition = position;
+    console.log({
+      attrs: node.attrs,
+      defaultLanguage: defaultLanguage()
+    });
```
thecodrr, line 320:
```suggestion
```

packages/editor/src/extensions/code-block/highlighter.ts
```diff
@@ -307,6 +314,11 @@ function updateSelection(
...
+    attributes.language = node.attrs.language ?? defaultLanguage();
```
thecodrr, line 321: This works but I think it'll cause issues with old code blocks (added before this change) as they'll get their language changed to the cached language as soon as the user makes a selection or a change in the document.

We should find a way to only run this once when the code block is added in the document. I think we can compare `oldState` and `newState` to find that out.

thecodrr (changes requested):

PR #7097, opened 2024-12-13
editor: added link popup description

luis-411:
The popup that appears when you want to edit a link, seemed a bit unintuitive, as you only saw the link in an input field twice:

image

Only the placeholders of the inputs suggested what is meant, so I added labels describing the fields:

image

### DIFF
--- a/packages/editor/src/toolbar/popups/link-popup.tsx
+++ b/packages/editor/src/toolbar/popups/link-popup.tsx
@@ -17,7 +17,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { Input } from "@theme-ui/components";
+import { Input, Text } from "@theme-ui/components";
 import { Flex } from "@theme-ui/components";
 import { useRefValue } from "../../hooks/use-ref-value.js";
 import { Popup } from "../components/popup.js";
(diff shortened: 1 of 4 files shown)


packages/editor/src/toolbar/popups/link-popup.tsx
```diff
@@ -64,19 +64,41 @@ export function LinkPopup(props: LinkPopupProps) {
...
-              })
-            }
-          />
+          <>
+            <Text
+              sx={{
+                mb: 1,
+                ml: 1,
+                fontSize: "body",
+                color: "accent"
```
thecodrr, line 73: I don't think we should be using the `accent` color here.

thecodrr (changes requested):

packages/editor/src/toolbar/popups/link-popup.tsx
luis-411: What color would you suggest?

packages/editor/src/toolbar/popups/link-popup.tsx
```diff
@@ -64,19 +64,41 @@ export function LinkPopup(props: LinkPopupProps) {
...
-              })
-            }
-          />
+          <>
+            <Text
+              sx={{
+                mb: 1,
+                ml: 1,
+                fontSize: "body",
+                color: "secondary"
```
thecodrr, line 73:
```suggestion
                color: "paragraph"
```

packages/editor/src/toolbar/popups/link-popup.tsx
```diff
@@ -64,19 +64,41 @@ export function LinkPopup(props: LinkPopupProps) {
...
+              }
+            />
+          </>
         )}
+        <Text
+          sx={{
+            mb: 1,
+            ml: 1,
+            fontSize: "body",
+            color: "secondary"
```
thecodrr, line 97:
```suggestion
            color: "paragraph"
```

thecodrr:
@luis-411 can you resolve the conflicts so I can merge this?

luis-411:
@thecodrr it should be fine now, right?

PR #7109, opened 2024-12-17
web: add go to next/previous tab shortcut

01zulfi:

### DIFF
--- a/apps/web/src/components/editor/action-bar.tsx
+++ b/apps/web/src/components/editor/action-bar.tsx
@@ -18,7 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { Button, Flex, Text } from "@theme-ui/components";
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import {
   ArrowLeft,
   Cross,
@@ -450,10 +450,25 @@ function Tab(props: TabProps) {
     : Note;
   const { attributes, listeners, setNodeRef, transform, transition, active } =
     useSortable({ id });
+  const activeTabRef = useRef<HTMLElement | null>(null);
+
+  useEffect(() => {
+    if (activeTabRef.current && isActive) {
+      const tab = activeTabRef.current;
+      tab.scrollIntoView({
+        behavior: "smooth",
+        block: "nearest",
+        inline: "nearest"
+      });
+    }
+  }, [isActive]);
 
   return (
     <Flex
-      ref={setNodeRef}
+      ref={(el) => {
+        setNodeRef(el);
+        activeTabRef.current = el;
+      }}
       className="tab"
       data-test-id={`tab-${id}`}
       sx={{
--- a/apps/web/src/components/editor/index.tsx
+++ b/apps/web/src/components/editor/index.tsx
@@ -124,6 +124,31 @@ export default function TabsView() {
   const isTOCVisible = useEditorStore((store) => store.isTOCVisible);
   const [dropRef, overlayRef] = useDragOverlay();
 
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (
+        (event.ctrlKey || event.metaKey) &&
+        event.altKey &&
+        event.key === "ArrowRight"
+      ) {
+        event.preventDefault();
+        useEditorStore.getState().openNextSession();
+      }
+      if (
+        (event.ctrlKey || event.metaKey) &&
+        event.altKey &&
+        event.key === "ArrowLeft"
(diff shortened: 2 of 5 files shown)


apps/web/src/stores/editor-store.ts
```diff
@@ -710,6 +710,28 @@ class EditorStore extends BaseStore<EditorStore> {
...
+          return this.openSession(sessions[0].id);
+        }
+        return this.openSession(sessions[index + 1].id);
+      case "left":
+        if (index === 0) {
+          return this.openSession(sessions[sessions.length - 1].id);
+        }
+        return this.openSession(sessions[index - 1].id);
+    }
+  };
```
thecodrr, line 733: Create 2 methods instead: `openPreviousSession` and `openNextSession`

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -165,7 +165,15 @@ function TipTap(props: TipTapProps) {
...
-        handleKeyDown(view, event) {
+        handleKeyDown(_, event) {
+          if (event.ctrlKey && event.key === "ArrowRight") {
+            event.preventDefault();
+            useEditorStore.getState().openNextSession("right");
+          }
+          if (event.ctrlKey && event.key === "ArrowLeft") {
+            event.preventDefault();
+            useEditorStore.getState().openNextSession("left");
+          }
```
thecodrr, line 176: Not sure this is the right place to handle this. Also Ctrl + Arrow Keys are used for caret navigation so we can't use those.

apps/web/src/components/editor/action-bar.tsx
```diff
@@ -439,10 +439,27 @@ function Tab(props: TabProps) {
...
+  }, [isActive]);
 
   return (
     <Flex
-      ref={setNodeRef}
+      ref={(el) => {
+        setNodeRef(el);
+        if (isActive) {
+          activeTabRef.current = el;
+        }
```
thecodrr, line 461: No need to wrap this in a condition.

thecodrr (changes requested):

apps/web/src/components/editor/tiptap.tsx
01zulfi: Nice catch on the Ctrl + Arrow shortcut. I've moved this shortcut to Ctrl + Alt + Arrow, let me know if this is okayy

apps/web/src/components/editor/action-bar.tsx
01zulfi: Without the condition, the tab strip would scroll to the last tab regardless of the active tab position.

apps/web/src/components/editor/index.tsx
```diff
@@ -121,6 +121,23 @@ export default function TabsView() {
...
+      if (event.ctrlKey && event.altKey && event.key === "ArrowLeft") {
+        event.preventDefault();
+        useEditorStore.getState().openPreviousSession();
+      }
+    };
+    document.body.addEventListener("keydown", onKeyDown);
+    return () => {
+      document.body.removeEventListener("keydown", onKeyDown);
+    };
+  }, []);
```
01zulfi, line 139: I've moved the `onKeyDown` handler for this keyboard shortcut here as I noticed that the `readonly` notes weren't handled.

apps/web/src/components/editor/action-bar.tsx
thecodrr: You are putting the condition in the wrong place then. It should be in the `useEffect`.

apps/web/src/components/editor/index.tsx
thecodrr: Makes sense.

apps/web/src/components/editor/index.tsx
```diff
@@ -121,6 +121,23 @@ export default function TabsView() {
   const isTOCVisible = useEditorStore((store) => store.isTOCVisible);
   const [dropRef, overlayRef] = useDragOverlay();
 
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      if (event.ctrlKey && event.altKey && event.key === "ArrowRight") {
+        event.preventDefault();
+        useEditorStore.getState().openNextSession();
+      }
+      if (event.ctrlKey && event.altKey && event.key === "ArrowLeft") {
```
thecodrr, line 130: We should do `(event.ctrlKey || event.metaKey)` to support `Cmd` on macOS.

thecodrr (changes requested):

thecodrr:
Good work!

PR #7117, opened 2024-12-19
editor: fix task list state & readonly bugs

### LINKED ISSUE #6637: Readonly of "Task List" can be bypassed
Create a Task list with some check boxes and make it Readonly, then:

Expected: There is NO way  to change the TaskList's content, the checkboxes can only be checked/unchecked and moved around.

Actual: Content can be pasted inside the Task List. Items can also be selected with a double mouse click and then content can be written/pasted in the selection's place.

**Device information:**
App version: 3.0.17-0ddbedf-desktop
OS: Linux 64-bit
Browser: Electron 29.4.6
Pro: false

### LINKED ISSUE #7116: Task title not completed in Windows and web app
When you complete all the tasks in a list it marks the Task header title complete automatically. In the Windows app (if you close and reopen) and the web app (if you close the browser tab and reopen) the task title is marked not complete and 0/0 tasks are complete, even though all tasks are complete. It appears correctly in the Android app.

**Device information:**
App version: 3.0.22-7f2c8ca-web
OS: Windows 10 64-bit
Browser: Firefox 133.0
Pro: true

01zulfi:
* fix task list complete state not showing on first load
* fix task list allowing paste when readonly

Closes #7116 
Closes #6637

### DIFF
--- a/packages/editor/src/extensions/task-list/task-list.ts
+++ b/packages/editor/src/extensions/task-list/task-list.ts
@@ -32,7 +32,11 @@ import {
   findParentNodeClosestToPos,
   getExactChangedNodes
 } from "../../utils/prosemirror.js";
-import { countCheckedItems, findRootTaskList, toggleChildren } from "./utils.js";
+import {
+  countCheckedItems,
+  findRootTaskList,
+  toggleChildren
+} from "./utils.js";
 import { Node as ProsemirrorNode } from "@tiptap/pm/model";
 import { TaskItemNode } from "../task-item/index.js";
 
@@ -49,7 +53,15 @@ export const TaskListNode = TaskList.extend({
     return {
       stats: {
         default: { checked: 0, total: 0 },
-        rendered: false
+        rendered: false,
+        parseHTML: (element) => {
+          if (!element.dataset.title) return { checked: 0, total: 0 };
+          const total = element.querySelectorAll("li.checklist--item").length;
+          const checked = element.querySelectorAll(
+            "li.checklist--item.checked"
+          ).length;
+          return { checked, total };
+        }
       },
       title: {
         default: null,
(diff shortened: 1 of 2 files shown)


packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -49,7 +53,15 @@ export const TaskListNode = TaskList.extend({
...
-        rendered: false
+        rendered: false,
+        parseHTML: (element) => {
+          if (!element.dataset.title) return { checked: 0, total: 0 };
+          const items = element.querySelectorAll("li.checklist--item");
+          const checked = Array.from(items).filter((item) =>
+            item.classList.contains("checked")
+          ).length;
+          return { checked, total: items.length };
+        }
```
thecodrr, line 64: I like this. We can remove the logic in https://github.com/streetwriters/notesnook/blob/0921fef267a19a971c4e6a989500d6d0bd4db3c4/packages/editor/src/extensions/task-list/task-list.ts#L225-L244 which is supposed to do the same thing.

thecodrr (changes requested):

packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -49,7 +53,15 @@ export const TaskListNode = TaskList.extend({
...
       stats: {
         default: { checked: 0, total: 0 },
-        rendered: false
+        rendered: false,
+        parseHTML: (element) => {
+          if (!element.dataset.title) return { checked: 0, total: 0 };
+          const items = element.querySelectorAll("li.checklist--item");
+          const checked = Array.from(items).filter((item) =>
+            item.classList.contains("checked")
+          ).length;
```
thecodrr, line 62: This is...unnecessary. Why not just do `querySelectorAll("li.checklist--item.checked").length`?

packages/editor/src/extensions/task-list/task-list.ts
```diff
@@ -49,7 +53,15 @@ export const TaskListNode = TaskList.extend({
     return {
       stats: {
         default: { checked: 0, total: 0 },
-        rendered: false
+        rendered: false,
+        parseHTML: (element) => {
+          if (!element.dataset.title) return { checked: 0, total: 0 };
```
thecodrr, line 58: What's this for?

01zulfi: For nested task lists, the snapshots have `checked` and `total` counts as 0 regardless if they have task items. 

Without this guard clause, `checked` and `total` for nested tasks lists were updating

thecodrr: That guard is incorrect in that case because even the top-most task list doesn't always have `data-title`. Let's instead do: `!!element.closest("ul")`. It's a bit inefficient but it'll be accurate.

thecodrr, line 58:
```suggestion
          if (!!element.closest("ul")) return { checked: 0, total: 0 };
```

thecodrr (approved):

PR #7143, opened 2024-12-25
web: fix status bar size changing

### LINKED ISSUE #7055: Status bar size changes between notes
Switching from a note with a scrollbar to one without or vice versa causes the status bar size to change.

**Device information:**
App version: 3.0.22-7f2c8ca-desktop
OS: Windows 10 64-bit
Browser: Electron 31.7.4
Pro: true

01zulfi:

Closes #7055

### DIFF
--- a/apps/web/src/components/status-bar/index.tsx
+++ b/apps/web/src/components/status-bar/index.tsx
@@ -61,6 +61,7 @@ function StatusBar() {
         borderTopColor: "separator",
         justifyContent: "space-between",
         display: ["none", "none", "flex"],
+        flexShrink: 0,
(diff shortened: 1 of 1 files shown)


thecodrr:
@01zulfi were you able to reproduce this on any browser? I tried but the status bar stayed the same. I also see that the status bar has a fixed height set in its styles so not sure if this is really happening or not.

01zulfi:
Yep, I was able to reproduce on Chrome and Edge. On Edge it was pretty noticeable, and very slightly noticeable on Chrome. Haven't tried Firefox

PR #7212, opened 2024-12-28
editor: add support for pasting unformatted links

### LINKED ISSUE #6663: I love your product
Hi, 

I switched to you guys a few months ago and love your product. 

One feature which annoys me to no end however and hampers my productivity is that I am often copy and pasting emails or weblinks to products. And that it hyperlinks into the text makes my life super difficult. If you can please make it an option to turn that off it would be much appreciated. 

**Device information:**
App version: 3.0.18-5f0d14f-web
OS: Windows 10 64-bit
Browser: Chrome 129.0.0.0
Pro: true

01zulfi:

Closes #6663 
Related to #6652 #5712

### DIFF
--- a/packages/editor/src/extensions/link/helpers/pasteHandler.ts
+++ b/packages/editor/src/extensions/link/helpers/pasteHandler.ts
@@ -21,30 +21,77 @@ import { Editor } from "@tiptap/core";
 import { MarkType } from "@tiptap/pm/model";
 import { Plugin, PluginKey } from "@tiptap/pm/state";
 import { find } from "linkifyjs";
+import { linkRegex } from "../link";
+
+function linkifyHtml(text: string): string {
+  const links = find(text).filter((i) => i.isLink);
+  let out = "";
+  let lastIndex = 0;
+  links.forEach((link) => {
+    out += text.slice(lastIndex, link.start);
+    out += `<a href="${link.href}" target="_blank" rel="noopener noreferrer nofollow">${link.value}</a>`;
+    lastIndex = link.end;
+  });
+  out += text.slice(lastIndex);
+  return out;
+}
 
 type PasteHandlerOptions = {
   editor: Editor;
   type: MarkType;
+  linkOnPaste: boolean;
 };
 
 export function pasteHandler(options: PasteHandlerOptions): Plugin {
+  let shiftKey = false;
   return new Plugin({
     key: new PluginKey("handlePasteLink"),
     props: {
+      handleKeyDown(_, event) {
+        shiftKey = event.shiftKey;
+        return false;
+      },
       handlePaste: (view, event, slice) => {
+        if (!options.linkOnPaste) {
+          return false;
+        }
+
+        const clipboardHtmlData = event.clipboardData?.getData("text/html");
+        if (clipboardHtmlData) {
+          return false;
+        }
+
         const { state } = view;
         const { selection } = state;
         const { empty } = selection;
-
-        if (empty) {
-          return false;
-        }
 
         let textContent = "";
 
         slice.content.forEach((node) => {
           textContent += node.textContent;
         });
+
+        // don't deal with markdown links in this handler
+        if (linkRegex.test(textContent)) {
+          // reset the regex since we don't want the above test method to affect other places where the regex is used
+          linkRegex.lastIndex = 0;
+          return false;
+        }
+
+        if (shiftKey) {
+          shiftKey = false;
+          return false;
+        }
+
+        if (empty) {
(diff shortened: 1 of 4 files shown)


thecodrr (commented):
This is really good work. Will take a deeper look soon. Looks okay though.

packages/editor/src/extensions/link/helpers/pasteHandler.ts
```diff
@@ -20,32 +20,65 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+        // don't deal with markdown links in this handler
+        if (linkRegex.test(textContent)) {
+          // reset the regex since we don't want the above test method to affect other places where the regex is used
+          linkRegex.lastIndex = 0;
+          return false;
+        }
+
+        if (shiftKey) {
+          return false;
+        }
```
thecodrr, line 67: Reset the `shiftKey` flag here otherwise links will be ignored every time.

packages/editor/src/extensions/link/helpers/pasteHandler.ts
```diff
@@ -20,32 +20,65 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            const html = linkifyHtml(textContent);
+            options.editor.commands.insertContent(html);
+            return true;
+          }
+          return false;
+        }
+
+        if (!options.linkOnPaste) {
+          return false;
+        }
```
thecodrr, line 80: This should be the first condition.

packages/editor/src/extensions/link/helpers/pasteHandler.ts
```diff
@@ -20,32 +20,65 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import { Editor } from "@tiptap/core";
 import { MarkType } from "@tiptap/pm/model";
 import { Plugin, PluginKey } from "@tiptap/pm/state";
+import linkifyHtml from "linkify-html";
```
thecodrr, line 23: Why aren't we using `linkifyjs`?

thecodrr (changes requested):

packages/editor/src/extensions/link/helpers/pasteHandler.ts
01zulfi: https://github.com/streetwriters/notesnook/blob/2878ef32b422ead2235340ff137d0cbc013e86e4/packages/editor/src/extensions/link/helpers/pasteHandler.ts#L39

`event.shiftKey` is either `true` or `false`, so we don't have to reset manually

packages/editor/src/extensions/link/helpers/pasteHandler.ts
01zulfi: That's is what I initially did and since `linkOnPaste` is always `true`, so the check would always run the handler. 

The reason I opted to add this check here is that if we ever decide to disable this option or make it user configurable, it wouldn't affect the earlier logic in the handler of pasting links.

Please let me know if we should move the condition to the top regardless

packages/editor/src/extensions/link/helpers/pasteHandler.ts
01zulfi: I found `linkifyHtml` to be much simpler for this usecase

packages/editor/src/extensions/link/helpers/pasteHandler.ts
thecodrr: What if someone pastes using `Ctrl + Shift + V` and then pastes again using the context menu? In that case the `shiftKey` will still be true.

packages/editor/src/extensions/link/helpers/pasteHandler.ts
thecodrr: I think moving it to top makes the most sense even if we don't use it.

packages/editor/src/extensions/link/helpers/pasteHandler.ts
thecodrr: Let's avoid adding a new dep. Either replace all usages of `linkify-js` with `linkify-html` or just use `linkify-js` everywhere.

packages/editor/src/extensions/link/helpers/pasteHandler.ts
01zulfi: Nice observation, thanks

packages/editor/src/extensions/link/helpers/pasteHandler.ts
01zulfi: Makes sense

PR #7272, opened 2025-01-04
editor: add clear current node keyboard shortcut

01zulfi:

### DIFF
--- a/packages/editor/src/extensions/key-map/key-map.ts
+++ b/packages/editor/src/extensions/key-map/key-map.ts
@@ -22,6 +22,7 @@ import { isInTable } from "@tiptap/pm/tables";
 import { CodeBlock } from "../code-block/index.js";
 import { showLinkPopup } from "../../toolbar/popups/link-popup.js";
 import { isListActive } from "../../utils/list.js";
+import { Node } from "@tiptap/pm/model";
 
 export const KeyMap = Extension.create({
   name: "key-map",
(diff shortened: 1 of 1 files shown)


packages/editor/src/extensions/key-map/key-map.ts
```diff
@@ -116,6 +117,15 @@ export const KeyMap = Extension.create({
...
           return false;
         }
+      },
+      [tiptapKeys.clearCurrentLine.keys]: ({ editor }) => {
+        let currentNode: Node;
+        if (isListActive(editor)) {
+          currentNode = editor.state.selection.$from.node(-1);
+        } else {
+          currentNode = editor.state.selection.$from.node();
+        }
```
thecodrr: What's this?

01zulfi: It deletes the entire list item, or should it just clear the list item content?

01zulfi: Removed this special handling for lists

Plus clearing a collapsed heading would keep the hidden content hidden. Now if a collapsed heading is cleared, it will un-collapse it first

thecodrr: > Now if a collapsed heading is cleared, it will un-collapse it first

Question: should we delete the collapsed content instead?

01zulfi: I think we shouldn't but I don't have a strong opinion on this

thecodrr: Let's ask the users.

PR #7290, opened 2025-01-08
web: wrap & remove auto-resize of note title

01zulfi:
* wrap note title if it overflows
* remove auto-resizing of note title based on its length

### DIFF
--- a/apps/web/src/components/editor/title-box.tsx
+++ b/apps/web/src/components/editor/title-box.tsx
@@ -27,6 +27,7 @@ import { replaceDateTime } from "@notesnook/editor";
 import { useStore as useSettingsStore } from "../../stores/setting-store";
 import { AppEventManager, AppEvents } from "../../common/app-events";
 import { strings } from "@notesnook/intl";
+import { NEWLINE_STRIP_REGEX } from "@notesnook/core";
 
 type TitleBoxProps = {
   id: string;
@@ -137,7 +138,7 @@ function TitleBox(props: TitleBoxProps) {
           e.target.value,
           dateFormat,
           timeFormat
-        );
+        ).replace(NEWLINE_STRIP_REGEX, " ");
         debouncedOnTitleChange(id, id, e.target.value, pendingChanges);
       }}
     />
--- a/apps/web/src/components/editor/title-box.tsx
+++ b/apps/web/src/components/editor/title-box.tsx
@@ -17,19 +17,10 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import React, {
-  useCallback,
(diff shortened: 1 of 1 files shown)


packages/core/__tests__/notes.test.ts
```diff
@@ -166,11 +166,11 @@ test("note title should allow trailing space", () =>
     }
   ));
 
-test("note title should not allow newlines", () =>
+test("note title should allow newlines", () =>
   noteTest({ title: "Hello\nhello", content: TEST_NOTE.content }).then(
     async ({ db, id }) => {
       const note = await db.notes.note(id);
-      expect(note?.title).toBe("Hello hello");
+      expect(note?.title).toBe("Hello\nhello");
```
thecodrr, line 173: We don't allow newlines in titles. Why is the test being changed here?

thecodrr (changes requested):

thecodrr:
This PR also isn't resizing the title input when the title exceeds screen width and wraps.

packages/core/__tests__/notes.test.ts
01zulfi: This PR adds support for multilines which includes newlines as well

01zulfi:
> This PR also isn't resizing the title input when the title exceeds screen width and wraps.
There's a limit on how small the title size can be. I'm not sure what the expected behavior is, do you have a suggestion?

thecodrr:
> There's a limit on how small the title size can be. I'm not sure what the expected behavior is, do you have a suggestion?

What do you mean?

01zulfi:
> What do you mean?

The title size has a minimum font size of 1.2em. The title font size does get reduced based on text length regardless of the wrapping.

thecodrr:
@01zulfi we can remove automatic title font resizing.

01zulfi:
> we can remove automatic title font resizing.

Done

packages/core/__tests__/notes.test.ts
thecodrr: We have to strip newlines. Multilines basically means wrapping not actual multiple lines.

packages/core/src/collections/notes.ts
```diff
@@ -110,7 +110,6 @@ export class Notes implements ICollection {
       }
 
       if (typeof item.title !== "undefined") {
-        item.title = item.title.replace(NEWLINE_STRIP_REGEX, " ");
```
thecodrr, line 113: This must be brought back.

packages/core/src/utils/title-format.ts
thecodrr, line 1: Undo these changes.

thecodrr (changes requested):
We need to remove newlines in `onChange`.

PR #7314, opened 2025-01-13
web: add a simple command palette

01zulfi:


apps/web/src/dialogs/command-palette-dialog.tsx
```diff
@@ -0,0 +1,223 @@
...
+                setSelected(0);
+              }
+            }
+            if (e.key === "ArrowDown") {
+              e.preventDefault();
+              setSelected(
+                (selected) => (selected + 1) % filteredCommands.length
+              );
+            }
+            if (e.key === "ArrowUp") {
```
ojoanalogo, line 153: I suggest you keep these keyboard keys as an enum in a separate file to avoid using "[magic strings](https://stoic.software/articles/magic-strings-the-anti-pattern/)".

01zulfi: Sounds good, maybe we can take this up in a separate PR

thecodrr (changes requested):

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+    : query.trim();
+  const result = queryClean.length > 0 ? match(queryClean, text) : false;
+
+  return (
+    <span
+      dangerouslySetInnerHTML={{
+        __html:
+          result && result.match
+            ? surround(text, {
+                result: result,
```
01zulfi, line 269: I have to use fuzzyjs's `match` again here to be able to use `surround`. Is this fine, or do you think we should refactor so that the `lookup` in `core` returns the object that `surround` expects?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+          return;
+        } else {
+          setCommands(res ?? []);
+        }
+      }, 500);
+
+      return () => {
+        clearTimeout(timeoutId);
+      };
+    }, [query]);
```
thecodrr, line 86: This doesn't need to be in a `useEffect`. Just run this logic in `onChange`.

apps/web/src/stores/command-palette-store.ts
```diff
@@ -0,0 +1,207 @@
...
+    this.set((state) => {
+      state.selected = selected;
+    });
+  };
+
+  setQuery = (query: string) => {
+    this.set((state) => {
+      state.query = query;
+    });
+  };
```
thecodrr, line 79: Why are we storing the query in the store? Is it actually needed anywhere else? Can this be in local state instead of in a zustand store? Same for other properties in this store.

01zulfi: I first made the store to move all logic outside of the component. After these review comments, I've realized we can make do with local state and put all other functions in a static util class

apps/web/src/common/key-map.ts
```diff
@@ -141,6 +142,15 @@ const KEYMAP = [
   //     themestore.get().toggleNightMode();
   //   },
   // },
+  {
+    keys: ["ctrl+k"],
+    description: "Open command palette",
+    global: false,
```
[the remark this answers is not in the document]
01zulfi: The shortcut wasn't working with `global: true`

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: I've added a `db.lookup.fuzzy` method. Not sure if that was exactly what you meant, do take a look please

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,317 @@
...
+  }
+);
+
+function Highlighter({ text, query }: { text: string; query: string }) {
+  const queryClean = query.startsWith(">")
+    ? query.slice(1).trim()
+    : query.trim();
+  const result =
+    queryClean.length > 0
+      ? db.lookup.fuzzy(queryClean, text, {
```
thecodrr, line 304: This is redundant work. Ideally, there should be a function in `lookup` that takes a list of objects (following whatever interface), a query, and suffix/prefix and returns a list of results that are already wrapped in a suffix/prefix so we can skip all this extra work.

thecodrr (changes requested):

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: I'm unsure if the changes are still in line with what you have in mind, do take a look please

thecodrr: This will do for now, I think.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,472 @@
...
+                  ? (e) => {
+                      setSelected(0);
+                      const query = e.target.value;
+                      setQuery(query);
+
+                      const res = search(query);
+                      if (res instanceof Promise) {
+                      } else {
+                        setCommands(res ?? []);
+                      }
```
thecodrr, line 153: Just invert the `if..else`.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,472 @@
...
+                  : debounce(async (e) => {
+                      setSelected(0);
+                      const query = e.target.value;
+                      setQuery(query);
+
+                      const res = search(query);
+                      if (res instanceof Promise) {
+                        const commands = await res;
+                        setCommands(commands ?? []);
+                        return;
```
thecodrr, line 164: What's this doing?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,472 @@
...
+                        setCommands(res ?? []);
+                      }
+                    }
+                  : debounce(async (e) => {
+                      setSelected(0);
+                      const query = e.target.value;
+                      setQuery(query);
+
+                      const res = search(query);
+                      if (res instanceof Promise) {
```
thecodrr, line 161: I don't think this is a good idea. Make it all `async` and get rid of this `if..else` conditions.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,472 @@
...
+      group: item.type,
+      type: item.type
+    };
+  });
+  return commands;
+}
+
+function getRecentCommands() {
+  return Config.get<Command[]>("commandPalette:recent", []);
+}
```
thecodrr, line 464: Can you move this to where all the other `recentCommands` functions are?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,472 @@
...
+      if (!acc[command.group]) {
+        acc[command.group] = [];
+      }
+      acc[command.group].push({
+        ...command,
+        icon: getCommandIcon(command),
+        index
+      });
+      return acc;
+    }, {} as GroupedCommands);
```
thecodrr, line 91: Why is it done here instead of in the `onChange` function?

thecodrr (commented):
This needs to be stress tested with a lot of notes, notebooks, tags etc. Can you generate 100,000 random notes, notebooks & tags and try using the command palette with that?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: I think groupifying the commands is only needed during render, that's why it is here. Do you mean we should have `grouped` be managed in state, or do you have something else in mind?

01zulfi:
> This needs to be stress tested with a lot of notes, notebooks, tags etc. Can you generate 100,000 random notes, notebooks & tags and try using the command palette with that?

Sure, I'll do that

01zulfi:
On my machine, the command palette starting lagging when navigating through hundreds of search results. I will look into this, I think we'll need to add virtualization for the search mode.

thecodrr:
@01zulfi yes, virtualization is a must.

apps/web/src/common/key-map.ts
```diff
@@ -165,6 +166,15 @@ const KEYMAP = [
...
```
thecodrr, line 172: You can remove this. It's no longer used.

apps/web/src/common/key-map.ts
```diff
@@ -165,6 +166,15 @@ const KEYMAP = [
   //     themestore.get().toggleNightMode();
   //   },
   // },
+  {
+    keys: ["ctrl+k"],
```
thecodrr, line 170: Add `cmd+k` as well. Also `ctrl+shift+k` should open the command palette in search mode.

packages/core/src/api/lookup.ts
```diff
@@ -163,6 +189,46 @@ export default class Lookup {
...
+      suffix?: string;
+      /**
+       * If true, only items that match the query will be returned
+       */
+      matchOnly?: boolean;
+    }
+  ): T[] {
+    if (query === "") return items;
+
+    const fuzzied: (T | undefined)[] = [];
```
thecodrr, line 207: It should only be `T[]`. You can filter out `undefined` items in the loop below.

packages/core/src/api/lookup.ts
```diff
@@ -163,6 +189,46 @@ export default class Lookup {
...
+            prefix: opts?.prefix,
+            suffix: opts?.suffix
+          })
+        });
+        continue;
+      }
+      fuzzied.push(item);
+    }
+
+    return fuzzied.filter((f) => f !== undefined);
```
thecodrr, line 229: No need for the `.filter` here. Do everything in the loop above.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,461 @@
...
+
+    useEffect(() => {
+      virtuosoRef.current?.scrollToIndex({
+        index: selected,
+        align: "end",
+        behavior: "auto"
+      });
+    }, [selected]);
+
+    async function onChange(e: React.ChangeEvent<HTMLInputElement>) {
```
thecodrr, line 69: This should be wrapped in a `useCallback`.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,461 @@
...
+      setQuery(query);
+      !isCommandMode(query) && setLoading(true);
+      const res = await search(query);
+      const highlighted = db.lookup.fuzzy(
+        prepareQuery(query),
+        res.map((r) => ({
+          ...r,
+          highlightedTitle: r.title
+        })) ?? [],
+        "highlightedTitle",
```
thecodrr, line 81: Why are we using a separate key for `highlightedTitle` instead of just replacing the `title`?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,461 @@
...
+        align: "end",
+        behavior: "auto"
+      });
+    }, [selected]);
+
+    async function onChange(e: React.ChangeEvent<HTMLInputElement>) {
+      setSelected(0);
+      const query = e.target.value;
+      setQuery(query);
+      !isCommandMode(query) && setLoading(true);
```
thecodrr, line 73: Use the proper construct for this i.e. an if statement.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,461 @@
...
+          highlightedTitle: r.title
+        })) ?? [],
+        "highlightedTitle",
+        {
+          prefix: "<b style='color: var(--accent-foreground)'>",
+          suffix: "</b>"
+        }
+      );
+      setCommands(highlighted ?? []);
+      setLoading(false);
```
thecodrr, line 88: This should be called in the `finally` block of a `try...finally` in case we get an error, we don't want to be stuck forever in loading mode.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,461 @@
...
+
+    const grouped = commands.reduce((acc, command) => {
+      const item = acc.find((c) => c.group === command.group);
+      if (item) {
+        item.count++;
+      } else {
+        acc.push({ group: command.group, count: 1 });
+      }
+      return acc;
+    }, [] as GroupedCommands);
```
thecodrr, line 99: Can this be wrapped in a `useMemo`?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,461 @@
...
+          onKeyDown={(e) => {
+            if (e.key == "Enter") {
+              e.preventDefault();
+              const command = commands[selected];
+              if (!command) return;
+              const action = getCommandAction({
+                id: command.id,
+                type: command.type
+              });
+              if (action) {
```
thecodrr, line 133: Why wouldn't there be an action for a command?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,461 @@
...
+            placeholder={"Search in notes, notebooks, and tags"}
+            sx={{ mx: 0, my: 2 }}
+            defaultValue={query}
+            onChange={isCommandMode(query) ? onChange : debounce(onChange, 500)}
+          />
+          {commands.length === 0 && (
+            <Box>
+              <Text variant="subBody">No results found</Text>
+            </Box>
+          )}
```
thecodrr, line 161: This should only be shown if there's something in the `query`.

thecodrr (changes requested):
This is starting to look really good.

01zulfi:
My current counts

| entity | count |
|--------|--------|
| notes | 110486 |
| notebooks | 101581 |
| tags | 100559 |

The search is now taking a few minutes to complete. Searching took few seconds when I had around 30k-70k items each. Some queries returned ~60k results and navigation worked flawlessly with no lag.

apps/web/src/common/key-map.ts
01zulfi: this PR has `ctrl+shift+k` for inserting links, do you have a suggestion for that?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: Highlighting the title key would also add the highlighted title to recent commands in local storage

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: Because... TypeScript 😆 

`getCommandAction` has a possibility of returning `undefined` because of `CommandActionMap.get`.  Do you think we should use a non-null asssertion operator (`!`) when we `get` from the map? For now, I replaced `if (action)` with `action?.`

thecodrr:
> The search is now taking a few minutes to complete.

That's too long. Can you run a basic profile to see what's taking the most time?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
thecodrr: Would be a good idea to add a comment for that.

apps/web/src/common/key-map.ts
thecodrr: Maybe we can do `ctrl+p` for search mode but that won't work on browsers, I think?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,472 @@
...
+    titleOnly: true
+  });
+  const list = (
+    await Promise.all([
+      notes.items(),
+      notebooks.items(),
+      tags.items(),
+      reminders.items()
+    ])
+  ).flat();
```
01zulfi, line 454: This takes up 99% of the time. I was able to verify with `console.time`

apps/web/src/common/key-map.ts
01zulfi: > won't work on browsers,

That's what I thought, but it works for some reason 😆 

Command palette trigger shortcuts now:
`ctrl+k` - open in command mode
`ctrl+p` - open in search mode

I think these are good for now, though I'll keep looking for better. If you have any suggestions, do let me know

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,595 @@
...
+    const nextIndex = (current.x + 1) % 2;
+    return { x: nextIndex, y: current.y };
+  },
+  left: (current, commands) => {
+    const currentCommand = commands[current.y];
+    if (currentCommand.group !== "recent") return current;
+    const nextIndex = (current.x - 1 + 2) % 2;
+    return { x: nextIndex, y: current.y };
+  }
+};
```
thecodrr, line 408: This is super confusing. Let's split these into their own functions that clearly describe what exactly they do instead of something like `arrowUp`, `arrowDown` etc. Function names should clearly explain what the function does instead of describing where its called or how its invoked.

thecodrr (changes requested):

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: I've renamed them to `moveSelectionRight` etc

Let me know if that works

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -608,7 +608,24 @@ function toIEditor(editor: Editor): IEditor {
...
+    toggleBulletList: () => editor.commands.toggleBulletList(),
+    toggleOrderedList: () => editor.commands.toggleOrderedList(),
+    toggleTaskList: () => editor.commands.toggleTaskList(),
+    toggleOutlineList: () => editor.commands.toggleOutlineList(),
+    insertHorizontalRule: () => editor.commands.setHorizontalRule(),
+    toggleCodeBlock: () => editor.commands.toggleCodeBlock(),
+    // todo: figure out following typescript error
+    insertMathBlock: () => editor.chain().focus().insertMathBlock().run(),
+    toggleQuoteBlock: () => editor.commands.toggleBlockquote(),
+    addAttachment: () => editor.storage.openAttachmentPicker?.("file")
```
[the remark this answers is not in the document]
01zulfi: I think these are still worth keeping till slash commands go live. Though if we still think they should be removed, just let me know, I'll remove them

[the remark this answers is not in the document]
thecodrr: Yes, I think they should be removed. Mostly because the `IEditor` interface represents a very general abstraction over the editor to allow us to support multiple editors in the future. Exposing all these specifics negates the whole point of having such an abstraction. I think Slash Commands are a better place for these, anyway.

thecodrr (approved):

thecodrr (changes requested):

packages/core/src/api/lookup.ts
```diff
@@ -163,6 +188,47 @@ export default class Lookup {
...
+            suffix: opts?.suffix
+          })
+        });
+        continue;
+      }
+      fuzzied.push(item);
+    }
+
+    return fuzzied;
+  }
```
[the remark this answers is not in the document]
01zulfi: I've added a sort option, let me know if that works

packages/core/__tests__/lookup.test.js
```diff
@@ -81,13 +81,57 @@ test("search notes with an empty note", () =>
...
+  }).then(async ({ db }) => {
+    await db.notes.add({
+      title: "hello there",
+      content: { type: "tiptap", data: "<p>note of the world<br></p>" }
+    });
+    let filtered = await db.lookup
+      .notes("hello there", undefined, { titleOnly: true })
+      .ids();
+    expect(filtered).toHaveLength(1);
+  }));
```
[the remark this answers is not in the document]
thecodrr: That'd be better. We can use `prefix` and `suffix` and make sure they are applied only in title even if the description has that word.

thecodrr (changes requested):

thecodrr (approved):

thecodrr:
@01zulfi this is done. Can we do the localization changes and merge it?

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,625 @@
...
+  type
+}: {
+  id: Command["id"];
+  type: Command["type"];
+}) {
+  switch (type) {
+    case "command":
+      return CommandActionMap.get(id);
+    case "command-dynamic":
+      return CommandActionMap.get(id);
```
thecodrr, line 494:
```suggestion
    case "command":
    case "command-dynamic":
      return CommandActionMap.get(id);
```

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -0,0 +1,625 @@
...
+  type
+}: {
+  id: Command["id"];
+  type: Command["type"];
+}) {
+  switch (type) {
+    case "command":
+      return CommandIconMap.get(id);
+    case "command-dynamic":
+      return CommandIconMap.get(id);
```
thecodrr, line 518:
```suggestion
    case "command":
    case "command-dynamic":
      return CommandIconMap.get(id);
```

Issue #7335, opened 2025-01-15
Navigation menu does not resize after being collapsed


luis-411:
### What happened?

The navigation menu (the pane on the left) collapses to just showing the icons when the width of the screen is below a certain value.
However, it does not go back to the normal size after increasing the screen width again.

### Steps to reproduce the problem

- Shrink the width of your screen until the menu collapses.
- Increase the width of your screen again.
- Notice how the menu is still collapsed.

### Version

3.0.24

### Platform/OS

Chrome/Chromium

### Relevant log output

```shell

```

PR #7377, opened 2025-01-22
web: fix nav menu collapse

### LINKED ISSUE #7393: Subfolder bug
Subfolders no longer work. Impossible to display them

**Device information:**
App version: 3.0.24-21e1e73-web
OS: OS X 10.15.7 64-bit
Browser: Chrome 131.0.0.0
Pro: true

luis-411:
Fixes: #7335

This should also fix #7393 since the same functionality is used when toggeling the subnotebook folder.

### DIFF
--- a/apps/web/src/components/split-pane/index.tsx
+++ b/apps/web/src/components/split-pane/index.tsx
@@ -96,7 +96,7 @@ export const SplitPane = React.forwardRef<
   const wrapSize = useRef(0);
   const childrenLength = childrenToArray(children).length;
   const autoSaveKey = autoSaveId ? `csp:${autoSaveId}` : undefined;
-  const previousSize = useRef(0);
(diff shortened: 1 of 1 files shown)


apps/web/src/components/split-pane/index.tsx
```diff
@@ -96,14 +96,15 @@ export const SplitPane = React.forwardRef<
   const wrapSize = useRef(0);
   const childrenLength = childrenToArray(children).length;
   const autoSaveKey = autoSaveId ? `csp:${autoSaveId}` : undefined;
+  const previousSize = useRef(0);
```
thecodrr, line 99: The naming could be improved.

```suggestion
  const lastCollapsedPaneSize = useRef(0);
```

thecodrr (approved):

Issue #7411, opened 2025-01-27
Search can't find links without http protocol


jermanuts:
### What happened?

If I searched for any link in "Notes", it shows no results even though I have that link in one of my notes.

![Image](https://github.com/user-attachments/assets/61fda0b9-7a1b-4f8f-b9bd-d914f48594cd)

### Steps to reproduce the problem

Search for any link without adding http protocol before it, as shown in image attached.

### Version

3.0.23

### Platform/OS

Windows

### Relevant log output

```shell

```

PR #7418, opened 2025-01-28
web: fix search special characters

luis-411:
Fixes: #7411

Special characters were not escaped properly

### DIFF
--- a/apps/web/src/components/route-container/index.tsx
+++ b/apps/web/src/components/route-container/index.tsx
@@ -95,7 +95,7 @@ function Header(props: RouteContainerProps) {
           onChange={debounce(
             (e) =>
               useSearchStore.setState({
-                query: `"${e.target.value.replace(/"/g, '""')}"`
+                query: e.target.value
               }),
             250
           )}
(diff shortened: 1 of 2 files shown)


apps/web/src/components/route-container/index.tsx
```diff
@@ -94,7 +93,10 @@ function Header(props: RouteContainerProps) {
           defaultValue={query}
           placeholder={strings.typeAKeyword()}
           onChange={debounce(
-            (e) => useSearchStore.setState({ query: e.target.value }),
+            (e) =>
+              useSearchStore.setState({
+                query: `"${e.target.value.replace(/"/g, '""')}"`
+              }),
```
thecodrr, line 99: This is not the place to do this. We have a `query-transformer` in `packages/core` that is responsible for handling such cases. I am not sure why the `github.com` query isn't working since all the necessary escaping is already being done.

thecodrr (changes requested):

apps/web/src/components/route-container/index.tsx
luis-411: It works fine, if I add `str.includes(".")` to `maybeColspec` and this could be done for various other special characters, as they also trigger an error, but I'm not sure if that is the proper way to do it.

thecodrr: I think we should use a regex to exclude all such characters. Quick way to test whether a character is invalid or not is by adding it in the search query and checking the browser console. There will be an SQL error if the query contains an invalid character.

thecodrr (approved):

PR #7436, opened 2025-02-03
web: keep some configs in local storage on logout

### LINKED ISSUE #7414: Don't change instance when logging out

Hello,

Would it be possible to not remove the custom instance when logging out the user ? Now when I logged out from my custom instance it resets to the official one.

Stay with the same instance after loggout.

_No response_

### Additional context

I tested in Linux.AppImage

01zulfi:

Closes #7414

### DIFF
--- a/apps/web/src/utils/config.ts
+++ b/apps/web/src/utils/config.ts
@@ -57,5 +57,30 @@ function has(predicate: (key: string) => boolean) {
   return false;
 }
 
-const Config = { set, get, clear, all, has, remove };
+function logout() {
+  const toKeep = [
+    "editorConfig",
+    "backupStorageLocation",
+    "serverUrls",
+    "corsProxy",
+    "theme:light",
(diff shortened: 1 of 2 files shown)


apps/web/src/utils/config.ts
```diff
@@ -57,5 +57,28 @@ function has(predicate: (key: string) => boolean) {
...
+    const val = get(key);
+    if (val !== undefined) acc[key] = val;
+    return acc;
+  }, {} as Record<string, any>);
+
+  clear();
+
+  Object.entries(toKeep).forEach(([key, val]) => {
+    set(key, val);
+  });
```
thecodrr, line 80: Can we simplify this? I don't think going full on functional here is improving things. Just use simple loops.

thecodrr (changes requested):

thecodrr (approved):

PR #7541, opened 2025-02-10
web: hide undo & redo buttons for readonly notes

01zulfi:

### DIFF
--- a/apps/web/__e2e__/models/editor.model.ts
+++ b/apps/web/__e2e__/models/editor.model.ts
@@ -41,6 +41,8 @@ export class EditorModel {
   private readonly newTabButton: Locator;
   readonly savedIcon: Locator;
   readonly notSavedIcon: Locator;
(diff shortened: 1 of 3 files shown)


apps/web/__e2e__/models/editor.model.ts
```diff
@@ -287,4 +291,12 @@ export class EditorModel {
...
     await this.page.keyboard.up("ControlOrMeta");
   }
+
+  async isUndoButtonVisible() {
+    return await this.undoButton.isVisible();
+  }
+
+  async isRedoButtonVisible() {
+    return await this.redoButton.isVisible();
+  }
```
thecodrr, line 301: No need for this. Just use an `expect` assertion instead. You'll have to expose `undoButton` and `redoButton` which is totally fine.

thecodrr (changes requested):

thecodrr (approved):

PR #7553, opened 2025-02-11
editor: set default table header bg to background-secondary

### LINKED ISSUE #7498: Padding in table heading cells. Please think about it once again 🙏

Before v3 came out around v2.5 the grey padding in table heading cells was removed. Below is a screenshot from Notesnook website:

Image

The grey padding had a special feature. It doesn't change with the change of theme. I have tried many background colors and none of them work like that. It becomes an issue when changing the themes. 

For example, if you make a cell with black background, it is fine with dark theme. However, if you change the theme to light theme, the dark color of the text makes the text illegible. 

This was a feature I used many many times in my workflow while handling tables. I deal with a lot of tables and changing background for heading cells after changing a theme becomes a hassle. If not by default, please at least add an option to do this again. 

Maybe add an option for this in the cell edit options. It's just an extra feature. If someone doesn't like the heading padding, they can just remove the first row in the table. That's it. 

_No response_

### Additional context

_No response_

01zulfi:

Closes #7498

### DIFF
--- a/packages/editor/styles/styles.css
+++ b/packages/editor/styles/styles.css
@@ -250,6 +250,7 @@ img.ProseMirror-separator {
 }
 
 .ProseMirror table th {
+  background-color: var(--background-secondary);
   font-weight: bold;
   text-align: left;
 }


Starship-Headlight:
Thanks a lot for this 😁

Just one small request. Could you please change the background color in dark theme to `#2b2b2b`. In this pull request it's set to `#202020` which is not clearly visible in dark themes. Thanks 😁

01zulfi:
@Starship-Headlight 

Re: your comment here https://github.com/streetwriters/notesnook/issues/7498#issuecomment-2648982891

Unfortunately, the current default theme deson't have a variable that has `2b2b2b` for dark mode and `f7f7f7` for light mode. 

Dark theme here: https://github.com/streetwriters/notesnook-themes/blob/main/themes/default-dark/v1/theme.json
Light theme here: https://github.com/streetwriters/notesnook-themes/blob/main/themes/default-light/v1/theme.json

--- 

Maybe we'll need to add a separate entry in theme for table header background. Do you have a suggestion @thecodrr?

Starship-Headlight:
@01zulfi  Maybe #7756 could help. Adding dynamic colors which we can also use as table heading background. That way problems with `2b2b2b` for dark mode and `f7f7f7` for light mode can also be solved.

PR #7582, opened 2025-02-13
web: move importer into webapp

01zulfi:


apps/web/src/dialogs/settings/importer/utils/bytes.ts
```diff
@@ -0,0 +1,32 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+export function bytesToSize(bytes: number, separator = "", postFix = "") {
```
thecodrr, line 20: You can use the `formatBytes` function from `@notesnook/common`.

apps/web/src/dialogs/settings/components/importer.tsx
thecodrr, line 1: Can we move the `importer` folder into `components`?

apps/web/src/dialogs/settings/importer/components/accordion.tsx
thecodrr, line 1: We already have an accordion component.

apps/web/src/utils/importer.ts
```diff
@@ -17,100 +17,62 @@ You should have received a copy of the GNU General Public License
...
-
-  for (const zip of zipFiles) {
-    let count = 0;
-    let filesRead = 0;
-
-    const attachments: Record<string, any> = {};
-
-    for await (const entry of createUnzipIterator(zip)) {
-      ++filesRead;
+export async function pushImportedNoteIntoDatabase(note: Note) {
```
thecodrr, line 30: Just say `importNote`.

apps/web/src/utils/importer.ts
```diff
@@ -17,100 +17,62 @@ You should have received a copy of the GNU General Public License
...
+  const attachments: Record<string, EncryptedAttachmentFields> = {};
+  await processNoteAttachments(note.attachments, attachments);
```
thecodrr, line 32: Hm. This looks odd. Why are attachments being passed as a parameter instead of getting returned by the `processNoteAttachments` function?

apps/web/src/utils/importer.ts
```diff
@@ -17,100 +17,62 @@ You should have received a copy of the GNU General Public License
...
-        if (isAttachment) {
-          await processAttachment(entry, attachments);
-        } else if (isNote) {
-          await processNote(entry, attachments);
-          ++count;
-        }
-      } catch (e) {
-        if (e instanceof Error) yield { type: "error" as const, error: e };
-      }
+async function processNoteAttachments(
```
thecodrr, line 48:
```suggestion
async function processAttachments(
```

apps/web/src/utils/importer.ts
```diff
@@ -17,100 +17,62 @@ You should have received a copy of the GNU General Public License
...
+  chunkSize: number;
+  iv: string;
+  size: number;
+  salt: string;
+  alg: string;
+  key: {
+    key?: string;
+    salt?: string;
+  };
+};
```
thecodrr, line 46: This should probably be a union of `Cipher` and `key`.

apps/web/src/utils/importer.ts
```diff
@@ -198,10 +161,20 @@ async function processNote(entry: ZipEntry, attachments: Record<string, any>) {
...
+  purple: "#7B1FA2",
+  blue: "#1976D2",
+  cerulean: "#03A9F4",
+  pink: "#C2185B",
+  brown: "#795548",
+  gray: "#9E9E9E",
+  green: "#388E3C",
+  orange: "#FFA000",
+  yellow: "#FFC107"
+};
```
thecodrr, line 177: Let's move this to the top. It's kind of a convention:

1. Types
2. Constants
3. Functions

thecodrr (changes requested):

apps/web/src/utils/importer.ts
01zulfi: I found a `Cipher` interface in the `crypto` package, but the fields didn't overlap fully. Currently I'm using this type definition, let me know if it is okay:

```ts
type EncryptedAttachmentFields = Awaited<
  ReturnType<typeof writeEncryptedFile>
> & {
  key: SerializedKey;
};
```

apps/web/src/utils/importer.ts
01zulfi: `attachments` were being passed in like this presumably because the process attachment and notes function were in a try catch block. I believe we don't need to handle errors here now as the importer package will do that for us, so II've made the change to return instead

apps/web/src/components/importer/components/file-provider-handler.tsx
```diff
@@ -0,0 +1,324 @@
...
+            } into database`
+          );
+
+          processedNotes.push(data);
+          await importNote(data);
+        },
+        iterate: async function* () {
+          for (const note of processedNotes) {
+            yield note;
+          }
```
thecodrr, line 101: This is not necessary since we are importing directly into the database.

apps/web/src/components/importer/components/file-provider-handler.tsx
```diff
@@ -0,0 +1,324 @@
...
+      file: provider?.supportedExtensions?.concat([".zip"])
+    }
+  });
+
+  useEffect(() => {
+    setFiles([]);
+  }, [provider]);
+
+  async function onStartImport() {
+    let processedNotes: Note[] = [];
```
thecodrr, line 77: No need to store the processed notes in memory. That's exactly what we want to avoid so users can import insanely huge amount of notes.

thecodrr (changes requested):

apps/web/src/utils/importer.ts
thecodrr: Yeah that could work too, I guess.

thecodrr (approved):

PR #7608, opened 2025-02-17
editor: fix youtube embeds

thecodrr:
This PR fixes youtube and other embeds by allowing `allow-same-origin` in the iframe sandbox ONLY if the source url is an http or https url.

### DIFF
--- a/packages/editor/src/extensions/embed/component.tsx
+++ b/packages/editor/src/extensions/embed/component.tsx
@@ -115,10 +115,10 @@ export function EmbedComponent(
         </Box>
         <Embed
           ref={embedRef}
-          src={src.startsWith("javascript:") ? "about:blank" : src}
+          src={src}
           width={"100%"}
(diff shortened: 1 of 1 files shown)


Starship-Headlight:
@thecodrr Hello, this update fixed the youtube issue however, before the security update for embeds, I used to use a bookmarking app with a share feature to embed my bookmarks in Notesnook. And the embed worked perfectly wherein I would click on a link in the embed and open the link in the browser. However, now, even after this fix, the links in the bookmark embed are unresponsive. Clicking on a link does nothing. And not just Raindrop. It's true for other websites too. Like if you click on any link in google maps e.g., `Directions` link in google map embed after clicking on a location, nothing happens. In Desktop app it is fine as you can right click on the link and copy it to open (which is also 2 extra steps) but in mobile app the links in embedded pages become completely useless. Could you please take a look at this again. Thank you.

PR #7623, opened 2025-02-19
web: fix tags not adding/removing in note header bar

### LINKED ISSUE #7485: Cannot add/delete Tags anymore (web / macos version)
Hello
Since a moment, I cannot add / delete tags anymore to notes. This only affects the web and macos versions for me, android works fine
- I CAN add and delete tags for themselves, but
- I cannot delete a tag in a note anymore
- I cannot add a tag in a note anymore: I can select the tag, but when I change to another note and come back to the note, the tag is not there anymore. In the tags list, however, the tag shows the right number of notes, but also the note linked is not shown. 

**Device information:**
App version: 3.0.25-6a9ddb7-web
OS: OS X 10.15.7 64-bit
Browser: Chrome 132.0.0.0
Pro: true

### LINKED ISSUE #7621: Desktop: Errors removing/updating tags on notes
Clicking the red x on a tag in a note does not remove the tag from the note on Desktop. Removing the same tag on the same note on mobile works, but has errors when updating on desktop; the note still shows the tag beneath the title, but when you click the tag to show all the tagged notes in the sidebar, the note in question does not appear.

Tagging notes seems to work intermittently on Desktop, too. Tagging a note appears to work (the tag shows on the note), but when you go to that tag in the sidebar, the note does not appear.

**Device information:**
App version: 3.0.26-2ab3218-desktop
OS: Windows 10 64-bit
Browser: Electron 31.7.4
Pro: true

01zulfi:

Closes #7621 
Closes #7485

### DIFF
--- a/apps/web/src/components/editor/header.tsx
+++ b/apps/web/src/components/editor/header.tsx
@@ -31,7 +31,6 @@ import { MenuItem } from "@notesnook/ui";
 import { navigate } from "../../navigation";
 import { Tag } from "@notesnook/core";
 import { strings } from "@notesnook/intl";
-import { showToast } from "../../utils/toast";
 
 type HeaderProps = { readonly: boolean; id: string };
 function Header(props: HeaderProps) {
--- a/apps/web/src/components/editor/header.tsx
+++ b/apps/web/src/components/editor/header.tsx
@@ -31,13 +31,16 @@ import { MenuItem } from "@notesnook/ui";
 import { navigate } from "../../navigation";
 import { Tag } from "@notesnook/core";
 import { strings } from "@notesnook/intl";
+import { showToast } from "../../utils/toast";
 
 type HeaderProps = { readonly: boolean; id: string };
 function Header(props: HeaderProps) {
   const { readonly, id } = props;
-  const tags = useEditorStore(
-    (store) => store.getActiveSession(["default", "readonly"])?.tags
+  const activeSession = useEditorStore((store) =>
+    store.getActiveSession(["default", "readonly"])
   );
+  const tags = activeSession?.tags;
+  const noteId = activeSession?.note.id;
 
   const setTag = useCallback(async function (
     noteId: string,
@@ -59,6 +62,8 @@ function Header(props: HeaderProps) {
   },
   []);
 
+  if (!noteId) return null;
+
   return (
     <>
       <Flex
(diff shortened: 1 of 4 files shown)


thecodrr:
Thanks!

apps/web/src/components/editor/header.tsx
```diff
@@ -31,19 +31,27 @@ import { MenuItem } from "@notesnook/ui";
...
-    (store) => store.getActiveSession(["default", "readonly"])?.tags
+  const activeSession = useEditorStore((store) =>
+    store.getActiveSession(["default", "readonly"])
   );
+  const tags = activeSession?.tags;
+  const noteId = activeSession?.note.id;
 
   const setTag = useCallback(async function (
-    noteId: string,
+    noteId: string | undefined,
```
thecodrr, line 46: No need for this. Just don't render the `Header` if `noteId` is `undefined`.

thecodrr (changes requested):

apps/web/src/components/editor/header.tsx
```diff
@@ -31,13 +31,16 @@ import { MenuItem } from "@notesnook/ui";
 import { navigate } from "../../navigation";
 import { Tag } from "@notesnook/core";
 import { strings } from "@notesnook/intl";
+import { showToast } from "../../utils/toast";
```
thecodrr, line 34:
```suggestion
```

apps/web/src/stores/editor-store.ts
```diff
@@ -467,9 +467,13 @@ class EditorStore extends BaseStore<EditorStore> {
...
-            updateSession(event.item.toId, undefined, {
-              tags: await db.notes.tags(event.item.toId)
-            });
+            for (const session of sessions) {
+              if ("note" in session && session.note.id === event.item.toId) {
+                updateSession(session.id, undefined, {
+                  tags: await db.notes.tags(event.item.toId)
+                });
+              }
+            }
```
thecodrr, line 476: There's no need to loop over the sessions here. We are already looping over them above.
```suggestion
            updateSession(session.id, undefined, {
              tags: await db.notes.tags(event.item.toId)
            });
```

thecodrr:
@01zulfi can you add tests for these? I think it's a basic enough functionality that should be tested.

apps/web/src/stores/editor-store.ts
01zulfi: `session` is undefined here. We need to loop over the sessions

thecodrr: You're right. Can you also change it for colors a little bit above? I think it's also broken.

01zulfi: @thecodrr done

thecodrr (approved):

PR #7655, opened 2025-02-24
global: (wip) move keyboard shortcuts metadata into common package

01zulfi:

### DIFF
--- a/apps/web/src/common/key-map.ts
+++ b/apps/web/src/common/key-map.ts
@@ -22,6 +22,9 @@ import { useEditorStore } from "../stores/editor-store";
 import { useStore as useSearchStore } from "../stores/search-store";
 import { useEditorManager } from "../components/editor/manager";
 import { CommandPaletteDialog } from "../dialogs/command-palette";
+import { keybindings } from "@notesnook/common";
+import { KeyboardShortcutsDialog } from "../dialogs/keyboard-shortcuts-dialog";
+import { KEYS } from "../../../../packages/core/dist/types/database/kv";
 
 function isInEditor(e: KeyboardEvent) {
   return (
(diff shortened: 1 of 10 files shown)


apps/web/src/common/key-map.ts
```diff
@@ -62,43 +64,28 @@ const KEYMAP = [
...
-      "command+option+right",
-      "ctrl+alt+right",
-      "command+option+shift+right",
-      "ctrl+alt+shift+right",
-      "ctrl+tab",
-      "command+tab"
-    ],
-    description: "Go to next tab",
+    keys: keybindings.nextTab.keys(IS_DESKTOP_APP),
+    description: keybindings.nextTab.description,
```
thecodrr: Getting the keys like this doesn't respect the `isMac` config. I think instead of exposing `keybindings` object like this directly, we should have a function for getting a specific key binding using its `id`. That way we can process the keys in a centralized way. Also instead of doing `keybindings.nextTab.description`, we should just `...getKeybinding()`.

apps/web/src/common/key-map.ts
thecodrr, line 1: 1. The `KEYMAP` array should be typed to avoid unintended errors.
2. Instead of creating an `KEYMAP` array by hand, we should create an object of `actions` that correspond to the respective key binding. The `registerKeyMap` function would then directly operate on the `keybindings` and take the `action` from the object we created. The object of `actions` could then also be reused in the command palette. Another idea is to reuse the commands in the command palette as actions since almost all key bindings have (or should have) a corresponding command.

apps/web/src/common/key-map.ts
```diff
@@ -109,13 +96,13 @@ const KEYMAP = [
...
+    keys: keybindings.closeAllTabs.keys(IS_DESKTOP_APP),
+    description: keybindings.closeAllTabs.description,
     action: () => useEditorStore.getState().closeAllTabs()
   },
   {
-    keys: ["command+f", "ctrl+f"],
-    description: "Search all notes",
+    keys: keybindings.searchInNotes.keys(IS_DESKTOP_APP),
+    description: keybindings.searchInNotes.description,
     global: false,
```
thecodrr: This is pointless.

apps/web/src/common/key-map.ts
01zulfi: > The registerKeyMap function would then directly operate on the keybindings

I'm not sure what it meant. I've pushed some changes do take a look

thecodrr (changes requested):
A few things can be improved here:

image

1. The legibility of the key identifiers is not that good.
2. We can use symbols to represent system keys like Command, Option, Shift, Arrow Keys etc. Look at how VSCode is showing the keys.
3. The keyboard shortcuts dialog should be accessible from somewhere other than `Ctrl + /`.

packages/common/src/utils/keybindings.ts
```diff
@@ -0,0 +1,490 @@
...
+    grouped[binding.category].push({
+      keys: Array.isArray(keys) ? keys : [keys],
+      description: binding.description
+    });
+  }
+
+  return grouped;
+}
+
+export function getGroupedTableKeybindingsMarkdown(): string {
```
thecodrr: I don't think this should be here. It should be in the script itself.

document-keyboard-shortcuts.script.ts
thecodrr, line 1: Let's move this script to `docs/help/scripts`. You will need to install `@notesnook/common` there. It should also be a .mjs file like all scripts.

packages/editor/src/hooks/use-editor.ts
```diff
@@ -37,6 +47,21 @@ export const useEditor = (
   const forceUpdate = useForceUpdate();
   const editorRef = useRef<TiptapEditor>(editor);
 
+  // TODO: remove
```
thecodrr: Remove when?

thecodrr (changes requested):

thecodrr (changes requested):
A few places that aren't using the new keybindings currently:

1. The command palette help shows hard coded shortcut keys
2. Menu items in editor show hard coded modifiers (check `src/toolbar/tools/*.tsx`)
3. `Mod-Alt-0` is hard coded in `paragraph.ts` extension
4. `Shift-Tab` to lift list item is also hard coded

packages/editor/src/hooks/use-editor.ts
thecodrr: @01zulfi what about this? It looks useless to me.

01zulfi: Yes, I need to just verify all the tiptap shortcuts one more time and then I'll remove it.

PR #7669, opened 2025-02-25
web: command palette fixes

### LINKED ISSUE #7666: Searching in the command palette causes styling bugs
### What happened?

- [ ] When searching for a recent command, the style code becomes visible. 

Image

- [ ] ~~Headings are searchable(is this a bug?).~~ Also the heading capitalization for "Recent" changes when matching letters.

Image

Search for a recent command in the command palette.
Search for the word "recent".

### Version

3.0.28-00250c6-web

### Platform/OS

Firefox

### Relevant log output

```shell

```

### LINKED ISSUE #7667: Executing a failed search in the command palette or quick open menu crashes app
### What happened?

Searching for a string that does not return any results and pressing enter causes the app to crash. Same error occurs in both command palette and quick open.

Image

Search for a string that returns no results.

### Version

3.0.28-00250c6-web

### Platform/OS

Firefox

### Relevant log output

```shell
h is undefined
onKeyDown@https://app.notesnook.com/assets/app-D-3-GnP3J2jH.js:206:61987
xI@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:9855
EI@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:10009
_I@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:10066
ug@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:31446
f7@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:31865
h2/<@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:36778
E6@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:293:36935
D9@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:8991
h2@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:33152
J3@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:17362
$I@https://app.notesnook.com/assets/root-BRUvyIlaax7M.js:290:17144
```

01zulfi:
* core: add getIdentifier in fuzzy; this allows the caller to provide a custom indentifier instead of relying on id field in the items list
* fix crashing when hitting enter when no results found
* fix commands list not refreshing when removing recent command
* fix recent command not showing up in the commands list
* fix searched commands showing up under wrong group label

Closes #7666 
Closes #7667

### DIFF
--- a/apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
+++ b/apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
@@ -69,7 +69,8 @@ export const CommandPaletteDialog = DialogManager.register(
 
     const commands = usePromise(async () => {
       select(0);
-      defaultCommands.current = await getDefaultCommands();
+      if (!defaultCommands.current)
+        defaultCommands.current = await getDefaultCommands();
       const commands = props.isCommandMode
         ? sortCommands(commandSearch(query, defaultCommands.current))
         : await dbSearch(query);
@@ -280,9 +281,10 @@ export const CommandPaletteDialog = DialogManager.register(
                   {command.group === "recent" && (
                     <Button
                       title={strings.removeFromRecents()}
-                      onClick={(e) => {
+                      onClick={async (e) => {
                         e.stopPropagation();
                         removeRecentCommand(command.id);
+                        defaultCommands.current = await getDefaultCommands();
                         commands.refresh();
                       }}
                       variant="secondary"
--- a/apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
+++ b/apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
@@ -69,10 +69,9 @@ export const CommandPaletteDialog = DialogManager.register(
 
     const commands = usePromise(async () => {
       select(0);
-      if (!defaultCommands.current)
-        defaultCommands.current = await getDefaultCommands();
+      defaultCommands.current = await getDefaultCommands();
       const commands = props.isCommandMode
-        ? commandSearch(query, defaultCommands.current)
+        ? sortCommands(commandSearch(query, defaultCommands.current))
         : await dbSearch(query);
       const groups = commands.reduce(
         (acc, command) => {
@@ -120,6 +119,7 @@ export const CommandPaletteDialog = DialogManager.register(
             if (e.key == "Enter") {
               e.preventDefault();
               const command = commands.value.commands[selected];
+              if (!command) return;
               command.action?.(command, {
                 openInNewTab: e.ctrlKey || e.metaKey
               });
@@ -350,6 +350,7 @@ function commandSearch(query: string, commands: Command[]) {
   return fuzzy(
     query,
     commands,
+    (command) => command.id + command.group,
     {
       title: 10,
       group: 5
@@ -419,6 +420,7 @@ function useDatabaseFuzzySearch() {
         for (const item of fuzzy(
           query,
           items,
+          (item) => item.id,
           {
             title: 10
           },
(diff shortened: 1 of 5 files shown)


packages/core/src/utils/fuzzy.ts
```diff
@@ -20,9 +20,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import { match, surround } from "fuzzyjs";
 import { clone } from "./clone";
 
-export function fuzzy<T extends { id: string }>(
+export function fuzzy<T>(
   query: string,
   items: T[],
+  getIdentifier: (item: T) => string,
```
01zulfi, line 26: The first bug highlighted here https://github.com/streetwriters/notesnook/issues/7666 happens because the `id` of recent and default commands match, and it gets incorrectly identified as the same command by `fuzzy`. I've added a `getIdentifier` param to fix this. Not sure if this is the best fix

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -529,3 +531,27 @@ function getCommandPaletteHelp(isCommandMode: boolean) {
...
+    }
+    const index = sortedWrtGroups.findIndex((c) => c[0].group === group);
+    if (index === -1) {
+      sortedWrtGroups.push([command]);
+    } else {
+      sortedWrtGroups[index].push(command);
+    }
+  }
+  return recent.concat(sortedWrtGroups.flat());
+}
```
thecodrr, line 557: You can just `.sort` by `group` instead of this custom logic.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -69,10 +69,9 @@ export const CommandPaletteDialog = DialogManager.register(
 
     const commands = usePromise(async () => {
       select(0);
-      if (!defaultCommands.current)
-        defaultCommands.current = await getDefaultCommands();
+      defaultCommands.current = await getDefaultCommands();
```
thecodrr, line 72: Why this change? We don't want to get default commands over and over again.

apps/web/src/dialogs/command-palette/commands.ts
```diff
@@ -234,7 +234,9 @@ export async function getDefaultCommands(): Promise<Command[]> {
 }
 
 export function getCommandById(id: string): Command | undefined {
-  return staticCommands.find((command) => command.id === id);
+  return staticCommands
+    .concat(getEditorCommands())
+    .find((command) => command.id === id);
```
thecodrr, line 239: Don't concat. Just search separately. It'll be more efficient.

thecodrr (changes requested):

thecodrr:
@01zulfi I think in future we should create separate PRs for each fix. That way the fixes can be cherry picked instead of merging it in bulk.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: We need the `getDefaultCommands` to re-run after a recent is removed from the user. Thanks for pointing this out, I realized we can add
```ts
defaultCommands.current = await getDefaultCommands()
``` 
in the click handler to fix this, instead of the above change

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
01zulfi: I'm facing a challenge where `.sort` respects the order of the searched elements, need to figure this out

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
thecodrr: Good idea.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
thecodrr: Were you able to figure something out? If not, we can keep this I think.

01zulfi: Nope

thecodrr (approved):

PR #7692, opened 2025-02-27
web: pinned tab fixes

### LINKED ISSUE #7683: Tab duplicated erroneously
### What happened?

When I'm in a pinned tab and click on a note that already has a tab open, a new tab opens.

to repair

### Version

3.0.28-00250c6-desktop (Appimage)

### Platform/OS

Linux

### Relevant log output

```shell

```

01zulfi:
* fix pinned tab not working after page reload
* fix note opening in new tab even if its tab is present when active note is pinned

Closes #7683

### DIFF
--- a/apps/web/src/stores/editor-store.ts
+++ b/apps/web/src/stores/editor-store.ts
@@ -679,10 +679,12 @@ class EditorStore extends BaseStore<EditorStore> {
     const noteId = typeof noteOrId === "string" ? noteOrId : noteOrId.id;
     const oldTabForNote = options.force ? null : getTabsForNote(noteId).at(0);
     const activeTab = getActiveTab();
+    const isReactivatingTab =
+      options.force && getTabsForNote(noteId).some((t) => t.id === activeTabId);
     const tabId =
+      // if a tab is pinned then always open a new tab.
       options.openInNewTab ||
-      // if a tab is pinned then always open a new tab, unless there's an existing tab for the note
-      (options.considerPinnedTab && activeTab?.pinned && !oldTabForNote)
+      (activeTab?.pinned && !oldTabForNote && !isReactivatingTab)
         ? addTab(getId())
         : oldTabForNote?.id || activeTabId || addTab(getId());
 
--- a/apps/web/src/stores/editor-store.ts
+++ b/apps/web/src/stores/editor-store.ts
@@ -659,6 +659,10 @@ class EditorStore extends BaseStore<EditorStore> {
       activeBlockId?: string;
       silent?: boolean;
       openInNewTab?: boolean;
+      /**
+       * Should be true if we want to open a new tab when the active tab is pinned
+       */
+      considerPinnedTab?: boolean;
     } = {}
   ): Promise<void> => {
     const {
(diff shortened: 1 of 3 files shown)


apps/web/src/stores/editor-store.ts
```diff
@@ -659,6 +659,10 @@ class EditorStore extends BaseStore<EditorStore> {
       activeBlockId?: string;
       silent?: boolean;
       openInNewTab?: boolean;
+      /**
+       * Should be true if we want to open a new tab when the active tab is pinned
+       */
+      considerPinnedTab?: boolean;
```
thecodrr, line 665: Why is this necessary?

thecodrr (changes requested):

apps/web/src/stores/editor-store.ts
01zulfi: This method, `openSession`, can be called in various situations. In these situations, `activeTab` doesn't have a consistent definition. For example, `openSession` is called here:
1. Opening a note on click on the notes list. When the method is called here, `activeTab` is the tab that is currently active in the editor NOT the tab we're trying to open.
2. Focusing a tab. When the method is called here, `activeTab` is the tab we're trying to open. If that tab is already pinned, I'll never be able to open the tab this way at all, it will always open a new tab.

The current logic to open in new tab for pinned active tab is flawed in this sense as `openSession` method doesn't have a way to differentiate between those two scenarios. I think the current approach where the caller of `openSession` decides to consider active tab is okay for now, the only other fix I could think of was maybe a refactor of several `editor-store` methods.

Do let me know what do you think about this

thecodrr: I am not seeing any issue without this flag.

> Focusing a tab. When the method is called here, activeTab is the tab we're trying to open. If that tab is already pinned, I'll never be able to open the tab this way at all, it will always open a new tab.

No, `activeTab` is that the tab that's already active. It's not the tab we are trying to open.

01zulfi: > I am not seeing any issue without this flag.

Pin a tab, and reload the app. The pinned tab cannot be opened at all, whether we try to click it or use the kbd shortcut

> No, activeTab is that the tab that's already active. It's not the tab we are trying to open.

I see, I'll check this again. 

---

Maybe the issue is with the `needHydration` field of the session? Not sure

thecodrr: Okay yes, that's definitely an issue.

thecodrr: Okay so when we rehydrate the session, we set `force` to true. Which makes `oldTabForNote` null. Hm...

thecodrr:
@01zulfi can you finalize this?

apps/web/src/stores/editor-store.ts
```diff
@@ -676,8 +680,9 @@ class EditorStore extends BaseStore<EditorStore> {
     const oldTabForNote = options.force ? null : getTabsForNote(noteId).at(0);
     const activeTab = getActiveTab();
     const tabId =
-      // if a tab is pinned then always open a new tab.
-      options.openInNewTab || activeTab?.pinned
+      options.openInNewTab ||
+      // if a tab is pinned then always open a new tab, unless there's an existing tab for the note
+      (options.considerPinnedTab && activeTab?.pinned && !oldTabForNote)
```
01zulfi, line 685: Another way of doing it could be by always defining the `oldTabForNote` variable. This does fix the pinned tab not opening bug, but introduces another one: if there is a note open in two tabs, trying to open the second tab will always open the first tab (this occurs after reloading the page)

```ts
    const oldTabForNote = getTabsForNote(noteId).at(0);
    const activeTab = getActiveTab();
    const tabId =
      options.openInNewTab ||
      (activeTab?.pinned && !oldTabForNote)
```

^ is this approach worth exploring, @thecodrr?

PR #7736, opened 2025-03-04
web: add toggle focus mode command

01zulfi:

### DIFF
--- a/apps/web/src/dialogs/command-palette/commands.ts
+++ b/apps/web/src/dialogs/command-palette/commands.ts
@@ -32,12 +32,15 @@ import {
   Notebook as NotebookIcon,
   Note as NoteIcon,
   Reminder as ReminderIcon,
-  Tag as TagIcon
+  Tag as TagIcon,
+  FocusMode,
+  NormalMode
 } from "../../components/icons";
 import { trashMenuItems } from "../../components/trash-item";
 import { hashNavigate, navigate } from "../../navigation";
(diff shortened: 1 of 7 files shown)


apps/web/src/dialogs/command-palette/commands.ts
```diff
@@ -440,6 +444,26 @@ function getEditorCommands(): Command[] {
...
+    commands.push({
+      id: "switch-to-focus-mode",
+      title: strings.focusMode(),
+      icon: NormalMode,
+      action: () => useAppStore.getState().toggleFocusMode(),
+      group: strings.editor(),
+      type: "command"
+    });
+  }
+
```
thecodrr, line 466: We can use a single command here: `Toggle focus mode`.

thecodrr (changes requested):

apps/web/src/dialogs/command-palette/commands.ts
01zulfi: I've opted to keep the conditional icon

```ts
 icon: isFocusMode ? FocusMode : NormalMode,
 ```
 
 Let me know if the icon should be static, or is this fine

thecodrr:
@01zulfi can you resolve the conflicts so this can be merged?

samuelort:
Great addition! Very much appreciate the command palette. Any thoughts of adding the insert commands to it as well? 

bild

PR #7758, opened 2025-03-10
web: fix list pane not expanding

01zulfi:
* when a nav item was clicked, list pane was removed entirely which made it difficult to re-open it, now when nav item is clicked we manually collapse, expand, or reset the list pane
* open list pane if 'reveal in list' is called
* open list pane if navigated to a different nav item

### DIFF
--- a/apps/web/src/app.tsx
+++ b/apps/web/src/app.tsx
@@ -147,16 +147,14 @@ function DesktopAppContents() {
   useEffect(() => {
     if (isListPaneVisible) {
       navPane.current?.expand(1);
-    } else {
-      if (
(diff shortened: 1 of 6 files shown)


ammarahm-ed:
@01zulfi One thing we want to do here is that when you click on a Navigation Item like `Notes` or `Notebooks` it should close the side-bar automatically when the hidden list pane opens.

01zulfi:
> @01zulfi One thing we want to do here is that when you click on a Navigation Item like Notes or Notebooks it should close the side-bar automatically when the hidden list pane opens.

Done

ammarahm-ed (approved):

PR #7759, opened 2025-03-10
web: fix editor rerendering on session type change from new to default

### LINKED ISSUE #7737: Cursor changes position and letters are dropped while typing during note creation
### What happened?

When creating a new note it starts out blank, with no name, and no entry in the notes list. The note only gets 'created' once I start typing. However, it takes about a quarter of a second create the new note, in which time I have usually typed 4 to 6 characters. During the note creation process the cursor is reset to the beginning of the note multiple times. If I started typing the word "hello", it ends up looking like "olhel", with the cursor *before* the first L. 

Once the note creation process completes, the interface becomes stable, and typing works just fine.

Also, sometimes letters don't register, and despite typing "hello" all I get is "ohel" or some variation thereof.

Open the webapp, create a new note, start typing a word into the new note, and don't stop typing while the note gets created.

### Version

3.0.29-9b6f7ff-web

### Platform/OS

Firefox, Chrome/Chromium

### Relevant log output

```shell

```

ammarahm-ed:
Closes #7737

### DIFF
--- a/apps/web/src/components/editor/index.tsx
+++ b/apps/web/src/components/editor/index.tsx
@@ -224,13 +224,14 @@ export default function TabsView() {
   );
 }
 
-const MemoizedEditorView = React.memo(
-  EditorView,
-  (prev, next) =>
+const MemoizedEditorView = React.memo(EditorView, (prev, next) => {
+  return (
     prev.session.id === next.session.id &&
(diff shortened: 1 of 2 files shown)


01zulfi (approved):
looks good to me 🚀

PR #7767, opened 2025-03-13
web: increase reminder dialog width

01zulfi:

### DIFF
--- a/apps/web/src/dialogs/add-reminder-dialog.tsx
+++ b/apps/web/src/dialogs/add-reminder-dialog.tsx
@@ -156,6 +156,7 @@ export const AddReminderDialog = DialogManager.register(
         testId="add-reminder-dialog"
         onClose={() => props.onClose(false)}
         sx={{ fontFamily: "body" }}
+        width={600}
         positiveButton={{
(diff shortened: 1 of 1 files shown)


apps/web/src/dialogs/add-reminder-dialog.tsx
```diff
@@ -156,6 +156,7 @@ export const AddReminderDialog = DialogManager.register(
         testId="add-reminder-dialog"
         onClose={() => props.onClose(false)}
         sx={{ fontFamily: "body" }}
+        width={600}
```
ammarahm-ed, line 159: Is this a fixed width? what happens on mobile where the total width might be around 450-500?

01zulfi: This width is only for large screens. For medium and small sizes, it will use 90% and 100% width respectively:

https://github.com/streetwriters/notesnook/blob/master/apps/web/src/components/dialog/index.tsx#L100

ammarahm-ed: Okay

PR #7780, opened 2025-03-18
web: don't remove properties pane when editor is refocused

### LINKED ISSUE #7770: When I click on the body of the text while the TOC is on, the TOC turns off.

When I click on the body of the text while the TOC is on, the TOC turns off.

3.1.0-beta.2-fb387f6-web

Please make it so that the TOC window on the right does not disappear when clicking on the body text while TOC is turned on.

_No response_

### Additional context

_No response_

01zulfi:

Closes #7770

### DIFF
--- a/apps/web/src/components/editor/index.tsx
+++ b/apps/web/src/components/editor/index.tsx
@@ -252,7 +252,6 @@ function EditorView({
   const lastChangedTime = useRef<number>(0);
   const root = useRef<HTMLDivElement>(null);
 
-  const toggleProperties = useEditorStore((store) => store.toggleProperties);
   const isFocusMode = useAppStore((store) => store.isFocusMode);
   const editor = useEditorManager((store) => store.editors[session.id]?.editor);
 
@@ -376,7 +375,6 @@ function EditorView({
         }}
         options={{
           readonly: session?.type === "readonly" || session?.type === "deleted",
-          onRequestFocus: () => toggleProperties(false),
           focusMode: isFocusMode
         }}
       />


apps/web/src/components/editor/index.tsx
```diff
@@ -382,7 +382,6 @@ function EditorView({
         }}
         options={{
           readonly: session?.type === "readonly" || session?.type === "deleted",
-          onRequestFocus: () => toggleProperties(false),
```
ammarahm-ed, line 385: We want to hide properties pane when you focus the editor, however for TOC we want to keep it open.

ammarahm-ed (changes requested):

apps/web/src/stores/editor-store.ts
```diff
@@ -1163,12 +1163,17 @@ class EditorStore extends BaseStore<EditorStore> {
...
+    toggleTableOfContents?: boolean
+  ) => {
     this.set((state) => {
       state.arePropertiesVisible =
         toggleState !== undefined ? toggleState : !state.arePropertiesVisible;
     });
-    this.toggleTableOfContents(false);
+    if (toggleTableOfContents !== false) {
+      this.toggleTableOfContents(false);
+    }
```
ammarahm-ed, line 1176: In what case we want to toggle table of contents with properties? I think this can be removed as it's not needed anymore. There is a dedicated function to toggle TOC already which should be used.

01zulfi: Whenever properties is toggled, it closes the table of contents, this ensures that only one pane is open at a time

ammarahm-ed: Let's keep the existing behavior and:

1. If you open TOC, it should close properties if already opened
2. If you close properties, do nothing with the TOC.
3. If you open properties, do nothing with the TOC.

From what I can tell, it is a bug in the beta app because the stable version behaves correctly. The TOC never closes unless you close it by clicking X.

ammarahm-ed: Just trying the beta app again, it seems that the UI has changed quite a bit and now properteis and toc are part of the editor, not floating on top. Both show in the same container I think?

01zulfi: They are separate panes

01zulfi: @ammarahm-ed do let me know what needs to be done

apps/web/src/components/editor/index.tsx
```diff
@@ -382,7 +382,9 @@ function EditorView({
...
+          onRequestFocus: () => {
+            toggleProperties(false, false);
+          },
```
thecodrr, line 387: Instead of all these changes, let's just remove this "close on focus" logic altogether.

thecodrr (changes requested):

PR #7805, opened 2025-03-27
web: don't close pinned tab on ctrl+w

### LINKED ISSUE #7776: Pinned notes should not be closed by Ctrl+w

When closing a lot of tabs I go berserk with `ctrl+w` and I always think that this will not close the pinned notes but it does.

If you have a note pinned then it should not be closed unless you unpin it.

01zulfi:

Closes #7776

### DIFF
--- a/apps/web/src/common/key-map.ts
+++ b/apps/web/src/common/key-map.ts
@@ -98,7 +98,11 @@ const KEYMAP = [
   {
     keys: ["ctrl+w", "command+w"],
     description: "Close active tab",
-    action: () => useEditorStore.getState().closeActiveTab()
+    action: () => {
+      const activeTab = useEditorStore.getState().getActiveTab();
+      if (activeTab?.pinned) return;
+      useEditorStore.getState().closeActiveTab();
+    }
   },
   {
     keys: ["ctrl+shift+w", "command+shift+w"],
--- a/apps/web/src/common/key-map.ts
+++ b/apps/web/src/common/key-map.ts
@@ -97,10 +97,14 @@ const KEYMAP = [
   },
   {
     keys: ["ctrl+w", "command+w"],
-    description: "Close active tab",
+    description:
+      "Close active tab or focus previously activated tab if active tab pinned",
     action: () => {
       const activeTab = useEditorStore.getState().getActiveTab();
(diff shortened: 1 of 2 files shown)


apps/web/src/stores/editor-store.ts
```diff
@@ -1101,9 +1101,11 @@ class EditorStore extends BaseStore<EditorStore> {
...
 
   closeActiveTab = () => {
-    const { activeTabId } = this.get();
-    if (!activeTabId) return;
-    this.closeTabs(activeTabId);
+    const activeTab = this.getActiveTab();
+    if (!activeTab) return;
+    if (activeTab.pinned) return;
+
+    this.closeTabs(activeTab.id);
```
thecodrr, line 1108: Not the right place to do this. We should handle this especially in the `ctrl+w` handler because `closeActiveTab` could be used in other places where we do want to close the tab even if it is pinned.

thecodrr (changes requested):

apps/web/src/common/key-map.ts
```diff
@@ -98,7 +98,11 @@ const KEYMAP = [
   {
     keys: ["ctrl+w", "command+w"],
     description: "Close active tab",
-    action: () => useEditorStore.getState().closeActiveTab()
+    action: () => {
+      const activeTab = useEditorStore.getState().getActiveTab();
+      if (activeTab?.pinned) return;
```
thecodrr, line 103: Let's follow the behavior of other apps (e.g. VSCode, Zed). Pressing Ctrl + W on a pinned tab activates the previously activated tab instead of doing nothing.

thecodrr (changes requested):

apps/web/src/stores/editor-store.ts
```diff
@@ -850,6 +850,16 @@ class EditorStore extends BaseStore<EditorStore> {
     return this.focusTab(tabs[index === 0 ? tabs.length - 1 : index - 1].id);
   };
 
+  focusPreviouslyActivatedTab = () => {
```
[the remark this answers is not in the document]
01zulfi: `focusPreviousTab` is already taken, going with `focusLastActiveTab

PR #7941, opened 2025-04-18
web: fix notebooks tree

01zulfi:
* fix flicker when creating a new notebook/subnotebook
* fix crash when creating subnotebook
* fix notebook not expanding when subnotebook is created against it

### DIFF
--- a/apps/web/src/components/virtualized-tree/index.tsx
+++ b/apps/web/src/components/virtualized-tree/index.tsx
@@ -30,7 +30,11 @@ import { CustomScrollbarsVirtualList, waitForElement } from "../list-container";
 
 export type VirtualizedTreeHandle<T> = {
   refresh: () => void;
-  refreshItem: (index: number, item?: T) => void;
+  refreshItem: (
+    index: number,
+    item?: T,
+    itemOptions?: { expand?: boolean }
+  ) => void;
 };
 export type TreeNode<T = any> = {
   id: string;
@@ -95,7 +99,7 @@ export function VirtualizedTree<T>(props: TreeViewProps<T>) {
     treeRef,
     () => ({
       async refresh() {
-        setNodes([]);
+        // setNodes([]);
         const { children } = await fetchChildren(
           rootId,
           -1,
@@ -105,7 +109,7 @@ export function VirtualizedTree<T>(props: TreeViewProps<T>) {
         setNodes(children);
         setExpandedIds(expandedIds);
       },
-      async refreshItem(index, item) {
+      async refreshItem(index, item, itemOptions) {
         const node = nodes[index];
         const removeIds: string[] = [node.id];
         for (const treeNode of nodes) {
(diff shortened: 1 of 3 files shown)


apps/web/src/components/virtualized-tree/index.tsx
```diff
@@ -95,7 +99,7 @@ export function VirtualizedTree<T>(props: TreeViewProps<T>) {
     treeRef,
     () => ({
       async refresh() {
-        setNodes([]);
+        // setNodes([]);
```
thecodrr, line 102: If we don't do this, the items duplicate during filter.

01zulfi: I see.

When a notebook/subnotebook is created, both `refresh` and `refreshItem` is triggered. Nodes are reset in `refresh` in the above line, but it doesn't get populated again before `refreshItem` tries to access nodes. Which causes an error and crashes the app.

I don't fully understand why removing the above line causes notebooks to be duplicated while filtering. I've created a separate `resetAndRefresh` method for filtering notebook usecase which does fix the issue.

thecodrr: > I don't fully understand why removing the above line causes notebooks to be duplicated while filtering.

It's something to do with virtuouso not removing old items from the list.

> I've created a separate resetAndRefresh method for filtering notebook usecase which does fix the issue.

Sounds good.

apps/web/src/components/virtualized-tree/index.tsx
```diff
@@ -120,8 +136,7 @@ export function VirtualizedTree<T>(props: TreeViewProps<T>) {
           return;
         }
 
-        // TODO: double check
-        if (node.hasChildren) {
+        if (itemOptions?.expand) {
```
thecodrr, line 131: What if the node doesn't have children?

thecodrr (changes requested):

apps/web/src/components/virtualized-tree/index.tsx
01zulfi: The `node.hasChildren` check to decide whether to expand the node is faulty. I think I added it way back but never came back to it again. 

`node.hasChildren` check assumes that the node should always be expanded if it has children. Whereas we only want to expand when a subnotebook is created against the node. `refreshItem` doesn't have enough information/context to check for this case, so I'm passing it in as a param.

thecodrr: Yes, but we should still check it in addition to `itemOptions.expand`.

01zulfi: Done, I have to move the check after we have fetched the latest children of the node otherwise `node.hasChildren` doesn't have updated information

thecodrr: > I have to move the check after we have fetched the latest children

But that way the notebook won't expand because `expandedIds` would be updated after we have fetched the children.

01zulfi: It does expand because we update `expandedIds` before it is passed in `setExpandedIds` as updated state

thecodrr: Yes but `expandedIds` is also used in `fetchChildren` for prefetching a node's children if its expanded.

thecodrr (approved):

PR #7950, opened 2025-04-21
web: show reset search button only when query text present

01zulfi:

### DIFF
--- a/apps/web/src/components/field/index.tsx
+++ b/apps/web/src/components/field/index.tsx
@@ -31,6 +31,7 @@ type Action = {
   disabled?: boolean;
   icon?: Icon;
   component?: JSX.Element;
+  hidden?: boolean;
 };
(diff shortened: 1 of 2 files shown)


apps/web/src/components/field/index.tsx
```diff
@@ -203,32 +204,36 @@ function Field(props: FieldProps) {
...
-                disabled={action.disabled}
-              >
-                {action.component ? (
-                  action.component
-                ) : action.icon ? (
-                  <action.icon size={16} />
-                ) : null}
-              </Button>
+              <>
+                {action.hidden ? null : (
```
thecodrr, line 208: Maybe we can filter out the `rightActions` instead of a condition here. Also this should be handled in `leftActions` too.

thecodrr (changes requested):

thecodrr (approved):

PR #7968, opened 2025-04-23
core: segregate logic for note headline and note headline title generation

### LINKED ISSUE #7970: The first line in a note is no longer the inly line that shows under the note's name in the notes view
Writing the first line then clicking enter used to have that first line only show up in the notes view under the note's name, now it shows all the lines stuc together which isn't useful, having it only show the very first line was much useful for a lot of reasons to give some information about certain notes.

Also, a quick thinf that I also noticed, swiping left to close the left panel is overly stiff currently and requires a lot of extra movement to turn off, it was much faster and easier before, I like the new UI but only those 2 things are an issue to me, is it possible to fix or at least have the first line of a note thing as a settings option? It's very important to me, please 

_______________
**Device information:**
App version: 3.1.0
Platform: android
Device: samsung-SM-G998B-14
Pro: true
Logged in: yes

01zulfi:

Closes #7970

### DIFF
--- a/packages/core/src/collections/notes.ts
+++ b/packages/core/src/collections/notes.ts
@@ -96,8 +96,8 @@ export class Notes implements ICollection {
         const content = await getContentFromData(type, data);
         if (!content) throw new Error("Invalid content type.");
 
-        headline = getNoteHeadline(content);
-        headlineTitle = getNoteHeadlineTitle(content);
+        headline = content.toHeadline();
+        headlineTitle = content.toTitle();
         dateEdited = Date.now();
         contentId = await this.db.content.add({
           noteId: id,
@@ -149,7 +149,7 @@ export class Notes implements ICollection {
         if (
           item.isGeneratedTitle &&
           HEADLINE_REGEX.test(titleFormat) &&
-          headline &&
+          headlineTitle &&
           currentNoteTitleFields?.title !== headlineTitle
         ) {
           item.title = titleFormat.replace(HEADLINE_REGEX, headlineTitle);
@@ -486,12 +486,4 @@ export class Notes implements ICollection {
       "internalLinks"
     ).internalLinks;
   }
-}
-
-function getNoteHeadline(content: Tiptap) {
-  return content.toHeadline();
-}
-
-function getNoteHeadlineTitle(content: Tiptap) {
-  return content.toHeadlineTitle();
 }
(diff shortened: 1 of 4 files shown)


packages/core/src/utils/html-parser.ts
```diff
@@ -44,7 +44,38 @@ function wrapIntoHTMLDocument(input: string) {
...
+    {
+      lowerCaseTags: false,
+      decodeEntities: true
+    }
+  );
+  parser.end(html);
+  return text;
+}
+
+export function extractHeadlineTitle(
```
thecodrr, line 75:
```suggestion
export function extractTitle(
```

packages/core/src/content-types/tiptap.ts
```diff
@@ -89,7 +90,11 @@ export class Tiptap {
   }
 
   toHeadline() {
-    return extractHeadline(this.data, 150);
+    return extractHeadline(this.data);
+  }
+
+  toHeadlineTitle() {
```
thecodrr, line 96:
```suggestion
  toTitle() {
```

packages/core/src/collections/notes.ts
```diff
@@ -493,6 +492,6 @@ function getNoteHeadline(content: Tiptap) {
   return content.toHeadline();
 }
 
-function headlineToTitle(headline: string) {
-  return headline.split(" ").splice(0, 10).join(" ");
+function getNoteHeadlineTitle(content: Tiptap) {
+  return content.toHeadlineTitle();
 }
```
thecodrr, line 497: No need for this.

thecodrr (changes requested):

packages/core/src/collections/notes.ts
```diff
@@ -148,12 +150,9 @@ export class Notes implements ICollection {
           item.isGeneratedTitle &&
           HEADLINE_REGEX.test(titleFormat) &&
           headline &&
```
thecodrr, line 152: Should this be here?

thecodrr (changes requested):

packages/core/src/collections/notes.ts
01zulfi: It should be `headlineTitle`, fixing...

thecodrr (approved):

RektalTk:
This fix is not working on version 3.1.1.

PR #7972, opened 2025-04-23
Always include email as 2FA fallback

### LINKED ISSUE #7745: Recovery option when 2fa device/codes are lost

I recently got my mobile device factory reset by a google system UI command activated mistakenly, sent by google to my android. This resulted in me loosing all my files and included in that, Aegis vault and every instance that could have some data to recover the 2fa. 2fa was enabled on my notesnook app which also got lost together with backup. In the process of gaining access to my diffrent accounts, i stumbled upon a good and secure feature - outlined below

I refere to @thecodrr 's own words "As a service it is our job to provide rescue when things go wrong, not the user's"  

There is currently no streamlined feature for doing account recovery when loosing 2fa, even though from a data science perspective, recovery of the data is fully possible when having email, password to Notesnook and access to email, as 2fa is not taking part in the encryption of the data itself, or generation of the salt, as far as I can see. Loosing 2fa is less likely than loosing your password, bu caamt possible factory reset commands from google and incomplete backup prompts a need for a more robust recovery system

Adding a feature that resets 2fa, but maintain security standards, is fully possible, and we will look to Proton Pass for a good example

Their solution is that a reset password functionality automatically disables 2FA, and the existing data on the account becomes unreadable after the password reset. One has to manually go in and recover/unlock the encrypted data by inputting either the old password or recovery codes

This feature would stay true to the encrypted nature and high security standard of Notesnook, and add crutial functionality - taking care of it's users 

Every possible way to try to get Aegis vault has been tried unsucessfully

### Additional context

Here is my case in detail, please help me recover access to my notesnook account: https://www.reddit.com/r/Notesnook/comments/1j4ldw0/access_to_account_with_lost_2fa/

KvalheimRacing:
Closes #7745

This PR:
- Web (`auth.tsx`): tweaks isValidMethod to include `"email"`
- Mobile (`two-factor.js`): tweaks getMethods to include `"email"`
- No API changes needed; behavior is purely client-side.

Verifying steps:
1. Web: run `npm run start:web` and confirm “Send code by Email” shows.
2. Android: download debug APK from CI, install, then confirm on-device.


thecodrr:
@KvalheimRacing thank you for the PR! Unfortunately, there are critical security implications if we merge this. I am also not sure if there's a need for this since you can already set email as a fallback 2FA method.

KvalheimRacing:
@thecodrr Thank you for your response. The issue is serious, as activating another 2fa method removes now the 2fa fallback method, please see through issue #7745, as @Leviob also observed:

```
@thecodrr, I was able to reproduce this and agree that this is a serious issue that should be fixed.

Steps I took:

Create an account
Verify email
Add a 2fa app
Logout
I only have the option to use the 2fa app or recovery codes.

If you follow the same steps as above, but do not add a 2fa app, the default behavior is to use email as a second factor. In this scenario I have the option to send a code to my email. I agree that it should always be an option to use email as a second factor. It is very counterintuitive that the email 2fa method would stop working when adding a 2fa app. 
```

My issue now is that I don't have access to my data, although I still have access to my email and remember my password. Many people have issues like this, and disabeling the email recovery option is counter intuitive and not a good practice. There should always be an option to recover an account with email, you have said this youself in blog posts. Please read through the entire #7745, you wrote:

"Resetting your Notesnook password should reset your 2FA. You will still need access to your email"

I have no such optionality, and I have no way to get into my account, although it is fully possible, the salt is there, I have access to my email, I have my password and so on. 2fa is not part of encryption and my data is not lost. I am just hindered to get access to them by current lacking functionality. I have very important data that I really need to use, and I would apprechiate any help in this regard

I refere to your own words Abdullah where you write: "As a service it is our job to provide rescue when things go wrong, not the user's"

Please read through, please help

Leviob:
@thecodrr, I agree that there is a serious need for this. The current setup will permanently lock people out of their accounts without warning when they add a 2FA device (and then lose that device or delete the 2FA seed).

I believe an email should always be available as a 2FA option. Do you disagree? What are the critical security implications of this? 

If someone wants to CHOOSE to remove email as a 2FA option to increase their security, that should be a choice. The current setup silently makes that choice for the user and is very counter-intuitive compared with every other app that I have used.

thecodrr:
The way it works is as follows:

- By default, all users have email as their 2FA method. This is mandatory and cannot be disabled.
- When a user changes/sets another 2FA method, it replaces the default 2FA method (i.e. email).
- A user can always choose to set email as their backup/fallback 2FA method afterwards.

The problem with always allowing email as 2FA is that any security structure is only as strong as its weakest link. Email _can_ be considered a weak link if its always set as default allowing someone to break into your account and/or download your encrypted data. Of course, if a user opts into (as opposed to opt out) such a behavior, that is up to them but as a privacy and security focused application, we cannot choose a less secure default.

> "Resetting your Notesnook password should reset your 2FA. You will still need access to your email"

This is true but you are missing out on one thing; resetting your password requires you to provide an encryption key (OR if you don't have that you can clear your data). You cannot, for example, reset your password with just access to an email.

 > I have no way to get into my account, although it is fully possible, the salt is there, I have access to my email, I have my password and so on. 2fa is not part of encryption and my data is not lost. I

The whole point of 2FA is to make it _harder_ for an attacker to get into your account. If you lose access to your 2FA method then you essentially become an attacker as in it becomes super hard to distinguish whether it really is you trying to access your account and data. An attacker _could_ have access to your email so we can obviously _not_ use that as any sort of identification.

> For example, some apps will have a "I lost my 2FA device, help!" button which will disable 2FA once they verify their email address ([Proton](https://proton.me/support/lost-two-factor-authentication-2fa), [LastPass](https://support.lastpass.com/s/document-item?language=en_US&bundleId=lastpass&topicId=LastPass/lost-phone-disable-mfa.html&_LANG=enus)), or enter a special code ([Bitwarden](https://bitwarden.com/help/lost-two-step-device/)).

In case of Bitwarden, they clearly state:

> If you don't have your recovery code saved somewhere outside of your vault, there is unfortunately no way for the team to recover the account or data therein. You will need to delete your account and start a new one.

We also provide a list of recovery codes when you set up 2FA.

Proton also doesn't allow anything of the sort _by default_. You can add a recovery email address but it is not added for you by default which is exactly what Notesnook does by giving you the option to set a fallback 2FA method.

> I refere to your own words Abdullah where you write: "As a service it is our job to provide rescue when things go wrong, not the user's"

I stand by my words but I am not willing to compromise Notesnook's security. If a user does not have the required credentials to get into their account then there's nothing we can do.

Leviob:
> as a privacy and security focused application, we cannot choose a less secure default.

I really appreciate that there is a default requirement for 2FA via email. The problem I have with the above opinion is that the email _already is the default_ 2FA method. By making the requested change, we're not lowering the security beyond what is already the default. (We're not removing 2FA, nor are we removing the password. If I had access to your email, I still wouldn't be able to log into your Notesnook account without a recovery key or your Notesnook password, right?) 

I agree with your argument about the point of adding a 2FA device being increasing the security, I just think the current flow will confuse people (including me) without a clear understanding of what the expected behavior is. Going forward, I see an appropriate solution being the addition of warnings that are very clear that adding a 2FA device will remove your current 2FA method option unless you manually add it again. I think people would prefer the explicit warning to accidentally locking themselves out of their account in the future. 

I still _personally_ think removing email as a 2FA method should be an extra step, rather than adding it being the extra step. But I concede that the current setup would be okay if there was more thorough verbiage in the UI.

I've seen some apps force the user to enter one of the recovery codes when they setup 2FA to make sure they actually saved the codes. It's an annoying extra step, but it helps prevent angry customers who don't understand security in the same way that you do. 

But that's not a solution for the unfortunate people who have already been directly affected by this. I guess this is a good reminder to backup your data often and audit your recovery methods.

KvalheimRacing:
@thecodrr, @Leviob, thank you both for your detailed responses and insights into the security considerations here

I completely understand and appreciate the focus Notesnook places on security and privacy; indeed, this is a key reason many users (including myself) trust the platform. However, the issue at hand is not about diminishing security but rather about enhancing user resilience in legitimate account recovery scenarios

Currently, enabling a new 2FA device silently removes email-based recovery without explicitly informing the user, leading to unexpected and severe lock-out scenarios—as has unfortunately happened in my case and confirmed reproducible by @Leviob. From a user-experience standpoint, this silent removal is counterintuitive and can and will cause significant distress

Your concern about email being a potential weakest link is valid; however, email is already the default 2FA method when users initially sign up. My proposed solution doesn't degrade the security model below its current default state—it merely maintains that existing baseline, preventing unintended permanent lock-outs. Users seeking heightened security can explicitly remove email recovery methods afterward if they choose to

Additionally, other platforms like Proton Pass offer similar user-centric recovery flows, explicitly informing users of potential security trade-offs. This approach ensures users are fully aware and responsible for their security choices, aligning perfectly with Notesnook’s user-empowering ethos

Therefore, I recommend:

- Clearly notifying users when enabling alternative 2FA methods about the removal of email fallback unless explicitly re-added

- By default, retaining email fallback to prevent accidental permanent lock-outs

- Optionally allowing advanced users to disable email-based fallback manually if they understand the implications and accept the risk

Additionally, introducing manual recovery methods could significantly enhance the recovery process, such as:

- Manual Verification Using Historical Data: Users could regain account access by correctly answering detailed questions about their usage history, such as creation dates of specific notes, note contents, attachment details, or account activity patterns

- Time-Locked Recovery: Initiate a recovery request that triggers a waiting period (e.g., 7 days). The user is notified multiple times via email during this period, allowing legitimate users to confirm recovery or cancel if unauthorized

- Video/Personal Verification: Allow users to request manual verification via a recorded video where they verbally state their username, registered email, and recent notes content. A support agent could then manually review and approve account recovery

Implementing these methods would offer secure yet flexible ways for users to recover their accounts in emergencies without substantially lowering security standards

You write @thecodrr :
> then you essentially become an attacker as in it becomes super hard to distinguish whether it really is you trying to access your account and data. An attacker could have access to your email so we can obviously not use that as any sort of identification
This assessment is wrong, as a manual verification method consisting of as many data points as one can gather would distinguish between an attacker and an account holder. The account holder could consent to notesnook accessing the encrypted data, and provide the email and password for verification of account note content

I sincerely ask you to reconsider this PR, or at minimum, provide additional recovery options. This issue impacts users directly and significantly - in my case insanely significantly, and a thoughtful, balanced solution is achievable without compromising the platform’s robust security posture

Thank you again for your attention and ongoing commitment to Notesnook’s excellence

thecodrr:
> Additionally, introducing manual recovery methods could significantly enhance the recovery process, such as:

We cannot offer such a method because:

a. We don't have access to a user's content or data aside from their email address
b. We don't ask the user to provide personal information such as an ID or a real picture which we can later use to verify
c. In my opinion, any such method is strongly in violation of users' privacy

> Clearly notifying users when enabling alternative 2FA methods about the removal of email fallback unless explicitly re-added

This can be added.

> By default, retaining email fallback to prevent accidental permanent lock-outs

I think this is unnecessary because the app already shows a list of recovery codes. However, we can improve the 2FA setup UX to automatically forward the user to setup a fallback 2FA method (skippable, of course).

> This assessment is wrong, as a manual verification method consisting of as many data points as one can gather would distinguish between an attacker and an account holder. The account holder could consent to notesnook accessing the encrypted data, and provide the email and password for verification of account note content

I think if we improve the UX and make 2FA setup and recovery super obvious, this should be unnecessary.

---

I think we can close this PR and work on a better fix.

KvalheimRacing:
**@thecodrr**, I appreciate your thoughtful reply, but I believe there are still a few misunderstandings that I'd like to clarify:

**Regarding manual recovery methods:**
I fully respect and agree with your dedication to maintaining the highest standards of privacy and security. However, a clearly consent-based rigourous manual recovery option does not inherently violate privacy. Users in desperate recovery scenarios (such as mine) could explicitly consent to the retrieval and inspection of encrypted data, temporarily providing access credentials such as email and password for identity verification. The critical point is requiring multiple verifiable data points. Such verification elements could be based on detailed historical data about the account activity, payment or device information, notes content and more, thereby clearly distinguishing legitimate account holders from potential attackers. These data points would be nearly impossible for an unauthorized person to replicate convinvingly. This consent-based approach is successfully used by various privacy-focused platforms precisely because it ensures user autonomy and control remain central

**Regarding retaining email fallback by default:**
Providing recovery codes at the initial setup stage doesn't resolve the silent removal of email fallback when adding a new 2FA method. The actual user-experience issue here is the absence of clear and explicit communication at the moment users add an alternative 2FA method. Improving the UX to automatically prompt users to explicitly confirm or re-add email fallback as a second factor would indeed be a helpful step

**Next Steps:**
I strongly encourage considering the addition of a clearly defined, consent-based rigorous manual recovery method, alongside any UX improvements. This dual approach would robustly protect user accounts from unintended permanent lock-outs, significantly enhancing the user experience without compromising security. I suggest clearly documenting and explicitly limiting such manual recovery options:

- They should only be triggered under highly controlled conditions

- They must require explicit user consent and detailed verification

- They should require a variety of datapoints

Implementing these precautions will significantly mitigate the risk of exploitation and maintain Notesnook's rigorous safety standards while addressing legitimate account access issues compassionately and practically

I truly value your openness to discussing these points and am confident we can find an effective solution together. Thank you again for your continued attention to this important matter

thecodrr:
@KvalheimRacing Thank you for your valuable input, we'll definitely discuss all these points and work on them if they align with Notesnook's principles and policies.

I am closing this PR for now. We can continue further discussion in the marked issue #7745.

PR #7976, opened 2025-04-24
web: save focus mode state in local storage

01zulfi:


thecodrr:
Why?

01zulfi:
I'd like the editor state to be persisted across page reloads. TOC state and editor margins already persist, so I think focus mode should follow suit

thecodrr:
Yes but focus mode is very different from TOC & editor margins. Why would you want the focus mode to persist when its only activated temporarily?

01zulfi:
I personally would want to have the focus mode open. I think this is an objective thing. Closing the PR for now

PR #8061, opened 2025-05-06
editor: fix search scroll not working on switching editors or opening new editor

### LINKED ISSUE #8036: Bug in search.
### What happened?

Search something in Notesnook (not a note). Make sure the word you are searching is included in 2 or more notes. Click on one note and search the word in the note. Click enter on the search bar and then click on the other note from the note list (do not cancel the search) now search the same word in second note. You will see that there is the word in the note but the note won't be scrolled to the word. If the word is mentioned more than once, you keep pressing enter in the search bar and it keeps increasing the number in `1`/4 (mentioned 4 times) but the note doesn't get scrolled.

. 

### Version

v3.2.0

### Platform/OS

Windows

### Relevant log output

```shell

```

01zulfi:

Closes #8036

### DIFF
--- a/packages/editor/src/toolbar/floating-menus/search-replace.tsx
+++ b/packages/editor/src/toolbar/floating-menus/search-replace.tsx
@@ -17,7 +17,6 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { useLayoutEffect } from "react";
 import { FloatingMenuProps } from "./types.js";
 import { SearchReplacePopup } from "../popups/search-replace.js";
 import { ResponsivePresenter } from "../../components/responsive/index.js";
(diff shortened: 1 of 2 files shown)


packages/editor/src/extensions/search-replace/search-replace.ts
```diff
@@ -293,7 +293,14 @@ export const SearchReplace = Extension.create<SearchOptions, SearchStorage>({
...
           );
-          scrollIntoView();
+
+          const htmlElement = this.editor.view.domAtPos(from).node;
+          if ("scrollIntoView" in htmlElement) {
+            htmlElement?.scrollIntoView({
+              behavior: "instant",
+              block: "center"
+            });
+          }
```
thecodrr: Why are you inlining this?

packages/editor/src/toolbar/floating-menus/search-replace.tsx
```diff
@@ -29,9 +29,14 @@ export function SearchReplaceFloatingMenu(props: FloatingMenuProps) {
...
   useLayoutEffect(() => {
-    const { searchTerm, ...options } = useEditorSearchStore.getState();
-    if (!options.isSearching) editor.commands.endSearch();
-    else editor.commands.search(searchTerm, options);
+    // defer search to ensure editor is fully initialized
+    const timeoutId = setTimeout(() => {
+      const { searchTerm, ...options } = useEditorSearchStore.getState();
+      if (!options.isSearching) editor.commands.endSearch();
+      else editor.commands.search(searchTerm, options);
+    }, 0);
```
thecodrr: Is the timeout really necessary?

01zulfi: If the search was active in another tab, and we open a new one, the search doesn't work as expected. We need to wait for the editor to be initialized.

Maybe instead of this timeout, I could look into the editor onCreate and initialize search there?

thecodrr: That sounds better as the timeout could fail to work in some edge cases.

PR #8065, opened 2025-05-07
editor: add scrollbar to color picker

01zulfi:

### DIFF
--- a/packages/editor/src/toolbar/popups/color-picker.tsx
+++ b/packages/editor/src/toolbar/popups/color-picker.tsx
@@ -171,13 +171,16 @@ export function ColorPicker(props: ColorPickerProps) {
         ) : null}
         <Flex>
           <Flex
-            className="hide-scrollbar"
             sx={{
               flex: 1,
               p: 1,
               overflowX: ["auto", "hidden"],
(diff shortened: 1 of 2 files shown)


01zulfi:
I added the `ScrollContainer` component but the styles of the scrollbar remain unchanged.

How it looks now (it uses the browser scrollbar style)

image

thecodrr:
> I added the ScrollContainer component but the styles of the scrollbar remain unchanged.

Maybe because you are applying it at the wrong level. It should work because it _is_ working everywhere else.

packages/editor/src/toolbar/popups/color-picker.tsx
```diff
@@ -171,13 +171,16 @@ export function ColorPicker(props: ColorPickerProps) {
         ) : null}
         <Flex>
           <Flex
-            className="hide-scrollbar"
```
thecodrr, line 174: We hide the scrollbar on mobile devices where the colors are horizontally scroll-able.

01zulfi:
I'm now using the `.scroll-bar` CSS class that was added in https://github.com/streetwriters/notesnook/pull/9085

PR #8070, opened 2025-05-08
web: fix navigation commands in command palette

01zulfi:

### DIFF
--- a/apps/web/src/dialogs/command-palette/commands.ts
+++ b/apps/web/src/dialogs/command-palette/commands.ts
@@ -72,8 +72,14 @@ export const commandActions = {
   command: (command: Command) => getCommandById(command.id)?.action(command),
   note: (command: Command, options?: { openInNewTab?: boolean }) =>
     useEditorStore.getState().openSession(command.id, options),
-  notebook: (command: Command) => navigate(`/notebooks/${command.id}`),
-  tag: (command: Command) => navigate(`/tags/${command.id}`),
+  notebook: (command: Command) => {
+    useAppStore.getState().setNavigationTab("notebooks");
+    navigate(`/notebooks/${command.id}`);
+  },
+  tag: (command: Command) => {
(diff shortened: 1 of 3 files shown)


apps/web/src/components/navigation-menu/tabs.ts
thecodrr, line 1: Why this change?

apps/web/src/dialogs/command-palette/commands.ts
```diff
@@ -72,8 +72,14 @@ export const commandActions = {
   command: (command: Command) => getCommandById(command.id)?.action(command),
   note: (command: Command, options?: { openInNewTab?: boolean }) =>
     useEditorStore.getState().openSession(command.id, options),
-  notebook: (command: Command) => navigate(`/notebooks/${command.id}`),
-  tag: (command: Command) => navigate(`/tags/${command.id}`),
+  notebook: (command: Command) => {
+    useAppStore.getState().setNavigationTab("notebooks");
+    navigate(`/notebooks/${command.id}`);
```
thecodrr, line 77: Instead of doing this in 2 steps everywhere, we could just listen to the `onNavigate` event (we already listen to it in `navigation-menu`) and change the tab based on the route. It would greatly simplify things.

thecodrr (changes requested):

apps/web/src/components/navigation-menu/tabs.ts
01zulfi: Don't think this change is necessary, reverted

thecodrr (changes requested):
Upon testing the new changes, it seems I missed one thing:

1. Now the tabs navigate even when you are opening a notebook via a shortcut. I think the `onNavigate` handler won't work here. The previous approach is better as its explicit whether we navigate or not.

01zulfi:
Oh nice catch, brought back to previous approach

Issue #7676, opened 2025-02-26
Focus note tile via Shift+Tab in new notes


codedust:
### What problem are you facing?

When creating a new note, it would be super handy to have some kind of shortcut to focus the note title.

### What's the solution you'd like to see in Notesnook?

When the note body is focused, it would be nice to switch to the note tile by pressing Shift+Tab:

![Image](https://github.com/user-attachments/assets/65b774f2-e436-437e-bad9-a784b0f7c3dc)

### What alternatives have you considered?

There could be another designated shortcut to focus the note title

Leviob:
Perhaps an alternative solution would be if the title was somehow part of the note. Like the first line was always treated as the title, and there is no title box at all. 
Or perhaps something similar to a markdown shortcut to enter the title from the note body.

PR #8130, opened 2025-05-22
web: add arrow up/down navigation b/w title and editor

01zulfi:

Related https://github.com/streetwriters/notesnook/issues/7676


thecodrr:
I am not sure arrow keys are good for this. What's wrong with `Shift + Tab` to move focus back to title? I think that should be supported by browsers by default based on `tabIndex`.

01zulfi:
Sounds good to me.

Moving from title to editor via `Tab` works as expected.

Though moving from editor to title via `Shift+Tab` doesn't work, I suspect that tiptap is stopping this keyboard event. We could handle `Shift+Tab` manually to move focus to title if the cursor is at the first position of the editor. Should I do this, @thecodrr?

thecodrr:
Closing this as it seems unnecessary.

PR #8141, opened 2025-05-27
web: add collapse sidebar button

### LINKED ISSUE #8094: It would be nice if there was a button to collapse the sidebar

Image

https://github.com/user-attachments/assets/f78bddb8-d1de-473d-92f8-f44471ccd7fe

could work this way

_No response_

### Additional context

_No response_

01zulfi:

Closes #8094 

https://github.com/user-attachments/assets/5eb2db77-8caf-4df8-8683-932874ff62c5

### DIFF
--- a/apps/web/src/app.tsx
+++ b/apps/web/src/app.tsx
@@ -207,7 +207,10 @@ function DesktopAppContents() {
                 zIndex: 3
               }}
             >
-              <NavigationMenu onExpand={() => navPane.current?.reset(0)} />
+              <NavigationMenu
+                onExpand={() => navPane.current?.reset(0)}
+                onCollapse={() => navPane.current?.collapse(0)}
+              />
             </Pane>
           )}
           {isFocusMode ? null : (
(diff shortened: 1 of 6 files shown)


thecodrr:
I don't think this is necessary. You can drag and collapse the sidebar. What's the point of a button?

Starship-Headlight:
This is a good feature if it could remember the custom width. I always find myself adjusting the sidebar to that perfect width where it's not completely collapsed but it's collapsed to the limit of names (like `Notes`, `monograph` etc) being visible. With this button I will not have to carefully adjust the width. I could just press the button and do that (given that this button remembers the past custom width). However, this button could be at the bottom. I think at the top it kind of crowds the place with the logo name and profile picture. Also, in the current implementation it just completely collapses the navigation menu which could be done by dragging so maybe it could be tweaked a little to give it more functionality.

marcosdiasalves:
+1 for this request; I do not understand why it was closed.

PR #8162, opened 2025-06-02
web: add support for notebook locking

01zulfi:


packages/core/src/common.ts
```diff
@@ -123,7 +123,9 @@ export const EVENTS = {
   mediaAttachmentDownloaded: "attachments:mediaDownloaded",
   vaultLocked: "vault:locked",
   vaultUnlocked: "vault:unlocked",
-  systemTimeInvalid: "system:invalidTime"
+  systemTimeInvalid: "system:invalidTime",
+  notebooksLocked: "notebook:locked",
+  notebookLockOpened: "notebook:lockOpened"
```
thecodrr, line 128: Shouldn't it be:

```suggestion
  notebookLocked: "notebook:locked",
  notebookUnlocked: "notebook:unlocked"
```

packages/core/src/collections/notebooks.ts
```diff
@@ -265,6 +292,59 @@ export class Notebooks implements ICollection {
       .get();
     return relation[0]?.fromId;
   }
+
+  async closeLockAll() {
+    EV.publish(EVENTS.notebooksLocked);
+    this.cache.lockOpenedNotebooks = [];
+  }
+
+  async openLock(id: Notebook["id"], password: string): Promise<boolean> {
```
thecodrr, line 301:
```suggestion
  async unlock(id: Notebook["id"], password: string): Promise<boolean> {
```

packages/core/src/collections/notebooks.ts
```diff
@@ -265,6 +292,59 @@ export class Notebooks implements ICollection {
       .get();
     return relation[0]?.fromId;
   }
+
+  async closeLockAll() {
```
thecodrr, line 296: The naming could be better.

packages/core/src/collections/notebooks.ts
```diff
@@ -37,6 +39,14 @@ export class Notebooks implements ICollection {
    * @internal
    */
   collection: SQLCollection<"notebooks", TrashOrItem<Notebook>>;
+  cache: {
+    lockOpenedNotebooks: Notebook["id"][];
+    lockedNotebooks: Notebook["id"][];
+  } = {
+    lockOpenedNotebooks: [],
```
thecodrr, line 46:
```suggestion
    unlockedNotebooks: [],
```

packages/core/src/types.ts
```diff
@@ -229,6 +229,7 @@ export interface Notebook extends BaseItem<"notebook"> {
   dateDeleted: null;
   itemType: null;
   deletedBy: null;
+  password?: Cipher<"base64"> | null;
```
thecodrr: Let's store the password in plaintext.

apps/web/src/components/note/index.tsx
```diff
@@ -403,6 +412,7 @@ export const noteMenuItems: (
       title: strings.archive(),
       isChecked: note.archived,
       icon: Archive.path,
+      isDisabled: context?.isInLockedNotebook,
```
thecodrr: Let's hide these instead of disabling them.

apps/web/src/views/notes.tsx
```diff
@@ -51,8 +59,53 @@ function Notes(props: NotesProps) {
...
   if (!context || !contextNotes) return <ListLoader />;
+
+  if (showUnlockNotebookPreview && context.type === "notebook") {
+    return (
+      <>
+        {header}
+        <UnlockNotebookPreview notebookId={context.id} />
+      </>
+    );
+  }
```
thecodrr, line 109: I think we should show a dialog instead. This doesn't look right.

thecodrr (changes requested):
A first-look review.

apps/web/src/views/notes.tsx
01zulfi: I think showing the password field directly in the list pane is better and less intrusive than a dialog. With a dialog, the user is forced to interact with it in case they navigate to a locked notebook, but with just the password field, user interaction is optional so they can continue to unlock it or navigate away

image

packages/core/src/common.ts
01zulfi: **The four notebook states are:**

1. Locked: a notebook which wasn't locked prior is locked
2. Unlocked: a notebook which is locked is not locked anymore i.e. the password requirement is removed
3. Lock opened: a notebook's lock is temporarily bypassed because the user provided the correct password
4. Lock closed: a notebook whose lock is temporarily bypassed is locked back

**Event names:**

1. `notebooksLocked`: this is accurate since the events for state 1 & 4 are handled identically and more than 1 notebook can have their "locks closed" at a time (from the status bar)
2. `notebookLockOpened`: this is to handle events for 2 & 3 which are identical. maybe we can rename the event to `notebookUnlockedOrLockOpened`

**Method names on the `notebooks` collection:**

1. `closeLockAll`: this handles state 4 which is to close all locks. I'm not sure what else we could name this, maybe `clearOpenLocks`?
2. `openLock`: this is to temporarily open the lock by user provided password
3. `lock`: for state 1
4. `unlock`: for state 2
5. `isLocked`: a utility to strictly check if the provided notebook is locked
6. `isLockOpen`: a utility to check if the notebook's lock is opened. the caveat here is that the caller must make sure that the provided notebook is locked

`isLocked` and `isLockOpen` is used together to in various places throughout the app

**Cache names:**

1. `lockedNotebooks`: self-explanatory
2. `lockeOpenedNotebooks`: to keep track of state 3

---

During working on the PR, I've gone through several names for many of the above. I find the current naming to be at least consistent throughout but there's still room for improvement. Before renaming them to your suggestions, I just wanted to provide some context. Do let me know what renames we can do after you've gone through this

apps/web/src/views/notes.tsx
thecodrr: While I agree that it's less intrusive, it looks very out of place. It fits in the case of a locked note but looks cramped and unrefined for a locked notebook. We have 2 choices:

- Improve the UI
- Move it to a dialog

I think if someone is clicking on a locked notebook (we already put an indicator that it's locked) then even if we show a dialog, it's okay.

01zulfi: The UX of a dialog needs more thought. Considering the following cases:

1. If a user navigates to a locked notebook accidently and they are shown a password dialog, they will just exit the dialog. What will we show in the list pane then? This is easily solved by just moving the user back to the page they navigated from.
2. If a user locks the active notebook, they enter the password -> the notebook is locked. But now they get shown another dialog to unlock the notebook even though it was not their intention. We cannot navigate the user away so what do we do when they cancel the dialog? Will we hide the list pane entirely? Will we just show the notebook header? Will we have an open lock button to show the dialog back?
3. The second case also goes for a locked notebook which is temporarily opened, But, then the lock was closed when the notebook in question was the active one.

Maybe we could track the user's behavior around the locked notebook to figure when and when not to show the dialog?

Though, for the locked notebook, the list pane is available anyways since we can't show notes. We could just show the password field there. We could work on improving the UI.

thecodrr: These are some very good points.

1. Yes, easily solved by aborting the navigation.
2. We won't show the dialog again. The notebook is locked but the user is inside so no point in making them enter the password again. This will be same if we go with inline UI.
3. We can go with a hybrid approach. An inline UI that shows an "unlock" button which opens the password dialog.

ammarahm-ed:
1. On locking Notebook on Web, if the main note list is open, it does not refresh to hide the notebook's notes.
2. For some reason closing the dropdown/context menus requires clicking twice
3. I think clicking on the locked notebook should not open the notebook, only show the unlock dialog
4. The wording in the "Open Notebook" dialog needs to be reviewed and updated. 
         a. Submit -> Open. 
         b. Notebook Password -> Password. 
         c. Description -> Enter the password for this notebook to open it.

victor-marino:
Hi!

Can I ask what's the hold up on this? I'm really interested in this feature, but I see it's been 9 months since the last comment.

Thanks!

PR #8195, opened 2025-06-11
clipper: improve clip with styles

01zulfi:


packages/clipper/src/index.ts
```diff
@@ -476,6 +477,29 @@ async function getPage(
...
+          resolveFetchOptions(config)
+        );
+        if (styleNode) {
+          head.appendChild(styleNode);
+        }
+      } else if (node instanceof HTMLStyleElement) {
+        head.appendChild(node.cloneNode(true));
+      }
+    }
+  }
```
01zulfi, line 501: @thecodrr the `getInlineNode` function (which calls `inlineStylesheets`) doesn't create/operate on the `head` at all. The `head` element is created here, so the inlining style logic is still here. Do let me know what would be the correct way to handle this

packages/clipper/src/domtoimage.ts
```diff
@@ -202,7 +186,13 @@ const VALID_ATTRIBUTES = [
...
   "height",
   "target",
-  "rel"
+  "rel",
+  "class",
+  "id",
+  "disabled",
+  "type",
+  "data-",
+  "aria-"
```
01zulfi: This might not be an exhaustive list of attributes that could be used in CSS. What is the significance of the valid attribute check?

thecodrr: I think this was added to trim down on useless attributes since we were already copying the styles to each element. Since that is no longer the case, we can remove this.

packages/clipper/src/index.ts
thecodrr: We can move this into a function in `styles.ts` and call it here.

thecodrr (changes requested):
Can we also remove the unused functions & code?

packages/clipper/src/domtoimage.ts
```diff
@@ -16,41 +16,31 @@ GNU General Public License for more details.
...
 
 // Default impl options
 const defaultOptions: Options = {
   inlineOptions: {}
 };
 
 async function getInlinedNode(node: HTMLElement, options: Options) {
-  const { fonts, images, stylesheets, inlineImages } =
-    options.inlineOptions || {};
+  const { fonts, stylesheets, inlineImages } = options.inlineOptions || {};
```
thecodrr: We can't ignore `images` since it is used to configure whether we want images in the clip or not. I suggst removing the `img` tags from the cloned node if `images` option is false.

thecodrr (changes requested):

thecodrr (changes requested):
1. We need to retain logic from https://github.com/streetwriters/notesnook/blob/master/packages/clipper/src/clone.ts#L131
2. `styles.ts` still has some unused code lying around. I think we can also get rid of `specificity`, `css-what` and `css-tokenizer`.
3. We need to add support for copying over attributes from the root element of a document `<html>` as that's sometimes used by websites for styling etc. (github).

packages/clipper/src/clone.ts
thecodrr, line 1: Yikes. No. The previous approach was okay (and faster as well since we were cloning the nodes in one go). We just need to transfer the filtering logic from `makeNodeCopy` function just like we filter out images if `options.images` is false.

thecodrr (changes requested):

packages/clipper/src/clone.ts
01zulfi: I was doing that initially, but I wasn't able to find appropriate dom methods for some of the nodes.

For example this:

```ts
if (original.nodeType === Node.COMMENT_NODE) return null;
```

To find the comment nodes, we have to manually traverse the entire dom anyway so I've opted to filter out images etc in this instead of just

```ts
const imgs = clone.querySelectorAll("img");
imgs.forEach(e=>e.remove());
```

Do let me know how to approach this

thecodrr: We can keep the comments. They don't really matter.

01zulfi: All of the filters except these two have been added

```ts
    if (original.nodeType === Node.COMMENT_NODE) return null;
```
```ts

    if (original.nodeType !== Node.TEXT_NODE && !isSVGElement(original)) {
      const { display, width, height } = window.getComputedStyle(original);
      if (display === "none" || (width === "0px" && height === "0px"))
        return null;

      if (isCustomElement(original)) {
        const isInline = display.includes("inline");
        const element = document.createElement(isInline ? "span" : "div");
        for (const attribute of original.attributes) {
          element.setAttribute(attribute.name, attribute.value);
        }
        return element;
      }
    }

```

I noticed we can perhaps use the `document.querySelectorAll("*")` selector to get all elements; this will allow us to add the above conditions as well. Is it worth it?

thecodrr: I don't think its necessary anymore. We are preserving the whole document including all the attributes and styles so there's no need for special handling for custom elements.

PR #8232, opened 2025-06-19
web: adjust tab width based on number of tabs

01zulfi:
https://github.com/user-attachments/assets/ceb38b45-b094-4a42-9e3e-3f5f5141255b

### DIFF
--- a/apps/web/src/components/editor/action-bar.tsx
+++ b/apps/web/src/components/editor/action-bar.tsx
@@ -18,7 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { Button, Flex, Text } from "@theme-ui/components";
-import React, { useEffect, useRef, useState } from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import {
   ArrowLeft,
   ArrowRight,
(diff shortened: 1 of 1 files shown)


thecodrr:
I am not sure what purpose this will serve. In browsers, you can recognize a tab by its favicon even if its title is not visible but that is not the case here. Making the tab item too small hides the title making it harder to quickly jump between tabs and forcing the user to guess.

01zulfi:
I think that's a valid point. For browsers, the favicon is major part of the tabs experience. Though I do feel adaptive tab widths is also important from a UX perspective. All tab implementations support this out of the box, except for I think VSCode, and code editors.

Maybe, if we could support custom note icons or emojis, it could help. Though that might be a challenge as well since the note icons are stateful: they show us whether the note is not saved, readonly etc

image

thecodrr:
> All tab implementations support this out of the box, except for I think VSCode, and code editors.

That's exactly why they don't support it either. It's impossible to identify a tab if its smaller than a certain width.

> Maybe, if we could support custom note icons or emojis, it could help. Though that might be a challenge as well since the note icons are stateful: they show us whether the note is not saved, readonly etc

Yes but that will be up to the user to set a custom emoji which they might never do. And an emoji and favicons have nothing in common. A favicon is unique and easily identifiable - how will you do that with emojis? Emojis are meant to be common.

01zulfi:
Yeah that sounds good to me. 

Here's what it would look like if we set initial width to `150` and reduce it to `100` gradually. We can still see characters of the note title.


https://github.com/user-attachments/assets/8daa5873-3d19-4725-8fff-5f4e057f4347


Though this is different from the current tab width, where it starts from the note title length and goes up to `120`.

---

Anyways, closing this PR, do let me know if you thing the above change is okay and I can reopen.

PR #8236, opened 2025-06-20
Play Audio Files

Hamster45105:
I've added support for playing audio files directly in the editor: 

image


packages/editor/src/extensions/attachment/component.tsx
```diff
@@ -18,19 +18,155 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+            try {
+              const byteCharacters = atob(data.split(",")[1] || data);
+              const byteNumbers = new Array(byteCharacters.length);
+              for (let i = 0; i < byteCharacters.length; i++) {
+                byteNumbers[i] = byteCharacters.charCodeAt(i);
+              }
+              const byteArray = new Uint8Array(byteNumbers);
+              const blob = new Blob([byteArray], { type: mime });
+              const url = URL.createObjectURL(blob);
+              setAudioSrc(url);
```
01zulfi: There's a `toBlobUrl` function in `src/utils/downloader.ts` which can be reused with slight modifications

packages/editor/src/extensions/attachment/component.tsx
```diff
@@ -18,19 +18,155 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+        })
+        .catch(console.error);
+    }
+  }, [isAudioFile, editor.storage, hash, mime]);
+
+  // Clean up blob URL on unmount
+  useEffect(() => {
+    return () => {
+      if (audioSrc) {
+        URL.revokeObjectURL(audioSrc);
```
01zulfi: Similarly here, there's a `revokeBloburl` util function which can be reused

packages/editor/src/extensions/attachment/component.tsx
```diff
@@ -18,19 +18,155 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+          display: "inline-flex",
+          flexDirection: "column",
+          position: "relative",
+          userSelect: "none",
+          backgroundColor: "var(--background-secondary)",
+          p: 2,
+          m: 1,
+          borderRadius: "default",
+          border: "1px solid var(--border)",
+          maxWidth: 350,
```
01zulfi: For me, the progress bar has very less width and is unusable. We could increase the width, maybe make it half the editor width or something. What do you think?

packages/editor/src/extensions/attachment/component.tsx
```diff
@@ -18,19 +18,155 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+  // Clean up blob URL on unmount
+  useEffect(() => {
+    return () => {
+      if (audioSrc) {
+        URL.revokeObjectURL(audioSrc);
+      }
+    };
+  }, [audioSrc]);
+
+  if (isAudioFile && audioSrc) {
```
01zulfi: I feel there's a lot of JSX duplication. Could you try to see if we can reduce this duplication?

01zulfi (changes requested):
@Hamster45105 , this looks good 🔥 

Left some comments below

packages/editor/src/extensions/attachment/component.tsx
Hamster45105: Yes good idea, I was only testing with short audio files so I didn't really consider this issue.

packages/editor/src/extensions/attachment/component.tsx
Hamster45105: I've tried to minimise duplication.

packages/editor/src/extensions/attachment/component.tsx
```diff
@@ -18,57 +18,92 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+) {
   const { editor, node, selected } = props;
-  const { filename, size, progress } = node.attrs;
+  const { filename, size, progress, mime, hash } = node.attrs;
   const elementRef = useRef<HTMLSpanElement>();
   const [isDragging, setIsDragging] = useState(false);
+  const [audioSrc, setAudioSrc] = useState<string>();
 
-  return (
+  // Check if this is an audio file
```
01zulfi: We can remove all of the comments, they aren't adding anything meaningful

packages/editor/src/extensions/attachment/component.tsx
```diff
@@ -84,30 +119,85 @@ export function AttachmentComponent(props: ReactNodeViewProps<FileAttachment>) {
...
+            })
+      }}
+      title={!isAudioFile || !audioSrc ? filename : undefined}
+      onDragStart={() => setIsDragging(true)}
+      onDragEnd={() => setIsDragging(false)}
+      data-drag-handle
+    >
+      {renderFileInfo()}
+      {isAudioFile && audioSrc && renderAudioPlayer()}
+      {renderToolbar()}
```
01zulfi: Don't think we need to separate out into functions when we can write JSX directly here

packages/editor/src/extensions/attachment/component.tsx
thecodrr, line 1: I feel like this should be its own extension like web clip or image etc. We could even make it a block level node instead of inline like it is now.

thecodrr (changes requested):

packages/editor/src/extensions/attachment/component.tsx
Hamster45105: Hi,

I don't have the time to work on this at the moment, but I will be sure to get around to it as soon as I can.

Hamster45105: @thecodrr I've made the requested changes.

packages/editor/src/extensions/audio/component.tsx
```diff
@@ -0,0 +1,158 @@
...
+                : ["downloadAttachment"]
+            }
+            sx={{
+              boxShadow: "menu",
+              borderRadius: "default",
+              backgroundColor: "var(--background)",
+              position: "absolute",
+              top: -35
+            }}
+          />
```
01zulfi, line 154: nit: instead of top left, can we have the toolbar at the top right corner (similar to the image toolbar)

Image

01zulfi (approved):
everything looks good to me 🔥 

just left a small comment below

packages/editor/src/extensions/audio/component.tsx
Hamster45105: Done 👍

01zulfi:
Hey @Hamster45105, we need to merge this into the `beta` branch. For that, we need to resolve the merge conflicts. Are you available to do that? If not, that's alright, I can wrap this up

Hamster45105:
I've resolved the conflict!

01zulfi (approved):
lgtm 🚀

packages/editor/src/utils/format-bytes.ts
```diff
@@ -0,0 +1,30 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+export function formatBytes(bytes: number, decimals = 1) {
```
thecodrr, line 20: We already have `formatBytes` in `@notesnook/common`. Use that.

apps/web/src/components/editor/picker.ts
```diff
@@ -174,6 +180,24 @@ async function pickImage(
   }
 }
 
+async function pickAudio(
```
thecodrr, line 183: This is unnecessary. Just set the `type` based on `file.type` in `pickFile`.

packages/editor/src/extensions/audio/audio.ts
```diff
@@ -0,0 +1,118 @@
...
+          return {
+            type: "audio",
+            hash: element.dataset.hash,
+            filename: element.dataset.filename,
+            mime: element.dataset.mime,
+            size: element.dataset.size ? parseInt(element.dataset.size) : 0
+          };
+        }
+      }
+    ];
```
thecodrr, line 89: The `tag` should be `audio` instead of a `div` or `span`.

packages/editor/src/extensions/audio/audio.ts
```diff
@@ -0,0 +1,118 @@
...
+  },
+
+  renderHTML({ HTMLAttributes }) {
+    return [
+      "div",
```
thecodrr, line 94:
```suggestion
      "audio",
```

thecodrr (changes requested):

thecodrr:
@Hamster45105 Also, maybe the audio player should be a block (i.e. on its own line like an image) instead of an inline element.

Hamster45105:
Hey, thanks for the review!

Unfortunately, I don't think I'm going to have the time to do this any time soon, but I'll be sure to come back to it when I can.

01zulfi:
since I was unable to push commits into this PR, closing it in favor of https://github.com/streetwriters/notesnook/pull/9364

PR #8263, opened 2025-06-27
web: fix note create option in quick open dialog

01zulfi:
* open new tab when query empty
* use editor-store methods for note creation


thecodrr (changes requested):
There's a bug here:

- Change default title format to `$headline$
- Create a note using the command palette
- Type something in the content area
- Notice note title is updating even though we already set it.

01zulfi:
Looks like that's a bug with `$headline$` itself

thecodrr:
How so?

01zulfi:
It is reproducible with normal note creation flow as well.

Create note in a new tab by setting a title -> add some content -> the title gets replaced

thecodrr:
> Create note in a new tab by setting a title -> add some content -> the title gets replaced

~Can't reproduce it.~ Yep, its a bug.

PR #8279, opened 2025-06-30
common: include archive in export

01zulfi:


thecodrr (changes requested):
Let's also include `archived` state in the md/html frontmatter.

PR #8284, opened 2025-06-30
web: add editor line height setting

01zulfi:

### DIFF
--- a/apps/web/src/components/editor/common.ts
+++ b/apps/web/src/components/editor/common.ts
@@ -23,3 +23,9 @@ export const EDITOR_ZOOM = {
   MIN: 30,
   STEP: 10
 };
+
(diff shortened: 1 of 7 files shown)


apps/web/src/components/editor/common.ts
```diff
@@ -23,3 +23,9 @@ export const EDITOR_ZOOM = {
   MIN: 30,
   STEP: 10
 };
+
+export const EDITOR_LINE_HEIGHT = {
+  DEFAULT: 1.2,
```
01zulfi, line 28: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#normal

thecodrr:
@ammarahm-ed you'll need to add this on your end as well.

ammarahm-ed:
@thecodrr mobile side done.

PR #8290, opened 2025-07-02
web: fix scroll to search result

01zulfi:

### DIFF
--- a/apps/web/src/components/search-result/index.tsx
+++ b/apps/web/src/components/search-result/index.tsx
@@ -215,11 +215,3 @@ async function menuItems(
     color: colors[0]
   });
 }
-
-function findSelectedMatchIndex(item: HighlightedResult, matchIndex: number) {
-  let activeIndex = 0;
-  for (let i = 0; i <= matchIndex - 1; ++i) {
-    activeIndex += item.content[i].length;
-  }
-  return activeIndex;
-}
--- a/apps/web/src/components/search-result/index.tsx
+++ b/apps/web/src/components/search-result/index.tsx
@@ -19,7 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import { HighlightedResult } from "@notesnook/core";
 import { Button, Flex, Text } from "@theme-ui/components";
-import React, { useState } from "react";
+import React, { Fragment, useState } from "react";
 import { useEditorStore } from "../../stores/editor-store";
 import { useStore as useNoteStore } from "../../stores/note-store";
 import ListItem from "../list-item";
@@ -112,11 +112,11 @@ function SearchResult(props: SearchResultProps) {
                 }}
               >
                 {item.title.map((match) => (
-                  <>
+                  <Fragment key={match.id}>
                     <span>{match.prefix}</span>
                     <span className="match">{match.match}</span>
                     {match.suffix ? <span>{match.suffix}</span> : null}
-                  </>
+                  </Fragment>
                 ))}
               </Text>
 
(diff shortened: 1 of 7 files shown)


apps/web/src/components/search-result/index.tsx
```diff
@@ -146,14 +146,14 @@ function SearchResult(props: SearchResultProps) {
                 useEditorStore.getState().openSession(item.id, {
                   rawContent: item.rawContent,
                   force: true,
-                  activeSearchResultIndex: findSelectedMatchIndex(item, index)
+                  activeSearchResultId: match[0].id
```
01zulfi, line 149: if there are multiple matches in the same search item, we only scroll to the first match, I don't think this should be a problem

apps/web/src/components/search-result/index.tsx
```diff
@@ -146,14 +146,14 @@ function SearchResult(props: SearchResultProps) {
...
                 });
               }}
               onMiddleClick={() => {
                 useEditorStore.getState().openSession(item.id, {
                   openInNewTab: true,
                   rawContent: item.rawContent,
-                  activeSearchResultIndex: findSelectedMatchIndex(item, index)
```
thecodrr, line 156: I think we can get rid of `findSelectedMatchIndex`.

packages/core/src/api/lookup.ts
```diff
@@ -62,10 +62,8 @@ type FuzzySearchField<T> = {
 };
 
 const MATCH_TAG_NAME = "nn-search-result";
-const MATCH_TAG_OPEN = `<${MATCH_TAG_NAME}>`;
-const MATCH_TAG_CLOSE = `</${MATCH_TAG_NAME}>`;
 const MATCH_TAG_REGEX = new RegExp(
-  `<${MATCH_TAG_NAME}>(.*?)<\\/${MATCH_TAG_NAME}>`,
+  `<${MATCH_TAG_NAME}(?:\\s+id="([^"]*)")?>(.*?)<\\/${MATCH_TAG_NAME}>`,
```
thecodrr: I think we can simplify the regex a bit: `<${MATCH_TAG_NAME}\s+id="(.+?)">`

packages/core/src/api/lookup.ts
```diff
@@ -796,10 +795,11 @@ export function splitHighlightedMatch(text: string): Match[][] {
...
   let totalLength = 0;
 
-  for (let i = 0; i < parts.length - 1; i += 2) {
+  for (let i = 0; i < parts.length - 1; i += 3) {
     const prefix = parts[i];
-    const match = parts[i + 1];
-    let suffix = parts.at(i + 2);
+    const matchId = parts[i + 1];
+    const match = parts[i + 2];
+    let suffix = i + 3 < parts.length ? parts[i + 3] : "";
```
thecodrr: There's no need for this since all match tags will have an id.

packages/core/src/api/lookup.ts
```diff
@@ -815,14 +815,17 @@ export function splitHighlightedMatch(text: string): Match[][] {
         suffix,
         Math.max(suffix.length / 2, 60)
       );
-      parts[i + 2] = remaining;
+      if (i + 3 < parts.length) {
```
thecodrr: There's no need for this since all match tags will have an id.

packages/core/src/api/lookup.ts
```diff
@@ -778,10 +776,11 @@ function highlightQueries(
     const regex = new RegExp(patterns.join("|"), "gi");
 
     let hasMatches = false;
+    let matchIdCounter = 0;
 
     const result = text.replace(regex, (match) => {
       hasMatches = true;
-      return `${MATCH_TAG_OPEN}${match}${MATCH_TAG_CLOSE}`;
+      return `<${MATCH_TAG_NAME} id="match-${++matchIdCounter}">${match}</${MATCH_TAG_NAME}>`;
```
thecodrr: Let's turn this into a function and reuse it.

packages/core/src/api/lookup.ts
```diff
@@ -981,9 +984,8 @@ function highlightHtmlContent(html: string, queries: string[]): string {
         // Reset regex state after test
         searchRegex.lastIndex = 0;
 
-        const processed = text.replace(
-          searchRegex,
-          "<nn-search-result>$1</nn-search-result>"
+        const processed = text.replace(searchRegex, (match) =>
+          createSearchResultTag(match, `match-${++matchIdCounter}`)
```
01zulfi, line 988: currently I'm generating match id manually via a counter, should we do this via `getId` instead? Is one way better from the other? I'm curious what do you think

PR #8311, opened 2025-07-07
web: add character, spaces, paragraphs count in status bar popup

01zulfi:

### DIFF
--- a/apps/web/src/components/editor/footer.tsx
+++ b/apps/web/src/components/editor/footer.tsx
@@ -28,23 +28,23 @@ import {
   Plus,
   Minus,
   EditorNormalWidth,
-  TableOfContents,
   ExitFullscreen,
   Fullscreen,
   EditorFullWidth,
-  NormalMode
+  NormalMode,
+  Cross
 } from "../icons";
-import {
-  useEditorConfig,
-  useNoteStatistics,
-  useEditorManager
-} from "./manager";
+import { useEditorConfig, useNoteStatistics } from "./manager";
 import { getFormattedDate } from "@notesnook/common";
-import { MAX_AUTO_SAVEABLE_WORDS } from "./types";
+import { MAX_AUTO_SAVEABLE_WORDS, NoteStatistics } from "./types";
 import { strings } from "@notesnook/intl";
 import { EDITOR_ZOOM } from "./common";
 import { useWindowControls } from "../../hooks/use-window-controls";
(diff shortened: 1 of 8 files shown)


apps/web/src/components/editor/tiptap.tsx
```diff
@@ -274,22 +319,64 @@ function TipTap(props: TipTapProps) {
...
+            editor as Editor,
+            transaction.selection
+          );
+          const selectedSpaces = getSelectedSpaces(
+            editor as Editor,
+            transaction.selection
+          );
+          const selectedCharacters = getSelectedCharacters(
+            editor as Editor,
+            transaction.selection
```
thecodrr: Instead of passing `transaction.selection` in each of them and then doing `.textBetween`, let's do `.textBetween` once and directly pass the text. That'll be more efficient. `getSelectedParagraphs` is a special case so it can be handled separately.

01zulfi: I also noticed we don't need special handling for `selection.empty` since it is already done above. Removed `getSelected*` functions altogether except for `getSelectedParagraphs`

apps/web/src/components/editor/tiptap.tsx
```diff
@@ -274,22 +319,68 @@ function TipTap(props: TipTapProps) {
...
             editor as Editor,
             transaction.selection
           );
+          const selectedCharacters = countCharacters(
+            editor.state.doc.textBetween(
+              transaction.selection.from,
+              transaction.selection.to,
+              "",
+              ""
+            )
```
thecodrr: I think it'll be faster to remove `\n` from `selectedText` then doing `textBetween` again.

apps/web/src/components/editor/footer.tsx
```diff
@@ -225,3 +235,108 @@ export default EditorFooter;
...
+  const root = document.getElementById("dialogContainer");
+
+  if (root) {
+    return new Promise((resolve) => {
+      const perform = (result: boolean) => {
+        ReactDOM.unmountComponentAtNode(root);
+        closeNoteStatisticsView();
+        resolve(result);
+      };
+      ReactDOM.render(
```
thecodrr: Can you check if we can use `PopupPresenter` from `@notesnook/ui` instead? That'll simplify things a lot since it handles positioning etc. automatically.

thecodrr (approved):

PR #8318, opened 2025-07-08
web: don't close note after deleting its attachment

### LINKED ISSUE #8298: After permanently deleting an individual note attachment, note context is lost.  Need to re-navigate back
### What happened?

After permanently deleting a note file attachment from the individual note properties, the individual note context is lost, and needs to be reopened from the note list.

Very inconvenient if you have multiple attachments to delete.

Expected Outcome:
- Should not lose note context on deletion of individual attachment

1. Upload a file attachment
2. Permanently delete file attachment using note properties

Video attached

### Version

3.2.3

### Platform/OS

macOS

### Relevant log output

```shell

```

01zulfi:
Closes #8298

### DIFF
--- a/apps/web/src/stores/attachment-store.ts
+++ b/apps/web/src/stores/attachment-store.ts
@@ -29,6 +29,7 @@ import { createWriteStream } from "../utils/stream-saver";
 import { Attachment } from "@notesnook/core";
 import { TaskManager } from "../common/task-manager";
 import { strings } from "@notesnook/intl";
+import { useEditorManager } from "../components/editor/manager";
 
 let abortController: AbortController | undefined = undefined;
 class AttachmentStore extends BaseStore<AttachmentStore> {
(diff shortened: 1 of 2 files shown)


apps/web/src/stores/attachment-store.ts
```diff
@@ -135,7 +136,33 @@ class AttachmentStore extends BaseStore<AttachmentStore> {
...
+          if (
+            !("note" in session) ||
+            !session.note.id ||
+            !session.note.contentId
+          ) {
+            continue;
+          }
+          if (!linkedNotes.includes(session.note.id)) {
+            continue;
+          }
```
01zulfi, line 150: I've added the session's content update logic here as opposed to the `unlink` event handling in `editor-store`.

This is because the `event` object doesn't have reference to note ids, so it is not possible to include this check in `editor-store`

```ts
if (!linkedNotes.includes(session.note.id)) {
    continue;
}
```

apps/web/src/stores/attachment-store.ts
```diff
@@ -135,7 +136,33 @@ class AttachmentStore extends BaseStore<AttachmentStore> {
...
+
+          useEditorManager
+            .getState()
+            .getEditor(session.id)
+            ?.editor?.updateContent(content.data);
+          useEditorStore.getState().updateSession(session.id, undefined, {
+            content
+          });
```
thecodrr: This is not a good idea. The note could be anything: `locked`, `readonly` etc. Just call `useEditorStore.getState().openSession` with `force: true` and let it handle everything.

01zulfi:
```ts
useEditorStore
  .getState()
  .openSession(session.note.id, { force: true })
 ```
 
 this doesn't update the note content in realtime. I still have to reopen the note to see that the attachment is removed

Comment (thecodrr) [reply apps/web/src/stores/attachment-store.ts]: > this doesn't update the note content in realtime. I still have to reopen the note to see that the attachment is removed

It should. That might be a bug.

Comment (01zulfi) [reply apps/web/src/stores/attachment-store.ts]: It wasn't a bug in the editor-store, rather the editor wasn't being rerendered

File: apps/web/src/components/editor/index.tsx [hunk 3]
```diff
@@ -237,13 +237,24 @@ export default function TabsView() {
...
+
+  if (
+    "attachmentsLength" in prev.session &&
+    "attachmentsLength" in next.session
+  ) {
+    return (
+      baseConditions &&
+      prev.session.attachmentsLength === next.session.attachmentsLength
+    );
+  }
```
thecodrr, line 255: Why is this in an `if` condition?

thecodrr (changes requested):

apps/web/src/components/editor/index.tsx
01zulfi: `attachmentsLength` is not available in every session `type`

I'm using an `if` condition to avoid optional chaining `prev.session?.attachmentsLength === next.session?.attachmentsLength`

PR #8347, opened 2025-07-14
web: fix github link in importer setting

01zulfi:


thecodrr:
The link is correct as is.

PR #8364, opened 2025-07-16
editor: fix internal links opening in new tab in readonly notes

01zulfi:


packages/editor/src/extensions/link/helpers/clickHandler.ts
```diff
@@ -30,6 +30,12 @@ export function clickHandler(options: ClickHandlerOptions): Plugin {
   return new Plugin({
     key: new PluginKey("handleClickLink"),
     props: {
+      handleDOMEvents: {
+        click: (_, event) => {
+          event.preventDefault();
+          return true;
+        }
```
thecodrr: Let's move all the logic from `handleClick` here. It's confusing to have 2 handlers for the same event.

thecodrr (approved):

PR #8365, opened 2025-07-16
web: fix nav pane on small screens

01zulfi:


thecodrr:
This will need some testing.

PR #8371, opened 2025-07-18
editor: add copy image toolbar button

01zulfi:


thecodrr:
Doesn't the browser and desktop app already have a copy image function on right click?

01zulfi:
That's correct. The right click option works fine. I was thinking maybe we could provide alternatives. Instead of a toolbar button, we could support `ctrl+c` keyboard shortcut for copying images?

thecodrr:
`ctrl+c` is a good idea.

01zulfi:
https://github.com/streetwriters/notesnook/pull/8386

Issue #8180, opened 2025-06-08
If the note title is in URL format (aaa.bbb), an http connection is requested when inserting a note link.


kos0114:
### What happened?

For example, if I create a note called 'extract.py', right-click on the note and select 'Copy link', and then paste it into another note, it will link to 'http://extract.py' instead of linking to the note.

### Steps to reproduce the problem

For example, if I create a note called 'extract.py', right-click on the note and select 'Copy link', and then paste it into another note, it will link to 'http://extract.py' instead of linking to the note.

### Version

3.1.1-8f97e54-desktop

### Platform/OS

Windows

### Relevant log output

```shell

```

Scott-will:
I looked at this one, same happens in notes app on iOS and in other text editors. It can be solved with custom code to parse it for nn://, is this something you want fixed @thecodrr ?

01zulfi:
Yes, this needs to be fixed. Are you willing to work on it, @Scott-will?

Scott-will:
Yep!

PR #8385, opened 2025-07-22
Handle pasting urls in html

Scott-will:
Fixes https://github.com/streetwriters/notesnook/issues/8180
Not deeply familiar with all the url formats, let me know if something that needs to be tested.
Tested html urls, markdowns, and something like http://example.com

cc @01zulfi

### DIFF
--- a/packages/editor/src/extensions/link/link.ts
+++ b/packages/editor/src/extensions/link/link.ts
@@ -232,9 +232,10 @@ export const Link = Mark.create<LinkOptions>({
         }
       }),
       markPasteRule({
-        find: (text) => {
+        find: (text, ev) => {
           const foundLinks: PasteRuleMatch[] = [];
-
+          const html = ev?.clipboardData?.getData("text/html");
+          if (html && html.includes("<a")) return [];
           if (text) {
             const links = find(text).filter((item) => item.isLink);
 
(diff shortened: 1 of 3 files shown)


thecodrr:
I am not sure if this fix requires all these changes. A few observations:

- `Copy link` on a note with a title that does not look like a link works
- Copy pasting HTML or MD content with links also works
- Somehow the editor is ignoring the `a` tag in the HTML with the `text` variant of the pasted content. Is there a way to stop this from happening?

packages/editor/src/extensions/link/link.ts
```diff
@@ -230,35 +231,6 @@ export const Link = Mark.create<LinkOptions>({
             href: regExp.exec(match[0])?.[1]
           };
         }
-      }),
-      markPasteRule({
-        find: (text) => {
-          const foundLinks: PasteRuleMatch[] = [];
```
thecodrr: What if we do just this instead:

```
        find: (text, ev) => {
          const html = ev?.clipboardData?.getData("text/html");
          if (html && html.includes("<a")) return [];
```

thecodrr (changes requested):

packages/editor/src/extensions/link/link.ts
Scott-will: Seems to work, can you share what you tried to paste when you say "Somehow the editor is ignoring the a tag in the HTML with the text variant of the pasted content. Is there a way to stop this from happening?". Il double check it covers this case as well

thecodrr: > can you share what you tried to paste 

I was talking about the `Copy link` example given in the original issue.

Scott-will: Updated, and tested and saw it working on desktop and web. Sorry it took a while, was camping

01zulfi (approved):
lgtm 👍

thecodrr (approved):

PR #8386, opened 2025-07-22
editor: support ctrl+c for images

### LINKED ISSUE #7352: Copy image from notesnook to another program
It is not possible to copy an image from notesnook to another program e.g., word, anki, etc.

Every time, when I right click on an image and click "copy image" and past it into another program, there is an empty image / just a white rectangle without anything in it.

**Device information:**
App version: 3.0.23-123705f-desktop
OS: Windows 10 64-bit
Browser: Electron 31.7.4
Pro: true

01zulfi:
Closes #7352


thecodrr:
@01zulfi can we add a test for this to make sure Ctrl+C keeps working for images?

PR #8395, opened 2025-07-25
editor: show image dimensions during drag resizing

01zulfi:
https://github.com/user-attachments/assets/a514d46e-0d4c-40a8-bec7-c33db4074aea

### DIFF
--- a/packages/editor/src/extensions/image/component.tsx
+++ b/packages/editor/src/extensions/image/component.tsx
@@ -51,6 +51,10 @@ export function ImageComponent(
   const [bloburl, setBloburl] = useState<string | undefined>(
     toBlobURL("", "image", mime, hash)
   );
+  const [resizing, setResizing] = useState<{
+    width: number;
+    height: number;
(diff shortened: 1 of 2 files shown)


NeadReport:
Yes! That is perfect. Thanks for working on this so quickly.

packages/editor/src/extensions/image/component.tsx
```diff
@@ -116,13 +123,47 @@ export function ImageComponent(
...
+              borderRadius: "default"
+            }}
+          >
+            <Text variant="subBody" sx={{ fontWeight: "bold" }}>
+              {resizing?.width}
+              {" × "}
+              {resizing?.height}
+            </Text>
+          </Box>
+        </ResponsivePresenter>
```
thecodrr: Why do we need `ResponsivePresenter` instead of simply position the `Box` where we want?

thecodrr (changes requested):

PR #8415, opened 2025-08-01
core: (wip) add support for monographs sync

01zulfi:

### DIFF
--- a/packages/core/src/api/monographs.ts
+++ b/packages/core/src/api/monographs.ts
@@ -20,50 +20,38 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import http from "../utils/http.js";
 import Constants from "../utils/constants.js";
 import Database from "./index.js";
-import { Note, isDeleted } from "../types.js";
+import { Monograph, Note, isDeleted } from "../types.js";
 import { Cipher } from "@notesnook/crypto";
 import { isFalse } from "../database/index.js";
-import { logger } from "../logger.js";
 
-type BaseMonograph = {
-  id: string;
-  title: string;
-  userId: string;
-  selfDestruct: boolean;
-};
-type UnencryptedMonograph = BaseMonograph & {
+type MonographApiRequestBase = Omit<
+  Monograph,
+  "type" | "dateModified" | "dateCreated" | "datePublished"
+>;
+type UnencryptedMonograph = MonographApiRequestBase & {
   content: string;
 };
-type EncryptedMonograph = BaseMonograph & {
+type EncryptedMonograph = MonographApiRequestBase & {
   encryptedContent: Cipher<"base64">;
 };
-type Monograph = UnencryptedMonograph | EncryptedMonograph;
+type MonographApiRequest = (UnencryptedMonograph | EncryptedMonograph) & {
+  userId: string;
+};
 
 export type PublishOptions = { password?: string; selfDestruct?: boolean };
 export class Monographs {
+  private readonly passwordEncryptionKey = "monographpasswordencryptionkey";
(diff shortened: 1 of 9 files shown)


packages/core/src/api/monographs.ts
```diff
@@ -20,50 +20,38 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
   encryptedContent: Cipher<"base64">;
 };
-type Monograph = UnencryptedMonograph | EncryptedMonograph;
+type MonographApiRequest = (UnencryptedMonograph | EncryptedMonograph) & {
+  userId: string;
+};
 
 export type PublishOptions = { password?: string; selfDestruct?: boolean };
 export class Monographs {
+  private readonly passwordEncryptionKey = "monographpasswordencryptionkey";
```
thecodrr: We can't use a hardcoded key here for obvious reasons. Let's encrypt the monograph password with user's key similar to how we encrypt items before they get synced.

packages/core/src/api/monographs.ts
```diff
@@ -20,50 +20,38 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+type EncryptedMonograph = MonographApiRequestBase & {
   encryptedContent: Cipher<"base64">;
 };
-type Monograph = UnencryptedMonograph | EncryptedMonograph;
+type MonographApiRequest = (UnencryptedMonograph | EncryptedMonograph) & {
+  userId: string;
+};
 
 export type PublishOptions = { password?: string; selfDestruct?: boolean };
 export class Monographs {
```
01zulfi, line 42: I'm keeping the `Monographs` api class intact and just using the addtional `monographsCollection` class internal to this. So the apps can continue interacting with this api class. 

We could perhaps merge the api and collection class, but I feel like separate is fine as well, what do you think?

thecodrr: It's a good approach. We can merge it later if necessary.

thecodrr (changes requested):

packages/core/src/api/monographs.ts
```diff
@@ -109,13 +96,17 @@ export class Monographs {
...
+    const monograph: MonographApiRequest = {
       id: noteId,
       title: note.title,
       userId: user.id,
       selfDestruct: opts.selfDestruct || false,
       ...(opts.password
         ? {
+            password: userKey
+              ? await this.db.storage().encrypt(userKey, opts.password)
+              : undefined,
```
thecodrr: Now that I think about it, this is also a bad idea. We'll have to re-encrypt and re-publish all these passwords whenever user's password changes.

What if instead of using user's encryption key, we create a new sub-master key just like we do in the case of attachments, and then use that to encrypt the password?

packages/core/src/api/user-manager.ts
```diff
@@ -459,6 +460,51 @@ class UserManager {
...
+        .decrypt(userEncryptionKey, user.monographPasswordsKey);
+      if (!plainData) return;
+      this.cachedMonographPasswordsKey = JSON.parse(plainData) as SerializedKey;
+      return this.cachedMonographPasswordsKey;
+    } catch (e) {
+      logger.error(e, "Could not get attachments encryption key.");
+      if (e instanceof Error)
+        throw new Error(
+          `Could not get attachments encryption key. Error: ${e.message}`
+        );
```
thecodrr:
```suggestion
      logger.error(e, "Could not get monographs encryption key.");
      if (e instanceof Error)
        throw new Error(
          `Could not get monographs encryption key. Error: ${e.message}`
        );
```

PR #8457, opened 2025-08-15
editor: improve checklist styling

01zulfi:
# Before

image

# After

image


thecodrr:
I think we should preserve the previous behaviour which added strikethrough on checking an item in task list.

thecodrr (changes requested):
What's with the formatting changes?

01zulfi:
I think I accidently used the default vscode formatter, fixed

thecodrr (approved):

PR #8466, opened 2025-08-18
web: increase list pane max size to 1000 for large screens

01zulfi:


apps/web/src/app.tsx
```diff
@@ -224,7 +226,7 @@ function DesktopAppContents() {
               initialSize={380}
               style={{ flex: 1, display: "flex" }}
               snapSize={120}
-              maxSize={500}
+              maxSize={isLargeScreen ? 1000 : 500}
```
thecodrr: Just set it to 1000 by default.

thecodrr (approved):

PR #8489, opened 2025-08-20
editor: improve table styling

01zulfi:
# before

image

# after

image

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -79,6 +79,7 @@
     "react-colorful": "^5.6.1",
     "redent": "^4.0.0",
     "refractor": "^4.8.1",
+    "simplebar-react": "^3.3.2",
     "strip-indent": "^4.0.0",
     "unfurl.js": "^6.4.0"
   },
--- a/packages/editor/package-lock.json
+++ b/packages/editor/package-lock.json
@@ -59,6 +59,7 @@
         "react-colorful": "^5.6.1",
         "redent": "^4.0.0",
         "refractor": "^4.8.1",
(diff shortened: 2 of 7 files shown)


thecodrr:
This is really awesome!

thecodrr:
@01zulfi can we also create full width tables by default like here: https://tiptap.dev/docs/editor/extensions/nodes/table

01zulfi:
Thanks for bringing that issue up. These are the two issues I'm facing

1. The table scrollbar doesn't show on page reload. But it starts showing after I click anywhere outside the editor
2. The column resize experience is very choppy when the table width overflows

Both of these issues are not present if we don't use the `ScrollContainer`. I've tried several ways to make it work, but not sure what's causing it or how to fix it; will investigate further...

thecodrr:
> 1. The table scrollbar doesn't show on page reload. But it starts showing after I click anywhere outside the editor

It doesn't show because the editor is doing hundreds of rerenders over and over again until you click outside the editor.

> Both of these issues are not present if we don't use the `ScrollContainer`. I've tried several ways to make it work, but not sure what's causing it or how to fix it; will investigate further...

We could try a different custom scrollbar library. Perhaps it'll work better?

packages/editor/package.json
```diff
@@ -79,6 +79,7 @@
     "react-colorful": "^5.6.1",
     "redent": "^4.0.0",
     "refractor": "^4.8.1",
+    "simplebar-react": "^3.3.2",
```
01zulfi, line 82: [simplebar](https://github.com/Grsmto/simplebar/tree/master) is working great out of the box

PR #8493, opened 2025-08-22
editor: make headings collapsible

01zulfi:
https://github.com/user-attachments/assets/eac7e947-3713-45ca-912b-ef0689437ca0

### DIFF
--- a/apps/web/src/components/editor/index.tsx
+++ b/apps/web/src/components/editor/index.tsx
@@ -703,8 +703,10 @@ function EditorChrome(props: PropsWithChildren<EditorProps>) {
       );
 
       requestAnimationFrame(() => {
-        editor.style.marginLeft = `-${negativeSpace}px`;
-        editor.style.marginRight = `-${negativeSpace}px`;
+        if (!isMobile() && !isTablet()) {
+          editor.style.marginLeft = `-${negativeSpace}px`;
+          editor.style.marginRight = `-${negativeSpace}px`;
+        }
         editor.style.paddingLeft = `${negativeSpace}px`;
         editor.style.paddingRight = `${negativeSpace}px`;
       });
--- a/packages/editor/src/extensions/heading/heading.ts
+++ b/packages/editor/src/extensions/heading/heading.ts
@@ -18,11 +18,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { tiptapKeys } from "@notesnook/common";
-import { textblockTypeInputRule } from "@tiptap/core";
+import {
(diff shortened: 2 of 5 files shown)


apps/web/src/components/editor/index.tsx
```diff
@@ -703,8 +703,10 @@ function EditorChrome(props: PropsWithChildren<EditorProps>) {
       );
 
       requestAnimationFrame(() => {
-        editor.style.marginLeft = `-${negativeSpace}px`;
-        editor.style.marginRight = `-${negativeSpace}px`;
+        if (!isMobile() && !isTablet()) {
+          editor.style.marginLeft = `-${negativeSpace}px`;
+          editor.style.marginRight = `-${negativeSpace}px`;
+        }
```
01zulfi, line 709: have to remove `marginLeft` for small screens otherwise the collapse heading icon gets cut off.
removing `marginRight` for consistency

packages/editor/src/extensions/heading/heading.ts
```diff
@@ -71,5 +100,140 @@ export const Heading = TiptapHeading.extend({
...
+              tr.setNodeAttribute(
+                pos,
+                "collapsed",
+                !currentNode.attrs.collapsed
+              );
+            }
+            return true;
+          });
+        }
+      }
```
thecodrr: This is the same function as used in Outline right? Maybe we can share the logic?

thecodrr (changes requested):
Good work!

packages/editor/src/extensions/heading/heading.ts
01zulfi: I tried to make the entire onClick reusable, but I felt we were passing in a lot of parameters and conditional logic making the function difficult to understand. I've made a util function for getting the click bounds.

packages/editor/src/extensions/callout/callout.ts
```diff
@@ -231,37 +234,16 @@ export const Callout = Node.create({
...
-        if (xStart && xEnd && yStart && yEnd) {
+        const resolvedPos = editor.state.doc.resolve(pos);
+        const bounds = getCollapsibleClickBounds(e, resolvedPos, "right");
+        if (
+          bounds &&
+          bounds.xStart &&
+          bounds.xEnd &&
+          bounds.yStart &&
+          bounds.yEnd
+        ) {
```
thecodrr: Why are we checking all these instead of a single condition?

thecodrr (changes requested):

thecodrr:
@01zulfi can you resolve the conflict so I can merge this?

Issue #8526, opened 2025-09-04
Notesnook Inbox API - send notes from third-party services to NN


01zulfi:
# What

A public Inbox API that'll allow any third-party service to send notes to a NN user's account

# How

1. Enabling Inbox API will create a public-private key pair for the user
2. Incoming inbox data will be encrypted with public key
3. On user's device, the inbox data will be decrypted with private key and dumped into the database
4. The public API will be protected via an API key generated by the user

# Acceptance Criteria

- [x] Create public-private key pair for the Inbox
- [x] Create API key generation for the public Inbox API
- [x] Implement `GET` public key and `POST` encrypted inbox items endpoints 
- [x] Create self-hostable public Inbox API server
- [x] Handle inbox item sync and client-side decryption
- [ ] Write docs for using the inbox api && selfhosting the inbox api && creating the inbox api from scratch

PR #8527, opened 2025-09-04
global: add enable inbox setting to generate/discard inbox key pair

01zulfi:

### DIFF
--- a/packages/core/src/api/user-manager.ts
+++ b/packages/core/src/api/user-manager.ts
@@ -24,7 +24,7 @@ import TokenManager from "./token-manager.js";
 import { EV, EVENTS } from "../common.js";
 import { HealthCheck } from "./healthcheck.js";
 import Database from "./index.js";
-import { SerializedKey } from "@notesnook/crypto";
+import { SerializedKeyPair, SerializedKey, Cipher } from "@notesnook/crypto";
 import { logger } from "../logger.js";
 
 const ENDPOINTS = {
@@ -44,6 +44,7 @@ class UserManager {
   private tokenManager: TokenManager;
   private cachedAttachmentKey?: SerializedKey;
   private cachedMonographPasswordsKey?: SerializedKey;
+  private cachedInboxKeys?: SerializedKeyPair;
   constructor(private readonly db: Database) {
     this.tokenManager = new TokenManager(this.db.kv);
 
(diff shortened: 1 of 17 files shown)


apps/web/src/stores/app-store.ts
```diff
@@ -70,6 +70,7 @@ class AppStore extends BaseStore<AppStore> {
   isVaultCreated = false;
   isAutoSyncEnabled = Config.get("autoSyncEnabled", true);
   isSyncEnabled = Config.get("syncEnabled", true);
+  isInboxEnabled = Config.get("inboxEnabled", false);
```
thecodrr: I think this should be based directly on whether user has `inboxKeys` or not instead of a locally stored value.

packages/core/src/utils/crypto.ts
```diff
@@ -29,6 +29,10 @@ export class Crypto {
     const password = passwordBytes.toString("base64");
     return await this.storage().generateCryptoKey(password);
   }
+
+  async generateRandomKeyPair() {
```
thecodrr: The keypair is already random so let's just call it
```suggestion
  async generateCryptoKeyPair() {
```

packages/core/src/api/user-manager.ts
```diff
@@ -460,6 +462,58 @@ class UserManager {
...
+      this.cachedInboxKeys = plainData;
+      return this.cachedInboxKeys;
+    } catch (e) {
+      logger.error(e, "Could not get inbox encryption keys.");
+      if (e instanceof Error)
+        throw new Error(
+          `Could not get inbox encryption keys. Error: ${e.message}`
+        );
+    }
+  }
```
thecodrr: We now have 3 methods that are more of the same. Let's refactor them so they share the common bits.

apps/web/src/stores/app-store.ts
```diff
@@ -70,6 +70,7 @@ class AppStore extends BaseStore<AppStore> {
   isVaultCreated = false;
   isAutoSyncEnabled = Config.get("autoSyncEnabled", true);
   isSyncEnabled = Config.get("syncEnabled", true);
+  isInboxEnabled = false;
```
thecodrr: Wait. Why are we doing this in `app-store` instead of `setting-store`?

01zulfi: Because I think it might be needed in places other than settings, do you think it should be part of the settings store?

thecodrr: Yes.

PR #8534, opened 2025-09-05
web: add settings export & import

01zulfi:

### DIFF
--- a/apps/web/src/dialogs/settings/settings-import-export.ts
+++ b/apps/web/src/dialogs/settings/settings-import-export.ts
@@ -1,0 +1,230 @@
+/*
+This file is part of the Notesnook project (https://notesnook.com/)
+
+Copyright (C) 2023 Streetwriters (Private) Limited
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { useStore as useSettingStore } from "../../stores/setting-store";
+import { saveAs } from "file-saver";
+import { showToast } from "../../utils/toast";
+import { showFilePicker, readFile } from "../../utils/file-picker";
+import { SettingsGroup } from "./types";
+import { sectionGroups, SettingsGroups } from ".";
+import { ServersSettings } from "./servers-settings";
+import { strings } from "@notesnook/intl";
+
+type ExportedSetting = Record<string, string | number | boolean>;
+
+interface ExportedSettings {
+  exportDate: string;
+  settings: Record<string, ExportedSetting>;
+}
+
+async function extractSettingsValues(
+  settingsGroups: SettingsGroup[]
+): Promise<ExportedSetting> {
+  const exportedSettings: ExportedSetting = {};
+
+  for (const group of settingsGroups) {
+    for (const setting of group.settings) {
+      if (setting.isHidden?.()) continue;
+
+      const components =
+        typeof setting.components === "function"
+          ? setting.components()
+          : setting.components;
+
+      for (const component of components) {
+        if (component.type === "custom") continue;
+        if (component.type === "dropdown" && "selectedOption" in component) {
+          const value =
+            typeof component.selectedOption === "function"
+              ? await component.selectedOption()
+              : component.selectedOption;
+          if (value === "-" || value === undefined) continue;
+          exportedSettings[setting.key] = value;
+        } else if (component.type === "toggle" && "isToggled" in component) {
+          const value =
+            typeof component.isToggled === "function"
+              ? component.isToggled()
+              : component.isToggled;
+          exportedSettings[setting.key] = value;
+        } else if (component.type === "input" && "defaultValue" in component) {
+          const value =
+            typeof component.defaultValue === "function"
+              ? component.defaultValue()
+              : component.defaultValue;
+          exportedSettings[setting.key] = value;
+        }
+      }
+    }
+  }
+
+  return exportedSettings;
+}
+
+export async function exportSettings(): Promise<void> {
+  try {
+    const toExport: ExportedSettings["settings"] = {};
+
+    for (const group of SettingsGroups) {
+      if (group.isHidden?.()) continue;
+      if (
+        sectionGroups.find((sg) =>
+          sg.sections.find((s) => s.key === group.section)?.isHidden?.()
+        )
+      ) {
+        continue;
+      }
+
+      const groupSettingsValues = await extractSettingsValues([group]);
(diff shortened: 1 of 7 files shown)


apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+
+      for (const component of components) {
+        if (component.type === "custom") continue;
+        if (component.type === "dropdown" && "selectedOption" in component) {
+          const value =
+            typeof component.selectedOption === "function"
+              ? await component.selectedOption()
+              : component.selectedOption;
+          if (value === "-" || value === undefined) continue;
+          exportedSettings[setting.key] = value;
```
thecodrr, line 65: Shouldn't it be `exportedSettings.settings[setting.key]`?

thecodrr (changes requested):
This looks mostly good.

apps/web/src/dialogs/settings/settings-import-export.ts
01zulfi: that is happening outside of this function

thecodrr: Okay that's confusing. I'll take another look. I think the import/export logic can be simplified.

01zulfi: Here is a sample file:

```json
{
  "exportDate": "2025-09-05T05:20:55.251Z",
  "settings": {
    "theme": {
      "color-scheme": "light"
    },
    "date-time": {
      "date-format": "DD-MM-YYYY",
      "time-format": "12-hour"
    },
    "trash": {
      "trash-cleanup-interval": "7"
    },
    "reminders": {
      "reminders": false
    },
    "backup": {
      "auto-backup": "0",
      "auto-backup-with-attachments": "0"
    },
    "app-lock": {
      "enable-app-lock": false
    },
    "general": {
      "hide-note-title": false
    },
    "advanced": {
      "proxy-config": ""
    },
    "editor": {
      "default-title": "$headline",
      "default-font": "sans-serif",
      "default-font-size": 14,
      "double-spacing": true,
      "markdown-shortcuts": false,
      "font-ligatures": true
    },
    "about": {
      "release-track": "beta"
    },
    "servers": {}
  }
}
```

thecodrr: The output looks good.

apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    for (const group of SettingsGroups) {
+      if (group.isHidden?.()) continue;
+      if (
+        sectionGroups.find((sg) =>
+          sg.sections.find((s) => s.key === group.section)?.isHidden?.()
+        )
+      ) {
+        continue;
+      }
```
thecodrr: Shouldn't we check all the the sections instead of returning even if only 1 section is hidden?

apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+
+type ExportedSetting = Record<string, string | number | boolean>;
+
+interface ExportedSettings {
+  exportDate: string;
+  settings: Record<string, ExportedSetting>;
+}
+
+async function extractSettingsValues(
+  settingsGroups: SettingsGroup[]
```
thecodrr: Why is this an array?

apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+      if (setting.isHidden?.()) continue;
+
+      const components =
+        typeof setting.components === "function"
+          ? setting.components()
+          : setting.components;
+
+      for (const component of components) {
+        if (component.type === "custom") continue;
+        if (component.type === "dropdown" && "selectedOption" in component) {
```
thecodrr: Maybe we should use `switch..case` (the default case should throw) here to make sure we are doing the check is exhaustive. We don't want to miss a type.

apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+          const value =
+            typeof component.selectedOption === "function"
+              ? await component.selectedOption()
+              : component.selectedOption;
+          if (value === "-" || value === undefined) continue;
```
thecodrr: What if `value` is `""`?

apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+      ) {
+        continue;
+      }
+
+      const groupSettingsValues = await extractSettingsValues([group]);
+      if (Object.keys(groupSettingsValues).length === 0) continue;
+      toExport[group.key] = groupSettingsValues;
+    }
+
+    toExport[ServersSettings[0].key] = useSettingStore.getState().serverUrls;
```
thecodrr, line 110: Why are we doing this separately?

apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    const exportData: ExportedSettings = {
+      exportDate: new Date().toISOString(),
+      settings: toExport
+    };
+
+    const fileName = `notesnook-settings-${
+      new Date().toISOString().split("T")[0]
+    }.json`;
```
thecodrr: Use `dayjs`

apps/web/src/dialogs/settings/settings-import-export.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+}
+
+async function selectSettingsFile() {
+  const [file] = await showFilePicker({
+    acceptedFileTypes: ".json"
+  });
+  if (!file) return;
+  return file;
+}
```
thecodrr: Why is this a separate function?

apps/web/src/dialogs/settings/settings-import-export.ts
01zulfi: sorry didn't get you, if any one of the section that the group belongs to is hidden we should skip export/import

We could check every section for that group, but I'm trying to understand how that would be beneficial

apps/web/src/dialogs/settings/settings-import-export.ts
01zulfi: this is in custom setting component so requires special handling during export/import

I also added special handling for the editor toolbar setting

apps/web/src/dialogs/settings/settings-import-export.ts
thecodrr: > sorry didn't get you, if any one of the section that the group belongs to is hidden we should skip export/import

Why? Doesn't each section contain different settings?

thecodrr:
We forgot about making this work cross-platform. Any ideas on how we can do that?

01zulfi:
I think we can make the mobile export/import have the same json structure as the web one. We can flatten the json so that it only includes setting id and not the group id, so that we don't have to deal with the case where groups aren't identical between the apps.

We could move all the setting ids into the core/common package and have the apps reference them.

thecodrr:
> We could move all the setting ids into the core/common package and have the apps reference them.

That's a good idea, actually.

PR #8547, opened 2025-09-10
editor: handle enter key on headings

01zulfi:

### DIFF
--- a/packages/editor/src/extensions/heading/heading.ts
+++ b/packages/editor/src/extensions/heading/heading.ts
@@ -61,31 +61,84 @@ export const Heading = TiptapHeading.extend({
             textAlign,
             textDirection
           });
-        },
-      toggleHeadingCollapse:
-        (pos: number) =>
-        ({ tr }: { tr: any }) => {
-          const node = tr.doc.nodeAt(pos);
-          if (node && node.type === this.type) {
-            tr.setNodeAttribute(pos, "collapsed", !node.attrs.collapsed);
-            return true;
-          }
-          return false;
         }
     };
   },
 
   addKeyboardShortcuts() {
-    return this.options.levels.reduce(
-      (items, level) => ({
-        ...items,
-        ...{
-          [tiptapKeys[`insertHeading${level}`].keys]: () =>
-            this.editor.commands.setHeading({ level })
+    return {
+      ...this.options.levels.reduce(
+        (items, level) => ({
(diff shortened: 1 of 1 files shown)


01zulfi:
There are two enter cases we need to consider.

## 1. Enter at end of the collapsed heading

I'm un-collapsing the heading and letting the default enter action take place

> let's just move the cursor below all the hidden nodes
All the nodes under that collapsed heading would be hidden. Maybe we can move the cursor to the next unhidden heading?

## 2. Enter at start of the heading

I thought about just ignoring the enter key altogether if the previous heading was collapsed. Though in the PR I added the behavior that it un-collapses all the previous headings up until level 1.

thecodrr:
> All the nodes under that collapsed heading would be hidden. Maybe we can move the cursor to the next unhidden heading?

We should move the cursor to the end of the hidden content for that specific heading. This is exactly how collapsed sections in code editors work. You can collapse a section and edit below it without it getting uncollapsed.

> 2. Enter at start of the heading
> I thought about just ignoring the enter key altogether if the previous heading was collapsed. Though in the PR I added the behavior that it un-collapses all the previous headings up until level 1.

I don't think that is necessary. Enter shouldn't affect the collapsed state of any heading. It should add a new line. That's the expected behavior.

01zulfi:
> This is exactly how collapsed sections in code editors work.

when I hit enter at the end of the heading in VSCode, it un-collapses the heading and then goes on with the enter action

> You can collapse a section and edit below it without it getting uncollapsed.

got it, will try to figure this out

> I don't think that is necessary. Enter shouldn't affect the collapsed state of any heading. It should add a new line. That's the expected behavior.

got it

thecodrr:
> when I hit enter at the end of the heading in VSCode, it un-collapses the heading and then goes on with the enter action

True. In Zed I am noticing a different behavior. Pressing Enter at the end doesn't uncollapse the section, it just adds a new line at the end of the block. Hmm...

I don't think it's possible to do what Zed is doing since we hide _everything_ under a heading including anything new that the user adds. We'll have to change the way we hide nodes. Instead of using Decorations, we can use global attributes. The global attribute will be added at the time user presses the collapse/expand button of the heading. That way, if a new content is added later on, it won't be hidden.

01zulfi:
closing in favor of https://github.com/streetwriters/notesnook/pull/8557

PR #8552, opened 2025-09-11
web: add manage inbox api tokens section

01zulfi:

image

### DIFF
--- a/packages/core/src/api/user-manager.ts
+++ b/packages/core/src/api/user-manager.ts
@@ -17,7 +17,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { User, InboxApiKey } from "../types.js";
+import { User } from "../types.js";
 import http from "../utils/http.js";
 import constants from "../utils/constants.js";
 import TokenManager from "./token-manager.js";
@@ -37,8 +37,7 @@ const ENDPOINTS = {
   revoke: "/connect/revocation",
   recoverAccount: "/account/recover",
   resetUser: "/users/reset",
-  activateTrial: "/subscriptions/trial",
-  inboxApiKeys: "/inbox/api-keys"
+  activateTrial: "/subscriptions/trial"
 };
 
 class UserManager {
@@ -579,52 +578,6 @@ class UserManager {
     );
 
     await this.setUser({ ...user, inboxKeys: undefined });
-  }
-
-  async getInboxApiKeys() {
-    const user = await this.getUser();
-    if (!user) return;
-
-    const token = await this.tokenManager.getAccessToken();
-    if (!token) return;
-
-    const inboxApiKeys = await http.get(
-      `${constants.API_HOST}${ENDPOINTS.inboxApiKeys}`,
-      token
-    );
-    return inboxApiKeys as InboxApiKey[];
-  }
-
-  async revokeInboxApiKey(key: string) {
-    const user = await this.getUser();
-    if (!user) return;
-
-    const token = await this.tokenManager.getAccessToken();
-    if (!token) return;
-
-    await http.delete(
-      `${constants.API_HOST}${ENDPOINTS.inboxApiKeys}/${key}`,
-      token
-    );
-  }
-
-  async createInboxApiKey(name: string, expiryDuration: number) {
-    const user = await this.getUser();
-    if (!user) return;
-
-    const token = await this.tokenManager.getAccessToken();
-    if (!token) return;
-
-    const payload: Omit<InboxApiKey, "lastUsedAt" | "key"> = {
-      name,
-      dateCreated: Date.now(),
-      expiryDate: expiryDuration === -1 ? -1 : Date.now() + expiryDuration
-    };
-    await http.post.json(
-      `${constants.API_HOST}${ENDPOINTS.inboxApiKeys}`,
-      payload,
-      token
-    );
   }
 
   async sendVerificationEmail(newEmail?: string) {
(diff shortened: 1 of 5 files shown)


apps/web/src/dialogs/settings/components/inbox-api-tokens.tsx
```diff
@@ -0,0 +1,348 @@
...
+      title="Token Limit Reached"
+      description="Cannot create more than 10 tokens at a time. Please revoke some existing tokens before creating new ones."
+      onClose={() => onClose(false)}
+      positiveButton={{
+        text: "OK",
+        onClick: () => onClose(false)
+      }}
+    />
+  );
+});
```
thecodrr: Just use `ConfirmDialog` instead of creating a new dialog.

apps/web/src/dialogs/settings/components/inbox-api-tokens.tsx
```diff
@@ -0,0 +1,348 @@
...
+    </Dialog>
+  );
+});
+
+type RevokeTokenDialogProps = BaseDialogProps<boolean> & {
+  token: InboxApiToken;
+  onRevoke: () => void;
+};
+
+const RevokeTokenDialog = DialogManager.register(function RevokeTokenDialog(
```
thecodrr: Just use `ConfirmDialog` instead of creating a new dialog.

apps/web/src/dialogs/settings/components/inbox-api-tokens.tsx
```diff
@@ -0,0 +1,348 @@
...
+      </Flex>
+    </Box>
+  );
+}
+
+type AddTokenDialogProps = BaseDialogProps<boolean> & {
+  onAdd: () => void;
+};
+
+const AddTokenDialog = DialogManager.register(function AddTokenDialog(
```
thecodrr: Use the `PromptDialog` instead of creating a new dialog.

apps/web/src/dialogs/settings/components/inbox-api-tokens.tsx
```diff
@@ -0,0 +1,348 @@
...
+            paddingY: 1,
+            paddingX: 2,
+            fontFamily: "monospace",
+            flex: 1,
+            bg: "background-secondary"
+          }}
+        />
+        <Button variant="icon" onClick={() => copyToClipboard(token.token)}>
+          {copied ? <Check size={14} /> : <Copy size={14} />}
+        </Button>
```
thecodrr: The api key should be shown only once after generation. That's the standard everywhere. Or we can ask for auth before exposing the api key?

apps/web/src/dialogs/settings/components/inbox-api-tokens.tsx
thecodrr: There should also be an option to set a custom expiration (with an option to never expire).

thecodrr: By custom I don't mean a calendar based date. It should show a number of options like (1 year, 1 month, 1 week, never etc.)

apps/web/src/dialogs/settings/components/inbox-api-tokens.tsx
01zulfi: I'm asking for password on copy, let me know if you have something else in mind

thecodrr: Instead of a copy button, how about a "view" button? The api key will be obfuscated by default. When user presses the view button, we'll ask for password, expose the api key, and also copy it to clipboard.

01zulfi: This is what I'm doing right now:

1. Viewing requires authorization, which on success triggers a 15s countdown
2. During this period the key is exposed and the copy button is available


https://github.com/user-attachments/assets/b29fefd3-7a41-422d-957b-8360b57b6f05

thecodrr: That works too.

packages/core/src/api/user-manager.ts
```diff
@@ -580,6 +581,52 @@ class UserManager {
...
+      name,
+      dateCreated: Date.now(),
+      expiryDate: expiryDuration === -1 ? -1 : Date.now() + expiryDuration
+    };
+    await http.post.json(
+      `${constants.API_HOST}${ENDPOINTS.inboxApiKeys}`,
+      payload,
+      token
+    );
+  }
```
thecodrr: Let's move these methods to their own class.

packages/core/src/api/user-manager.ts
```diff
@@ -580,6 +581,52 @@ class UserManager {
...
+  async createInboxApiKey(name: string, expiryDuration: number) {
+    const user = await this.getUser();
+    if (!user) return;
+
+    const token = await this.tokenManager.getAccessToken();
+    if (!token) return;
+
+    const payload: Omit<InboxApiKey, "lastUsedAt" | "key"> = {
+      name,
+      dateCreated: Date.now(),
```
thecodrr: No need to set this here. Set it on the server.

packages/core/src/api/user-manager.ts
```diff
@@ -580,6 +581,52 @@ class UserManager {
...
+      expiryDate: expiryDuration === -1 ? -1 : Date.now() + expiryDuration
```
thecodrr: Send expiry duration to the server instead of the expiry date. We can calculate the expiry date on the server.

thecodrr: Actually scratch that. It's okay to set this here.

thecodrr (approved):

PR #8557, opened 2025-09-12
editor: make headings collapsible

01zulfi:
* use a global attribute to hide hidden nodes

### DIFF
--- a/packages/editor/src/extensions/heading/heading.ts
+++ b/packages/editor/src/extensions/heading/heading.ts
@@ -18,11 +18,49 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 import { tiptapKeys } from "@notesnook/common";
-import { textblockTypeInputRule } from "@tiptap/core";
+import {
+  findParentNodeClosestToPos,
+  textblockTypeInputRule
+} from "@tiptap/core";
 import { Heading as TiptapHeading } from "@tiptap/extension-heading";
+import { isClickWithinBounds } from "../../utils/prosemirror";
+import { Selection, Transaction } from "@tiptap/pm/state";
+import { Node } from "@tiptap/pm/model";
+
+const COLLAPSIBLE_BLOCK_TYPES = [
+  "paragraph",
+  "heading",
+  "blockquote",
+  "bulletList",
+  "orderedList",
+  "checkList",
+  "taskList",
+  "table",
+  "callout",
+  "codeblock",
+  "image",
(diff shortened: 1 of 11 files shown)


packages/editor/src/extensions/code-block/component.tsx
```diff
@@ -49,6 +49,8 @@ export function CodeblockComponent(
     (l) => l.filename === language || l.alias?.some((a) => a === language)
   );
 
+  if (node.attrs.hiddenUnder) return null;
```
thecodrr: Is this necessary to add on every react node? Maybe there's something that can be added in `react-node-view.tsx` that does the same thing for all react nodes.

packages/editor/src/extensions/heading/heading.ts
```diff
@@ -206,9 +227,91 @@ export const Heading = TiptapHeading.extend({
...
+      isCollapsing &&
+      cursorPos >= nextPos &&
+      cursorPos < nextPos + nextNode.nodeSize
+    ) {
+      shouldMoveCursor = true;
+    }
+
+    if (COLLAPSIBLE_BLOCK_TYPES.includes(nextNode.type.name)) {
+      if (isCollapsing && typeof nextNode.attrs.hiddenUnder !== "string") {
+        tr.setNodeAttribute(nextPos, "hiddenUnder", headingId);
```
thecodrr, line 287: Is there any specific reason we are storing this id in each attribute?

packages/editor/src/extensions/code-block/component.tsx
01zulfi: It wasn't required for attachment or image react nodes, but adding the check in `react-node-view` sounds like a good idea

packages/editor/src/extensions/heading/heading.ts
01zulfi: This is for handling nested collapsed headings. So, when we un-collapse a heading it should only unhide the nodes that belongs to it. If there are any nested collapsed heading it won't get un-collapsed because a parent heading got un-collapsed

thecodrr: That's good thinking.

thecodrr:
@01zulfi  can you resolve the conflicts?

01zulfi:
Done, now this PR restores the revert https://github.com/streetwriters/notesnook/commit/a0597e13fee8dbee90c15960646822a41e219076 and refactors the collapse implementation

PR #8561, opened 2025-09-13
web: refactor server URL configuration to use environment variables

BeardedTek:
This PR adds the ability to set the following server URL's via Environment variables for the web app to be set at both build time and at docker container startup:
 - Sync Server
 - Identity Server
 - SSE Server
 - Monograph Server

This should allow for easier self hosting of the app as outline in https://github.com/beardedtek/notesnook-docker

### DIFF
--- a/apps/web/src/common/db.ts
+++ b/apps/web/src/common/db.ts
@@ -39,14 +39,33 @@ async function initializeDatabase(persistence: DatabasePersistence) {
     await useKeyStore.getState().setValue("databaseKey", databaseKey);
   }
 
+  // Get server URLs from environment variables (with NN_ prefix) or user config
+  const getServerUrl = (host: string, defaultValue: string): string => {
+    // Check for environment variable with NN_ prefix first
+    const envVar = `NN_${host}`;
+    const envValue = (import.meta as any).env[envVar];
+    if (envValue) {
+      return envValue;
+    }
+    
+    // Fall back to user config
+    const configUrls = Config.get("serverUrls", {}) as Record<string, string>;
+    const configValue = configUrls[host];
+    if (configValue) {
+      return configValue;
+    }
+    
+    // Fall back to default
+    return defaultValue;
+  };
+
   db.host({
-    API_HOST: "https://api.notesnook.com",
-    AUTH_HOST: "https://auth.streetwriters.co",
-    SSE_HOST: "https://events.streetwriters.co",
-    ISSUES_HOST: "https://issues.streetwriters.co",
-    MONOGRAPH_HOST: "https://monogr.ph",
-    SUBSCRIPTIONS_HOST: "https://subscriptions.streetwriters.co",
-    ...Config.get("serverUrls", {})
+    API_HOST: getServerUrl("API_HOST", "https://api.notesnook.com"),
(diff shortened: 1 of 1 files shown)


thecodrr:
Thank you.

apps/web/src/common/db.ts
```diff
@@ -39,14 +39,33 @@ async function initializeDatabase(persistence: DatabasePersistence) {
     await useKeyStore.getState().setValue("databaseKey", databaseKey);
   }
 
+  // Get server URLs from environment variables (with NN_ prefix) or user config
+  const getServerUrl = (host: string, defaultValue: string): string => {
+    // Check for environment variable with NN_ prefix first
+    const envVar = `NN_${host}`;
+    const envValue = (import.meta as any).env[envVar];
```
thecodrr, line 46: This is a bad idea. Why are you casting to `as any`?

apps/web/src/common/db.ts
```diff
@@ -39,14 +39,33 @@ async function initializeDatabase(persistence: DatabasePersistence) {
     await useKeyStore.getState().setValue("databaseKey", databaseKey);
   }
 
+  // Get server URLs from environment variables (with NN_ prefix) or user config
+  const getServerUrl = (host: string, defaultValue: string): string => {
```
thecodrr, line 43: Use the `hosts` type from `@notesnook/core`.

```suggestion
  const getServerUrl = (host: keyof typeof hosts, defaultValue: string): string => {
```

apps/web/src/common/db.ts
```diff
@@ -39,14 +39,33 @@ async function initializeDatabase(persistence: DatabasePersistence) {
...
+    // Check for environment variable with NN_ prefix first
+    const envVar = `NN_${host}`;
+    const envValue = (import.meta as any).env[envVar];
+    if (envValue) {
+      return envValue;
+    }
+    
+    // Fall back to user config
```
thecodrr, line 51: You don't need to add these unnecessary comments.

apps/web/src/common/db.ts
thecodrr: One other thing: the `getServerUrl` should be outside the `initializeDatabase` function. We don't really do local functions.

apps/web/src/common/db.ts
BeardedTek: Honestly, I'm mainly a python developer and just trying to get the ball rolling.  Thank you for your guidance.  I'll make the changes in the AM (GMT-8 here in Alaska)

thecodrr: No problem. That's how we learn :)

thecodrr:
bump @BeardedTek

BeardedTek:
I've been running around putting out fires this week.  It's still on my list, just fell down a bit.

01zulfi:
Hey @BeardedTek, are you still looking to finish this? If not, that's alright, I can wrap this up.

BeardedTek:
@01zulfi go ahead.  I've gotten pretty busy with other project.

01zulfi:
closing in favor of https://github.com/streetwriters/notesnook/pull/9351

PR #8568, opened 2025-09-15
web: fix settings not opening again after navigation from attachment manager

01zulfi:

### DIFF
--- a/apps/web/src/common/dialog-manager.tsx
+++ b/apps/web/src/common/dialog-manager.tsx
@@ -45,7 +45,7 @@ class _DialogManager {
         onClose: (result: any) => void;
       }
     >;
-    return new Promise<Result>((resolve, reject) => {
+    return new Promise<Result | false>((resolve, reject) => {
       const close = () => {
         this.openedDialogs.delete(component);
         root.unmount();
@@ -53,7 +53,7 @@ class _DialogManager {
       };
       this.openedDialogs.set(component, () => {
         close();
-        reject(new Error("Dialog force closed."));
+        resolve(false);
       });
       root.render(
         <Dialog
@@ -72,11 +72,14 @@ class _DialogManager {
   }
 
   closeAll() {
-    this.openedDialogs.clear();
     const dialogs = document.querySelectorAll(
       ".ReactModalPortal,[data-react-modal-body-trap]"
     );
     dialogs.forEach((elem) => elem.remove());
+
+    for (const close of this.openedDialogs.values()) {
+      close();
+    }
   }
 
   register<Props extends BaseDialogProps<any>>(


apps/web/src/components/attachment/index.tsx
```diff
@@ -272,6 +273,7 @@ const AttachmentMenuItems: (
                   onClick: () => {
                     useEditorStore.getState().openSession(note.id);
                     DialogManager.closeAll();
+                    hashNavigate("/", { replace: true, notify: false });
```
thecodrr: This feels like a temporary solution. Why exactly can't the settings open after navigation and why do we need this?

thecodrr (changes requested):

apps/web/src/components/attachment/index.tsx
01zulfi: When the settings dialog is closed, we reset the hash router:
https://github.com/streetwriters/notesnook/blob/master/apps/web/src/navigation/hash-routes.tsx#L77

But in this attachment component, we call the `closeAll` method, which doesn't trigger the close action for the dialog, just removes it from the DOM: https://github.com/streetwriters/notesnook/blob/master/apps/web/src/common/dialog-manager.tsx#L74

That's why I'm resetting the hash router manually here

thecodrr: How about we fix the `closeAll` method? I think we should get rid of the promise rejection on force closing a dialog. There's no utility to it and it just makes things worse. We should call close for all the opened dialogs and allow them to close gracefully.

apps/web/src/common/dialog-manager.tsx
```diff
@@ -45,15 +45,15 @@ class _DialogManager {
...
+    return new Promise<Result | false>((resolve, reject) => {
       const close = () => {
         this.openedDialogs.delete(component);
         root.unmount();
         container.remove();
       };
       this.openedDialogs.set(component, () => {
         close();
-        reject(new Error("Dialog force closed."));
+        resolve(false);
```
01zulfi, line 56: I'm resolving the promise with `resolve(false)`, don't think it should be an issue, just wanted to highlight this

apps/web/src/components/attachment/index.tsx
```diff
@@ -59,6 +59,7 @@ import { PromptDialog } from "../../dialogs/prompt";
 import { DialogManager } from "../../common/dialog-manager";
 import { useStore as useSelectionStore } from "../../stores/selection-store";
 import { strings } from "@notesnook/intl";
+import { hashNavigate } from "../../navigation";
```
thecodrr:
```suggestion
```

apps/web/src/common/dialog-manager.tsx
```diff
@@ -72,11 +72,9 @@ class _DialogManager {
   }
 
   closeAll() {
-    this.openedDialogs.clear();
-    const dialogs = document.querySelectorAll(
-      ".ReactModalPortal,[data-react-modal-body-trap]"
-    );
-    dialogs.forEach((elem) => elem.remove());
```
thecodrr: Should we keep this just in case?

thecodrr (approved):

PR #8577, opened 2025-09-16
editor: Add keybindings for highlighting the currently selected text

Sudlon:
Add the following keybindings for highlighting the currently selected text in the editor:
- Mod + Shift + H
- Mod + Alt + H

### DIFF
--- a/docs/help/contents/keyboard-shortcuts.md
+++ b/docs/help/contents/keyboard-shortcuts.md
@@ -76,3 +76,4 @@ The following keyboard shortcuts will help you navigate Notesnook faster.
 | Text align left | Ctrl ⇧ L | Ctrl ⇧ L | ⌘ ⇧ L |
 | Text align right | Ctrl ⇧ R | Ctrl ⇧ R | ⌘ ⇧ R |
 | Underline | Ctrl U | Ctrl U | ⌘ U |
+| Toggle highlight | Ctrl Alt H / Ctrl ⇧ H | Ctrl Alt H / Ctrl ⇧ H | ⌘ ⌥ H / ⌘ ⇧ H |
--- a/packages/common/src/utils/keybindings.ts
+++ b/packages/common/src/utils/keybindings.ts
@@ -388,6 +388,12 @@ export const tiptapKeys = {
     description: "Underline",
     category: "Editor",
     type: "tiptap"
+  },
+  toggleHighlight: {
+    keys: ["ctrl+alt+h", "ctrl+shift+h"],
+    description: "Toggle highlight",
+    category: "Editor",
+    type: "tiptap"
   }
 } satisfies Record<string, TipTapKey>;
 
--- a/packages/editor/src/extensions/highlight/highlight.ts
+++ b/packages/editor/src/extensions/highlight/highlight.ts
@@ -19,6 +19,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import "@tiptap/extension-text-style";
 import { Extension } from "@tiptap/core";
+import { config } from "../../utils/config.js";
+import { tiptapKeys } from "@notesnook/common";
 
 export interface HighlightOptions {
   types: string[];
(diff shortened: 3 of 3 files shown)


packages/common/src/utils/keybindings.ts
```diff
@@ -388,6 +388,12 @@ export const tiptapKeys = {
     description: "Underline",
     category: "Editor",
     type: "tiptap"
+  },
+  toggleHighlight: {
+    keys: ["ctrl+alt+h", "ctrl+shift+h"],
```
thecodrr: Why both of these combinations?

thecodrr (changes requested):

packages/common/src/utils/keybindings.ts
Sudlon: Both are used in other apps, so I thought this would create the smoothest experience for most people making the transition. Also, I know OneNote allows both keybindings, so it is is not uncommon to have (these) two, apparently.

If you want only one and need a vote: I myself prefer "ctrl+alt+h".

01zulfi: Can we use `Mod-Alt-h` instead of `ctrl+alt+h` to follow the tiptap key naming convention?

01zulfi (changes requested):

packages/common/src/utils/keybindings.ts
Sudlon: Ah yes of course. Will update right away

thecodrr: Let's use one since there's only so many keyboard shortcuts possible. We should use the one used by most apps out there.

Sudlon: A quick online search gave me:
- ALT-version: OneNote, MS Word (I guess the entire MS Office)
- SHIFT-version: Evernote, Notion
- OTHER: Obsidian (custom keybindings), OfficeLibre (custom keybindings), Bear (ctrl+shift+m), StandardNotes: no idea, but NOT ctrl+shift+h because that is used for something else

There is not really one used by most apps it seems. It would be awesome if users could overwrite the default keybindings with there own custom binding, like Obsidian and OfficeLibre, but that's for another time. :)

So yeah, I guess with both ALT and SHIFT versions you have the biggest other apps/user bases covered, but if you just want one (and I definitely get your point), then we need to choose based on preference. You're the boss, so let me know what you want and I'll make the change accordingly.

thecodrr: Let's go with `ctrl + alt + h`

Sudlon: Done! 

Also re-ran the "document-keyboard-shortcuts" script

packages/editor/src/extensions/highlight/highlight.ts
```diff
@@ -99,11 +101,16 @@ export const Highlight = Extension.create<HighlightOptions>({
...
 
-  // addKeyboardShortcuts() {
-  //   return {
-  //     "Mod-Shift-h": () => this.editor.commands.toggleHighlight(),
-  //   };
-  // },
+  addKeyboardShortcuts() {
+    return {
+      [tiptapKeys.toggleHighlight.keys[0]]: () => this.editor.commands.toggleHighlight(getCurrentHighlightColor()),
+      [tiptapKeys.toggleHighlight.keys[1]]: () => this.editor.commands.toggleHighlight(getCurrentHighlightColor())
```
thecodrr: Remove this.

docs/help/contents/keyboard-shortcuts.md
thecodrr, line 1: What's up with the formatting changes?

packages/common/src/utils/keybindings.ts
```diff
@@ -388,6 +388,12 @@ export const tiptapKeys = {
     description: "Underline",
     category: "Editor",
     type: "tiptap"
+  },
+  toggleHighlight: {
+    keys: ["Mod-Alt-h"],
```
thecodrr: This should just be a string.

packages/editor/src/extensions/highlight/highlight.ts
```diff
@@ -99,11 +101,16 @@ export const Highlight = Extension.create<HighlightOptions>({
...
+  },
 
-  // addKeyboardShortcuts() {
-  //   return {
-  //     "Mod-Shift-h": () => this.editor.commands.toggleHighlight(),
-  //   };
-  // },
+  addKeyboardShortcuts() {
+    return {
+      [tiptapKeys.toggleHighlight.keys[0]]: () => this.editor.commands.toggleHighlight(getCurrentHighlightColor()),
```
thecodrr:
```suggestion
      [tiptapKeys.toggleHighlight.keys]: () => this.editor.commands.toggleHighlight(getCurrentHighlightColor()),
```

docs/help/contents/keyboard-shortcuts.md
Sudlon: No idea. After running the script, it was like this, so I left it in

thecodrr (approved):

Issue #8582, opened 2025-09-17
Exporting collapses whitespace


bingud:
### What happened?

Exporting or copying a note collapses whitespace, and exporting a note as text changes tabs into spaces.

### Steps to reproduce the problem

Copy or export a note

### Version

3.2.4-75c677c-web

### Platform/OS

Chrome/Chromium

### Relevant log output

```shell

```

bingud:
@thecodrr It looks like this preserves leading whitespace, but still collapses other whitespace.  
Is this intentional?

PR #8603, opened 2025-09-22
core: fix leading spaces in txt export && remove unneeded space at start of txt

01zulfi:
Closes https://github.com/streetwriters/notesnook/issues/8582

### DIFF
--- a/packages/core/src/content-types/__tests__/__snapshots__/tiptap.test.js.snap
+++ b/packages/core/src/content-types/__tests__/__snapshots__/tiptap.test.js.snap
@@ -126,6 +126,16 @@ But this post was supposed to be about the "Secrets" of Typescript Generics, rig
 "
 `;
 
+exports[`convert HTML to markdown with leadingWhitespaces > html-to-md-leadingWhitespaces.md 1`] = `
+"two spaces
+
+four spaces  
+two spaces  
+four spaces
+
+"
+`;
+
 exports[`convert HTML to markdown with outlinelists > html-to-md-outlinelists.md 1`] = `
 "Testing outline list:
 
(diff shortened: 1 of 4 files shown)


packages/core/src/content-types/__tests__/__snapshots__/tiptap.test.js.snap
```diff
@@ -126,6 +126,16 @@ But this post was supposed to be about the "Secrets" of Typescript Generics, rig
...
+exports[`convert HTML to markdown with leadingWhitespaces > html-to-md-leadingWhitespaces.md 1`] = `
+"two spaces
+
+four spaces  
+two spaces  
+four spaces
+
+"
+`;
+
```
01zulfi, line 138: haven't figured out a fix for markdown export, it can be done in a later pull request

packages/core/src/content-types/tiptap.ts
```diff
@@ -423,6 +423,18 @@ function convertHtmlToTxt(html: string, wrap = true) {
   });
 }
 
+function preserveParagraphLeadingWhitespaces(html: string) {
+  return html.replace(/<p([^>]*)>([ \t]*)/g, (match, attrs, leadingSpaces) => {
```
thecodrr: This is _not_ a good idea. We have 2 options here:

1. Use a proper HTML parser to do this
2. Do this inside `html-to-txt`'s `paragraph` formatter above. (preferred)

thecodrr (changes requested):

packages/core/src/content-types/tiptap.ts
01zulfi: gone with option 2

Issue #8597, opened 2025-09-21
How can I prevent Notesnook from converting my text into Markdown?


kate0037:
Every time I copy text that starts with `*`, Notesnook converts it into Markdown. How can I disable this feature? I have already turned off the `markdown shortcuts` in the settings.

For example, if I copy this text:

 `*   test`
`    *   some text`
 `    *  other text`

the result is:

<img width="410" height="170" alt="Image" src="https://github.com/user-attachments/assets/30354076-6ad2-4508-9297-6f1d3c8ea0b0" />

If I change some content, I can't copy it back into the original .txt file since the `*` is missing. This is very frustrating since a lot of my files are written in this style.

01zulfi:
Does pasting with `ctrl shift v` help your usecase?

kate0037:
> Does pasting with `ctrl shift v` help your usecase?

It works, thank you!
If I right-click the mouse and select `paste without formatting`, it doesn't work. I wonder what the reason is since Ctrl + Shift + V is its shortcut.

01zulfi:
> If I right-click the mouse and select paste without formatting, it doesn't work.

I haven't figured out why this doesn't work, well at least the `ctrl+shift+v` is working.

Also, I've raised a pull request to not parse markdown from the clipboard at all if the user has "markdown shortcuts" disabled (can't guarantee if it gets merged) https://github.com/streetwriters/notesnook/pull/8604

XP1:
There is a "Copy as" submenu when you right-click the note:

<img alt="Notesnook - Copy as markdown" src="https://github.com/user-attachments/assets/18ad4f07-4ca2-4b8e-8535-c0de26206684"/>

I would like to see this added to the selection right-click context menu. Please support my feature request:
[Copy as plain text / markdown in selection right-click context menu #9015](https://github.com/streetwriters/notesnook/issues/9015)

PR #8604, opened 2025-09-23
editor: skip markdown clipboard parsing if markdown shortcuts is disabled

01zulfi:
Closes https://github.com/streetwriters/notesnook/issues/8597

### DIFF
--- a/packages/editor/src/extensions/clipboard/clipboard-text-parser.ts
+++ b/packages/editor/src/extensions/clipboard/clipboard-text-parser.ts
@@ -27,9 +27,10 @@ export function clipboardTextParser(
   text: string,
   $context: ResolvedPos,
   plain: boolean,
-  view: EditorView
+  view: EditorView,
+  parseMarkdownOnPaste = true
 ): Slice {
-  if (!plain && isProbablyMarkdown(text)) {
+  if (!plain && parseMarkdownOnPaste && isProbablyMarkdown(text)) {
     const node = ClipboardDOMParser.fromSchema(view.state.schema).parse(
       new DOMParser().parseFromString(
         markdowntoHTML(text, { allowDangerousHtml: false }),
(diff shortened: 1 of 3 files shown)


packages/editor/src/extensions/clipboard/clipboard-text-parser.ts
```diff
@@ -27,9 +27,10 @@ export function clipboardTextParser(
   text: string,
   $context: ResolvedPos,
   plain: boolean,
-  view: EditorView
+  view: EditorView,
+  parseMarkdownOnPaste = true
 ): Slice {
-  if (!plain && isProbablyMarkdown(text)) {
+  if (!plain && parseMarkdownOnPaste && isProbablyMarkdown(text)) {
```
01zulfi, line 33: **unrelated comment**: `plain` is `true` when content is pasted via `ctrl+shift+v` but `false` if pasted view `right click > paste as plain text`. I haven't found any answer for this yet

thecodrr: `plain` is set by `prosemirror` I think. Might be worth investigating.

thecodrr:
The changes look okay but I am not sure if we should mix the two. It's possible that a user has disabled markdown shortcuts but still wants to paste markdown. If we mix the two it'll annoy users.

01zulfi:
> It's possible that a user has disabled markdown shortcuts but still wants to paste markdown

I would agree, but markdown shortcuts was recently upgraded to premium.

thecodrr:
> I would agree, but markdown shortcuts was recently upgraded to premium.

That doesn't change what I said though. Paid users could still want to disable markdown but keep pasting markdown.

kate0037:
> > I would agree, but markdown shortcuts was recently upgraded to premium.
> 
> That doesn't change what I said though. Paid users could still want to disable markdown but keep pasting markdown.

Hi! I would like to suggest that it would be much better if you could provide an option in the settings for users to choose whether to turn this feature on or off. I believe Notesnook offers more features than its competitors, but it can be frustrating when a feature I don't need is enabled by default and cannot be disabled. For example, Notesnook shows a preview for notes, but there is no way to turn it off.

PR #8623, opened 2025-10-01
web: fix text in upgrade dialog

01zulfi:


apps/web/src/dialogs/buy-dialog/upgrade-dialog.tsx
```diff
@@ -97,7 +97,7 @@ export const UpgradeDialog = DialogManager.register(function UpgradeDialog(
           {[
             getFeature("storage"),
             getFeature("fileSize"),
-            getFeature(feature.id),
```
thecodrr, line 100: We need to show the limited feature here but not if it's already in the list.

thecodrr (changes requested):

apps/web/src/dialogs/buy-dialog/upgrade-dialog.tsx
01zulfi: The limited feature already gets highlighted 
image


Or do you mean the limited feature should be in the centre of the list?

thecodrr: That's the case for tags and other items in the default list. Sometimes a user accesses a feature that's not in the default list. We have to show that too.

thecodrr (approved):

PR #8733, opened 2025-10-08
core: handle inbox item sync & decryption

01zulfi:

### DIFF
--- a/packages/core/src/api/sync/index.ts
+++ b/packages/core/src/api/sync/index.ts
@@ -49,6 +49,7 @@ import {
 import {
   SYNC_COLLECTIONS_MAP,
   SyncableItemType,
+  SyncInboxItem,
   SyncTransferItem
 } from "./types.js";
 import { DownloadableFile } from "../../database/fs.js";
@@ -233,7 +234,7 @@ class Sync {
   async fetch(deviceId: string, options: SyncOptions) {
     await this.checkConnection();
 
-    await this.connection?.invoke("RequestFetchV2", deviceId);
+    await this.connection?.invoke("RequestFetchV3", deviceId);
 
     if (this.conflictedNoteIds.length > 0) {
       await this.db
@@ -478,6 +479,19 @@ class Sync {
 
       return true;
     });
+
+    this.connection.on(
+      "SendInboxItems",
+      async (inboxItems: SyncInboxItem[]) => {
+        if (this.connection?.state !== HubConnectionState.Connected) {
+          return false;
+        }
+
+        await this.handleInboxItems(inboxItems);
+
+        return true;
+      }
+    );
   }
 
   private async getKey() {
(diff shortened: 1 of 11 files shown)


packages/core/src/api/sync/index.ts
```diff
@@ -233,7 +234,7 @@ class Sync {
   async fetch(deviceId: string, options: SyncOptions) {
     await this.checkConnection();
 
-    await this.connection?.invoke("RequestFetchV2", deviceId);
+    await this.connection?.invoke("RequestFetchV3", deviceId);
```
thecodrr: I think we should have some kind of fallback logic here. It's possible that someone self hosting Notesnook doesn't yet have the latest server but he is using the latest clients. We should handle that case.

packages/core/src/api/sync/index.ts
```diff
@@ -516,6 +530,98 @@ class Sync {
       }
     });
   }
+
+  private async handleInboxItems(inboxItems: SyncInboxItem[]) {
```
thecodrr: Let's move this to `merger.ts`

packages/core/src/api/sync/index.ts
```diff
@@ -516,6 +530,98 @@ class Sync {
...
+      try {
+        if (await this.db.notes.exists(item.id)) {
+          this.logger.info("Inbox item already exists, skipping.", {
+            inboxItemId: item.id
+          });
+          continue;
+        }
+
+        const inboxKeys = await this.db.user.getInboxKeys();
+        if (!inboxKeys) continue;
```
thecodrr: Do this once outside the loop. I know we have caching but still no point in calling it on each iteration.

packages/core/src/api/sync/index.ts
```diff
@@ -516,6 +530,98 @@ class Sync {
...
+
+        const decryptedKey = await this.db.storage().decryptAsymmetric(
+          {
+            publicKey: inboxKeys.publicKey,
+            privateKey: inboxKeys.privateKey
+          },
```
thecodrr: Why not just pass `inboxKeys` instead of destructuring here like this?

packages/core/src/api/sync/index.ts
```diff
@@ -516,6 +530,98 @@ class Sync {
...
+          {
+            alg: item.key.alg,
+            iv: "",
```
thecodrr: Create a new type to exclude these fields.

packages/core/src/api/sync/index.ts
```diff
@@ -516,6 +530,98 @@ class Sync {
...
+        const decryptedItem = await this.db.storage().decrypt(
+          { key: decryptedKey },
+          {
+            alg: item.alg,
+            iv: item.iv,
+            cipher: item.cipher,
+            format: "base64",
+            length: item.length,
+            salt: ""
+          }
```
thecodrr: Why not pass the item as-is?

packages/core/src/api/sync/index.ts
```diff
@@ -516,6 +530,98 @@ class Sync {
...
+          notebookIds?: string[];
+          tagIds?: string[];
+          type: "note";
+          source: string;
+          version: 1;
+          content?: {
+            type: "html";
+            data: string;
+          };
+        };
```
thecodrr: Instead of `as` create a new type for it.

packages/core/src/api/sync/index.ts
01zulfi: `salt` is not present in `SyncInboxItem`

packages/core/src/api/sync/merger.ts
```diff
@@ -146,3 +147,78 @@ export function isContentConflicted(
...
+      });
+      const decryptedItem = await db.storage().decrypt(
+        { key: decryptedKey },
+        {
+          alg: item.alg,
+          iv: item.iv,
+          cipher: item.cipher,
+          format: "base64",
+          length: item.length,
+          salt: ""
```
thecodrr: Use the salt from the key then.

01zulfi: The inbox keys? They don't have a salt. Or, do you mean the salt from user's database encryption key?

thecodrr: No from the `decryptedKey`

01zulfi: `decryptedKey` is just a string, I'm not sure how to derive salt from there

thecodrr: How are you decrypting without the salt? We need the salt.

01zulfi: `salt` is not passed to the decryption functions at all, that's why I'm getting away with just passing an empty string here

thecodrr: Ah I see, you modified `KeyUtils.transform`. I don't think that's a good idea. We will need to include the salt from the `inbox-api-server` so it can be used here.

01zulfi: PR to send salt https://github.com/streetwriters/notesnook-sync-server/pull/57

> Ah I see, you modified KeyUtils.transform. I don't think that's a good idea.

This is done to accommodate the absence of salt in the key, not sure how to handle this.

packages/crypto/src/keyutils.ts
```diff
@@ -92,6 +92,11 @@ export default class KeyUtils {
       return this.deriveKey(sodium, password, salt);
     } else if ("key" in input && !!input.salt && !!input.key) {
       return { key: sodium.from_base64(input.key), salt: input.salt };
+    } else if ("key" in input && !!input.key) {
+      return {
+        key: sodium.from_base64(input.key, base64_variants.URLSAFE_NO_PADDING),
+        salt: ""
+      };
```
thecodrr: This is no longer necessary.

PR #8743, opened 2025-10-09
editor: restrict attachment upload for not logged-in users

01zulfi:


apps/web/src/components/editor/tiptap.tsx
```diff
@@ -410,7 +410,8 @@ function TipTap(props: TipTapProps) {
             openInNewTab: openInNewTab
           });
         } else window.open(url, "_blank");
-      }
+      },
+      isLoggedIn: () => useUserStore.getState().isLoggedIn
```
thecodrr: I don't think we should expose this to the editor. It feels like a breach of scope. How about if we just show the error message if user tries to attach an image etc? Or if we do want to hide menu items, a better implementation would be to use the permission handler.

01zulfi: Permission handler was a nice suggestion, thanks

Also, I'm not hiding the menu items anymore, just showing the dialog whenever the user tries to insert an image/attachment.

thecodrr (approved):

PR #8857, opened 2025-10-27
core: escape paranthesis in notes search query

01zulfi:

### DIFF
--- a/packages/core/__tests__/lookup.test.js
+++ b/packages/core/__tests__/lookup.test.js
@@ -141,7 +141,7 @@ test("search reminders", () =>
   }));
 
 describe("notesWithHighlighting", () => {
-  test.only("search notes with parentheses in query should load the item", () =>
+  test("search notes with parentheses in query should load the item", () =>
     noteTest({
       title: "(with parantheses)"
     }).then(async ({ db }) => {


packages/core/src/api/lookup.ts
```diff
@@ -1154,7 +1154,8 @@ function textContainsTokens(text: string, tokens: QueryTokens) {
   const lowerCasedText = text.toLowerCase();
 
   const createTagPattern = (token: string) => {
-    return `<${MATCH_TAG_NAME}\\s+id="(.+?)">${token}<\\/${MATCH_TAG_NAME}>`;
+    const escapedToken = token.replace(/[()]/g, "\\$&");
+    return `<${MATCH_TAG_NAME}\\s+id="(.+?)">${escapedToken}<\\/${MATCH_TAG_NAME}>`;
```
thecodrr, line 1158: We are already escaping tokens in `query-transformer`. Why is that not working?

01zulfi: I found this in `query-transformer.ts` https://github.com/streetwriters/notesnook/blob/master/packages/core/src/utils/query-transformer.ts#L133

I think we need to escape the `(` or `)` characters by doing this `.replace(/[()]/g, "\\$&")` to prevent regex errors

thecodrr: Looking at it again, I think your change is more appropriate.

thecodrr:
@01zulfi please add tests for this

PR #8872, opened 2025-10-28
web: enable inbox && add inbox api docs

01zulfi:

Linked: [NN-5]


apps/web/src/components/notebook/index.tsx
```diff
@@ -286,6 +289,23 @@ export const notebookMenuItems: (
...
       multiSelect: true
+    },
+    {
+      type: "button",
+      key: "copyid",
+      title: "Copy ID",
+      icon: Copy.path,
+      onClick: async () => {
+        try {
+          await navigator.clipboard.writeText(notebook.id);
```
thecodrr: Use the `writeToClipboard` utility.

apps/web/src/components/notebook/index.tsx
```diff
@@ -286,6 +289,23 @@ export const notebookMenuItems: (
...
+          showToast("success", "Notebook ID copied to clipboard");
+        } catch (e) {
+          console.error(e);
+          showToast("error", "Failed to copy Notebook ID");
+        }
+      },
+      isHidden: !settingStore.get().isInboxEnabled,
+      multiSelect: false
```
thecodrr: No need for this.

docs/help/contents/inbox-api/inbox-api.md
```diff
@@ -0,0 +1,118 @@
...
+
+One API key should be created by default. You can create multiple API keys if needed.
+
+# [Desktop/Web](#/tab/web)
+
+`Settings > Inbox > Create Key`.
+
+---
+
+### 3. Post data to Inbox API endpoint.
```
thecodrr: The `json` plus table seems redundant. A good example of how API documentation should look like is this: https://developer.paddle.com/api-reference/products/create-product

We should also add an example that can be run via `curl`. Another thing we need to add is how to send pre-encrypted data directly to the Notesnook server without going through the Inbox API server. Perhaps we should also add an example doing it using OpenSSL + cURL.

thecodrr (changes requested):

apps/web/src/dialogs/settings/index.tsx
```diff
@@ -120,8 +120,7 @@ const sectionGroups: SectionGroup[] = [
       {
         key: "inbox",
         title: "Inbox",
-        icon: Inbox,
-        isHidden: () => true // hidden until complete
+        icon: Inbox
```
thecodrr: Inbox should only be visible if user is logged in.

thecodrr (changes requested):

docs/help/contents/inbox-api/inbox-api.md
01zulfi: > A good example of how API documentation should look like is this: https://developer.paddle.com/api-reference/products/create-product

I've improved the current doc, not sure which specific changes you meant.

> Another thing we need to add is how to send pre-encrypted data directly to the Notesnook server without going through the Inbox API server. Perhaps we should also add an example doing it using OpenSSL + cURL.

This would require us to explain how the data is encrypted, the docs for which are present in the other repo. Maybe we can say something like "You can skip the inbox server altogether and send encrypted data directly to the Notesnook server, learn more here (link to the inbox server repo)"

thecodrr (approved):

PR #8877, opened 2025-10-29
editor: fix images from older version not rendering in outline list

01zulfi:


thecodrr (changes requested):
Can you add tests for this?

thecodrr (approved):

PR #8882, opened 2025-10-30
editor: fix disappearing nodes under collapsed headings if heading is edited

01zulfi:
* design flaw in using the hiddenUnder attribute on hidden nodes, a hidden attribute should be sufficient

### DIFF
--- a/packages/editor/src/extensions/heading/heading.ts
+++ b/packages/editor/src/extensions/heading/heading.ts
@@ -89,14 +89,14 @@ export const Heading = TiptapHeading.extend({
       {
         types: COLLAPSIBLE_BLOCK_TYPES,
         attributes: {
-          hiddenUnder: {
-            default: null,
+          hidden: {
+            default: false,
             keepOnSplit: false,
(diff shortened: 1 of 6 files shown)


packages/editor/src/extensions/heading/heading.ts
```diff
@@ -289,15 +283,33 @@ function toggleNodesUnderHeading(
...
-      } else if (!isCollapsing && nextNode.attrs.hiddenUnder === headingId) {
-        tr.setNodeAttribute(nextPos, "hiddenUnder", null);
+      if (isCollapsing) {
+        tr.setNodeAttribute(currentPos, "hidden", true);
+      } else {
+        if (insideCollapsedHeading) {
+          if (
+            nextNode.type.name === "heading" &&
+            // @ts-expect-error 'nestedHeadingLevel' is possibly 'null'
+            nextNode.attrs.level <= nestedHeadingLevel
```
thecodrr: Why not add a check instead of `ts-expect-error`?

PR #8889, opened 2025-10-31
editor: fix changing level of collapsed heading

01zulfi:
* preserve collasped state
* reset visibility of hidden nodes under changed heading

### DIFF
--- a/packages/editor/src/extensions/heading/heading.ts
+++ b/packages/editor/src/extensions/heading/heading.ts
@@ -24,7 +24,7 @@ import {
 } from "@tiptap/core";
 import { Heading as TiptapHeading } from "@tiptap/extension-heading";
 import { isClickWithinBounds } from "../../utils/prosemirror.js";
-import { Selection, Transaction } from "@tiptap/pm/state";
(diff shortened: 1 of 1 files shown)


packages/editor/src/extensions/heading/heading.ts
```diff
@@ -344,3 +349,43 @@ function findEndOfCollapsedSection(
...
 }
+
+const headingUpdatePlugin = new Plugin({
+  key: new PluginKey("headingUpdate"),
+  appendTransaction(transactions, oldState, newState) {
+    const tr = newState.tr;
+    let modified = false;
+
+    transactions.forEach((transaction) => {
+      if (!transaction.docChanged) return;
```
thecodrr: Why are we iterating on all the transactions?

01zulfi: for `transaction.docChanged` check, or is that not required?

thecodrr: You can check that outside i.e. if any transaction has docChanged only then continue. No need to do all the logic inside the loop.

thecodrr (approved):

PR #8905, opened 2025-11-03
editor: fix old inline images in bullet,numbered,task&check lists

01zulfi:

### DIFF
--- a/packages/editor/src/extensions/task-item/task-item.ts
+++ b/packages/editor/src/extensions/task-item/task-item.ts
@@ -21,6 +21,7 @@ import { mergeAttributes } from "@tiptap/core";
 import { TaskItem } from "@tiptap/extension-task-item";
 import { TaskItemComponent } from "./component.js";
 import { createNodeView } from "../react/index.js";
+import { ensureLeadingParagraph } from "../../utils/prosemirror.js";
 
 export type TaskItemAttributes = {
   checked: boolean;
@@ -56,7 +57,8 @@ export const TaskItemNode = TaskItem.extend({
     return [
       {
         tag: ".checklist > li",
-        priority: 51
+        priority: 100,
+        getContent: ensureLeadingParagraph
       }
     ];
   },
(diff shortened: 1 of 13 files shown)


packages/editor/src/extensions/task-item/task-item.ts
```diff
@@ -56,7 +57,18 @@ export const TaskItemNode = TaskItem.extend({
     return [
       {
         tag: ".checklist > li",
-        priority: 51
+        priority: 100,
+        getContent: (node, schema) => {
+          const parser = DOMParser.fromSchema(schema);
+          const fragment = parser.parse(node).content;
```
thecodrr: Instead of doing this for each type of list item separately, can we just do it for all list items once? I don't think there's any difference in logic.

01zulfi: Can you guide on where to write this logic if we have to write only once?

thecodrr: We can do it in `list-item.ts`

01zulfi: I removed the `getContent` from task list and outline list and only kept it in list item. This is what I observed:

1. The old inline images in bullet/numbered list migrated successfully
2. The image disappeared from outline list
3. The image did appear in the task list but the task list item containing the image got merged into the task list item above it. (I think this might be an issue with `priority` in the parseHTML of task items though)

Investigating further...

thecodrr: Can you share the code?

01zulfi: I just removed `getContent` from `outline-list-item.ts` and `task-item.ts` and kept it in `list-item.ts`

```ts
 getContent: (node, schema) => {
          const parser = DOMParser.fromSchema(schema);
          const fragment = parser.parse(node).content;
          const firstNode = fragment.firstChild;
          if (firstNode && firstNode.type.name !== "paragraph") {
            const emptyParagraph = schema.nodes.paragraph.create();
            return fragment.addToStart(emptyParagraph);
          }

          return fragment;
  }
        ```

thecodrr: Just tested and this won't work. Let's just reuse the `getContent` function across extensions.

thecodrr: You might also want to do this for `check-list-item`

PR #8918, opened 2025-11-05
core: fix user keys re-encryption when updating password

01zulfi:

### DIFF
--- a/packages/core/src/api/user-manager.ts
+++ b/packages/core/src/api/user-manager.ts
@@ -666,6 +666,15 @@ class UserManager {
         usesFallback: await this.usesFallbackPWHash(old_password)
       });
 
+    // retrieve user keys before deriving a new encryption key
+    const oldUserKeys = {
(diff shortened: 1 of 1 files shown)


packages/core/src/api/user-manager.ts
```diff
@@ -666,6 +666,15 @@ class UserManager {
         usesFallback: await this.usesFallbackPWHash(old_password)
       });
 
+    // retrieve user keys before deriving a new encryption key
+    const oldUserKeys = {
+      attachmentsKey: await this.getAttachmentsKey(),
+      monographPasswordsKey: await this.getMonographPasswordsKey(),
+      inboxKeys: (await this.hasInboxKeys())
+        ? await this.getInboxKeys()
+        : undefined
```
thecodrr, line 675: This'll create those keys if they don't exist for a user (which is not a huge deal but just something to be aware of).

PR #8919, opened 2025-11-05
web: show monograph's view count

01zulfi:
image

### DIFF
--- a/apps/web/src/components/publish-view/index.tsx
+++ b/apps/web/src/components/publish-view/index.tsx
@@ -122,8 +122,7 @@ function PublishView(props: PublishViewProps) {
         sx={{
           flexDirection: "column",
           border: "1px solid var(--border)",
-          borderRadius: "default",
-          "& div:not(:last-child)": { borderBottom: "1px solid var(--border)" }
+          borderRadius: "default"
         }}
       >
         {monograph?.publishedAt ? (


apps/web/src/components/publish-view/index.tsx
```diff
@@ -140,12 +142,38 @@ function PublishView(props: PublishViewProps) {
...
+                        px: 1,
+                        py: 0.7,
+                        borderRadius: "default",
+                        fontSize: "subBody",
+                        fontWeight: "bold",
+                        border: "1px solid",
+                        borderColor: "accent"
+                      }}
+                    >
+                      {viewCount} {viewCount === 1 ? "view" : "views"}
```
thecodrr: Better to use `strings` here from `intl`.

packages/core/src/api/monographs.ts
```diff
@@ -187,4 +190,17 @@ export class Monographs {
     if (!monographPasswordsKey) return "";
     return this.db.storage().decrypt(monographPasswordsKey, password);
   }
+
+  async getStats(monographId: string) {
```
thecodrr:
```suggestion
  async stats(monographId: string) {
```

apps/web/src/components/publish-view/index.tsx
```diff
@@ -45,6 +45,7 @@ function PublishView(props: PublishViewProps) {
   );
   const [isPasswordProtected, setIsPasswordProtected] = useState(false);
   const [selfDestruct, setSelfDestruct] = useState(false);
+  const [viewCount, setViewCount] = useState<number | undefined>();
```
thecodrr: Better to create a `stats` state instead of individual `viewCount` state.

PR #8953, opened 2025-11-13
editor: enable heading in table, change collapse icon pos, && disable empty heading collapse

01zulfi:
image

### DIFF
--- a/packages/editor/src/extensions/heading/heading.ts
+++ b/packages/editor/src/extensions/heading/heading.ts
@@ -214,7 +214,7 @@ export const Heading = TiptapHeading.extend({
       }
 
       icon.onmousedown = onIconClick;
-      icon.ontouchstart = onIconClick;
+      icon.ontouchend = onIconClick;
 
       heading.appendChild(contentWrapper);
       heading.appendChild(icon);


packages/editor/src/extensions/heading/heading.ts
thecodrr, line 94: What if we move the empty heading migration here? It will run only once.

packages/editor/src/extensions/heading/heading.ts
```diff
@@ -394,3 +393,88 @@ const headingUpdatePlugin = new Plugin({
     return modified ? tr : null;
   }
 });
+
+function migrateEmptyCollapsedHeading(element: HTMLElement) {
```
thecodrr: I get the first migration but what are the other 2 loops for?

01zulfi: Here are the steps.

1. Loop backwards to find the heading we need to migrate. Migrate the heading by removing `data-collapsed`
2. Loop backwards again to find a higher level collapsed heading before the migrated heading. If we find one, there's no need to migrate the hidden nodes, since they should stay hidden even if the migrated heading is not collapsed anymore
3. If we don't find a collapsed higher level previous heading, we loop forwards from the migrated heading and remove `data-hidden` from relevant nodes

I'm still trying to figure out a better way of doing this

thecodrr: Let's just skip it. If anyone faces this issue, they can write into the heading while will make the arrow appear. This migration will be buggy and slow and quite pointless as most headings will have content.

01zulfi (commented):
the PR makes the collapse icon implementation same for both mobile&web, we should manually test on mobile as well

PR #8972, opened 2025-11-17
web: fix text not selectable from entire height of editor

### LINKED ISSUE #8498: Expand focus area for text entry all the way to the bottom of the program window

Sorry, it's a bit hard to describe this feature request....

I have noticed that Notesnook behaves a bit different then other notes apps or text editors when it comes to setting the focus for text entry for a note. You have to click either into the existing text or within a few lines below the text otherwise, Notesnook will receive the focus but you can't enter text.

I hope this explains it:

Image

Please note that the red area is even bigger when you are just starting a note and there is barely any text above.

Even worse: When you DO HAVE the focus on text entry and you want to add something to the end of your note, it can happen that you click too far below the text which then EXITS text entry mode.

Version: 3.2.3-dd42b82-desktop
Type: Flatpak

I would prefer it if the area for text entry is expanded all the way to the bottom.

Yes, this may seem like such a minor inconvenience HOWEVER currently Notesnook is **a bit tedious to use on a Laptop with a touch pad**. When switching apps from the task manager at the bottom of the screen, I constantly click into the read area and wonder why Notesnook doesn't accept my text input. This means when switching to Notesnook and the focus isn't on text entry, I have to move the mouse cursor halfway up the screen and click into the Note to be able to enter text.

The behavior is also different from any note taking app or text editor that I know. For example LibreOffice, MS Word, Kate, etc all accept text input even if you click into the bottom of the program.

_No response_

### Additional context

_No response_

keysmashes:
fixes #8498

### DIFF
--- a/apps/web/src/components/editor/tiptap.tsx
+++ b/apps/web/src/components/editor/tiptap.tsx
@@ -620,7 +620,7 @@ function TiptapWrapper(
       sx={{
         flex: 1,
         flexDirection: "column",
-        ".tiptap.ProseMirror": { pb: 150 },
+        ".tiptap.ProseMirror": { height: "100%" },
         ".editor-container": {
           opacity: isHydrating ? 0 : 1,
           zoom: editorConfig.zoom + "%"


01zulfi (approved):
looks good to me

keysmashes:
I don't suppose this could get merged if it's an acceptable change?

PR #8980, opened 2025-11-19
editor: fade checklist when checked

01zulfi:
# before

image

# after

image

### DIFF
--- a/packages/editor/styles/styles.css
+++ b/packages/editor/styles/styles.css
@@ -707,6 +707,10 @@ p > *::selection {
   mask-size: cover;
 }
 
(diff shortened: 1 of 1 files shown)


InsaneEagle-dot:
This is really good. It will be insane improvement in checklists. Please add this if possible 🙏

packages/editor/styles/styles.css
```diff
@@ -707,6 +707,10 @@ p > *::selection {
   mask-size: cover;
 }
 
+.simple-checklist > li.checked {
```
thecodrr:
```suggestion
.simple-checklist > li.checked p {
```

PR #8987, opened 2025-11-21
editor: improve inline code styling

01zulfi:
# before

image

# after

image

### DIFF
--- a/packages/editor/styles/styles.css
+++ b/packages/editor/styles/styles.css
@@ -54,10 +54,10 @@
   border: 1px solid var(--border);
   color: var(--paragraph-secondary);
   border-radius: 5px;
-  padding: 0px 5px 0px 5px;
+  padding: 2px 5px 2px 5px;
   font-family: ui-monospace, SFMono-Regular, SF Mono, Consolas, Liberation Mono,
     Menlo, monospace !important;
-  font-size: 10pt !important;
+  font-size: 85%;
 }
 
 .ProseMirror code > span {


packages/editor/styles/styles.css
```diff
@@ -54,10 +50,11 @@
 }
 
 .ProseMirror p code {
-  background-color: var(--background);
+  background-color: var(--background-secondary);
   border: 1px solid var(--border);
+  color: var(--accent);
```
thecodrr: Not sure about this. I think it should be `paragraph-secondary` and then people can theme it however they want.

thecodrr (changes requested):

packages/editor/styles/styles.css
01zulfi: I think we should keep `accent`. All examples I've seen so far on Discord thread has the inline code color be an accent (different to the text color).

thecodrr: Yes, but that depends on the theme. We are already using `paragraph-secondary` for inline & block math nodes so it'd work well. Theme developers can then change the color to whatever they think fits best with `background-secondary`. It's also possible (not likely though) that the `accent` color isn't legible with `background-secondary`.

01zulfi: Oh I see, so the theme itself can be updated. But aren't secondary colors supposed to a slightly different shade of the primary color? It still won't be a completely different color, something that would we would like in inline code. 

For now I've updated the color to `paragraph-secondary`, but am still willing to make an exception for inline code and set it to `accent`

thecodrr: It's funny because GitHub (the most popular platform used by developers) uses very plain inline code elements 😆 and I don't think anyone has any issues with that. I think a few people want things to look a certain way while everyone else couldn't care less. 

> But aren't secondary colors supposed to a slightly different shade of the primary color?

That's up to the theme developer.

thecodrr: We have to be careful when making aesthetic changes because most users of Notesnook are not active on Discord or any other platform and cannot voice their opinions. Changing things (aesthetically) because one or two users asked for them is a bad idea. For example, we changed the way check lists worked based on very minimal feedback and it caused a huge backlash.

packages/editor/styles/styles.css
```diff
@@ -54,10 +50,11 @@
 }
 
 .ProseMirror p code {
-  background-color: var(--background);
+  background-color: var(--background-secondary);
   border: 1px solid var(--border);
```
thecodrr, line 54: Can you try without the border? I think it'll look better.

01zulfi: without border

image

packages/editor/styles/styles.css
01zulfi: Got it, that makes sense

packages/editor/styles/styles.css
thecodrr: I think there should be very minimal vertical paddings. The code looks too close to the edges now.

thecodrr: Or we could change the font size but that'd look odd.

PR #8995, opened 2025-11-25
core: support monograph.publishUrl field

01zulfi:

### DIFF
--- a/packages/core/src/api/monographs.ts
+++ b/packages/core/src/api/monographs.ts
@@ -133,10 +133,13 @@ export class Monographs {
           })
     };
 
-    const method = update ? http.patch.json : http.post.json;
     const deviceId = await this.db.kv().read("deviceId");
+    const method = update ? http.patch.json : http.post.json;
+    const url = update
+      ? `${Constants.API_HOST}/monographs?deviceId=${deviceId}`
+      : `${Constants.API_HOST}/monographs/v2?deviceId=${deviceId}`;
     const { id, datePublished, publishUrl } = await method(
-      `${Constants.API_HOST}/monographs/v2?deviceId=${deviceId}`,
+      url,
       monograph,
       token
     );


packages/core/src/types.ts
```diff
@@ -498,6 +498,7 @@ export interface Monograph extends BaseItem<"monograph"> {
   datePublished: number;
   selfDestruct: boolean;
   password?: Cipher<"base64">;
+  slug?: string;
```
thecodrr: Instead of having `slug` here, we should instead have `publishUrl` that contains the full monograph url including the domain etc. The `publishUrl` will be passed down by the server, of course.

01zulfi: got it, though will the server know about `MONOGRAPH_HOST`, will it be we send this value to the server when publishing/updating?

thecodrr: We can set a new environment variable for MONOGRAPH_HOST on the server.

packages/core/src/database/migrations.ts
```diff
@@ -425,6 +425,14 @@ export class NNMigrationProvider implements MigrationProvider {
             .addColumn("title", "text")
             .execute();
         }
+      },
+      "a-2025-11-26": {
```
thecodrr:
```suggestion
      "a-2026-02-11": {
```

apps/web/src/components/publish-view/index.tsx
```diff
@@ -96,23 +99,27 @@ function PublishView(props: PublishViewProps) {
...
               whiteSpace: "nowrap",
               textDecoration: "none",
               overflow: "hidden",
-              px: 1
+              px: 1,
+              opacity: publishUrl.status === "fulfilled" ? 1 : 0.8
             }}
           >
-            {`${hosts.MONOGRAPH_HOST}/${monograph?.id}`}
+            {publishUrl.status === "fulfilled" ? publishUrl.value : ""}
```
thecodrr: Shouldn't we fallback to `monograph.publishUrl`?

01zulfi: Nice suggestion, thanks

PR #8996, opened 2025-11-25
editor: fix heading align

01zulfi:

### DIFF
--- a/packages/editor/src/extensions/heading/heading.ts
+++ b/packages/editor/src/extensions/heading/heading.ts
@@ -23,8 +23,11 @@ import {
   textblockTypeInputRule
 } from "@tiptap/core";
 import { Heading as TiptapHeading } from "@tiptap/extension-heading";
+import { isClickWithinBounds } from "../../utils/prosemirror.js";
 import { Plugin, PluginKey, Selection, Transaction } from "@tiptap/pm/state";
 import { Node } from "@tiptap/pm/model";
+import { useToolbarStore } from "../../toolbar/stores/toolbar-store.js";
+import { Decoration, DecorationSet } from "prosemirror-view";
 
 const COLLAPSIBLE_BLOCK_TYPES = [
   "paragraph",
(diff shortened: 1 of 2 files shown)


packages/editor/styles/styles.css
```diff
@@ -903,6 +878,25 @@ del.diffdel {
...
   align-items: center;
+  justify-content: flex-start;
+}
+
+.ProseMirror h1[style*="text-align: center"],
+.ProseMirror h2[style*="text-align: center"],
+.ProseMirror h3[style*="text-align: center"],
+.ProseMirror h4[style*="text-align: center"],
+.ProseMirror h5[style*="text-align: center"],
+.ProseMirror h6[style*="text-align: center"] {
```
thecodrr: Why not use `class` instead?

thecodrr (changes requested):

packages/editor/styles/styles.css
01zulfi: the `setTextAlign` editor command sets the inline style `text-align`, that is what I'm using for the CSS selector

Not sure I understand how th use `class` instead?

thecodrr: Use the `textAlign` attribute in the `addNodeView` to set the class.

01zulfi: instead of a class, I opted to use a data attribute. With attribute I don't need to worry about removing other text align classes, so I think it is a bit cleaner. Let me know if `class` is still a better choice for this

packages/editor/src/extensions/heading/heading.ts
```diff
@@ -56,6 +56,19 @@ export const Heading = TiptapHeading.extend({
...
+        keepOnSplit: false,
+        parseHTML: (element) => element.dataset.textAlign,
+        renderHTML: (attributes) => {
+          if (attributes.textAlign) {
+            return {
+              "data-text-align": attributes.textAlign
+            };
+          }
+          return {};
+        }
```
thecodrr: This is unnecessary. `textAlign` would be present in the node `attrs` just like `textDirection`.

packages/editor/src/extensions/heading/heading.ts
```diff
@@ -238,11 +251,9 @@ export const Heading = TiptapHeading.extend({
             heading.dataset.hidden = updatedNode.attrs.hidden;
           else delete heading.dataset.hidden;
 
-          if (updatedNode.attrs.textAlign)
-            heading.style.textAlign =
-              updatedNode.attrs.textAlign === "left"
-                ? ""
-                : updatedNode.attrs.textAlign;
+          if (updatedNode.attrs.textAlign) {
+            heading.dataset.textAlign = updatedNode.attrs.textAlign;
```
thecodrr: Use class here.

packages/editor/src/extensions/heading/heading.ts
```diff
@@ -238,11 +239,13 @@ export const Heading = TiptapHeading.extend({
             heading.dataset.hidden = updatedNode.attrs.hidden;
           else delete heading.dataset.hidden;
 
-          if (updatedNode.attrs.textAlign)
+          if (updatedNode.attrs.textAlign) {
             heading.style.textAlign =
               updatedNode.attrs.textAlign === "left"
                 ? ""
                 : updatedNode.attrs.textAlign;
```
01zulfi, line 268: accidently removed this in the previous commit, added it back in. We need the inline `text-align` style to align the longer headings (which take up full editor width)

I was thinking we could've used the [style*="text-align: xyz"] CSS selector since we have to provide this style anyways, though a class works too

thecodrr: `[style*="text-align: xyz"]` that selector is very slow. We can't use it.

PR #8997, opened 2025-11-25
editor: readonly table fixes

01zulfi:
* fix tables not hiding under collapsed headings in readonly view
* allow column resizing in readonly view
* fix scrollbar not showing in readonly view


packages/editor/src/extensions/table/table.ts
```diff
@@ -477,19 +477,19 @@ export const Table = Node.create<TableOptions>({
...
+      // ...(isResizable
+      // [
+      columnResizing({
+        cellMinWidth: this.options.cellMinWidth,
+        View: TableNodeView(this.editor),
+        showResizeHandleOnSelection: this.options.showResizeHandleOnSelection,
+        isResizable: this.options.resizable
+      }),
+      // ],
+      // : [tiptapTableView(this.options.cellMinWidth)]),
```
01zulfi: I think we should allow column resizing in a readonly editor. It is similar action to collapse/expand headings and the state isn't preserved in the note's content.

Though I'm not sure about the significance of rendering the `TableView` in `tiptapTableView` when we can just render the `TableNodeView` and make some tweaks to handle to handle readonly case.

@thecodrr what do you think about this?

thecodrr: > It is similar action to collapse/expand headings and the state isn't preserved in the note's content.

It is preserved. Or do you mean that it's not preserved when in readonly mode?

packages/editor/src/extensions/table/table.ts
```diff
@@ -477,19 +477,19 @@ export const Table = Node.create<TableOptions>({
...
-              cellMinWidth: this.options.cellMinWidth,
-              View: TableNodeView(this.editor),
-              showResizeHandleOnSelection:
-                this.options.showResizeHandleOnSelection
-            })
-          ]
-        : [tiptapTableView(this.options.cellMinWidth)]),
+      // ...(isResizable
+      // [
+      columnResizing({
```
thecodrr: I don't get it. If `this.options.resizable` is `false` why do we need the `columnResizing` plugin?

packages/editor/src/extensions/table/table.ts
01zulfi: Yes I mean in readonly mode

PR #9001, opened 2025-11-26
core: archived notes not showing in attachment's linked notes

01zulfi:

### DIFF
--- a/packages/core/src/collections/relations.ts
+++ b/packages/core/src/collections/relations.ts
@@ -397,7 +397,8 @@ class RelationsArray<TType extends keyof RelatableTable> {
           )
           .$if(
             !!this.types?.includes("note" as TType) &&
-              this.db.notes.cache.archived.length > 0,
+              this.db.notes.cache.archived.length > 0 &&
+              this.reference.type !== "attachment",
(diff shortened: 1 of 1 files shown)


packages/core/src/collections/relations.ts
```diff
@@ -397,7 +397,8 @@ class RelationsArray<TType extends keyof RelatableTable> {
           )
           .$if(
             !!this.types?.includes("note" as TType) &&
-              this.db.notes.cache.archived.length > 0,
+              this.db.notes.cache.archived.length > 0 &&
+              this.reference.type !== "attachment",
```
thecodrr, line 401: Not a good idea to do this here. It'll affect the whole app even where we don't want archived notes showing in attachments relations.

01zulfi: I check all the places where we are fetching attachment's linked notes in `web` and `core`. All those cases should fetch archived notes as linked notes, that's why I hardcoded this condition here. Do you think we should pass this as an `options.includeArchivedNotes` (or similar) parameter?

thecodrr (approved):

packages/core/src/collections/relations.ts
thecodrr: Let's keep it for now.

PR #9016, opened 2025-12-01
Fix file size calculation

NeedsChloesure:
Notesnook's calculation of the chunks in an encrypted attachment is incorrect because it does not remove the header bytes from each chunk prior to calculating. This results in the file size erroneously failing if your file ends on a chunk border (your file is a multiple of 512*1024) or otherwise has a file size that results in the extra padding added to the encrypted blob pushing over a chunk boundary.

I tested this on my computer, and it works there. I noticed that the mobile app also has this problem, so I applied the patch there as well. I didn't test on mobile, I ran into difficulty building the Android app and gave up after a while. It should work though.

### DIFF
--- a/apps/mobile/app/common/filesystem/download.ts
+++ b/apps/mobile/app/common/filesystem/download.ts
@@ -70,7 +70,7 @@ export async function downloadFile(
       throw new Error(error);
     }
 
-    const totalChunks = Math.ceil(size / chunkSize);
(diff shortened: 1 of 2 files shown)


thecodrr:
Thanks for this. The changes make sense but we'll still need to test this.

cc @01zulfi

thecodrr:
I am able to reproduce this issue using this file: 
[upload_test.txt](https://github.com/user-attachments/files/23897503/upload_test.txt)

PR #9027, opened 2025-12-05
web: show spaces in note,notebook,tag,reminder list item titles

01zulfi:
I'm not sure why these list item titles are `nowrap`. I think we should render the spaces in the titles as that's what the user intended

### DIFF
--- a/apps/web/src/components/list-item/index.tsx
+++ b/apps/web/src/components/list-item/index.tsx
@@ -202,7 +202,7 @@ function ListItem<TItem extends Item, TContext>(
           data-test-id={`title`}
           variant={"body"}
           sx={{
-            whiteSpace: "nowrap",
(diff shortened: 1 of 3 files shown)


thecodrr:
They are nowrap because we don't want the titles to wrap to second line.

apps/web/src/components/list-item/index.tsx
```diff
@@ -202,7 +202,7 @@ function ListItem<TItem extends Item, TContext>(
           data-test-id={`title`}
           variant={"body"}
           sx={{
-            whiteSpace: "nowrap",
+            whiteSpace: "pre-wrap",
```
thecodrr: Use `pre` instead or some other value.

thecodrr (changes requested):

PR #9040, opened 2025-12-10
editor: reduce horizontal rule height

01zulfi:
# before

image

# after

image

### DIFF
--- a/packages/editor/styles/styles.css
+++ b/packages/editor/styles/styles.css
@@ -40,7 +40,7 @@
   position: relative;
   background-color: var(--separator);
   width: 100%;
-  height: 5px;
+  height: 2px;
   margin-top: 4px;
   margin-bottom: 4px;
 }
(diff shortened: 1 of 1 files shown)


thecodrr:
The reason the HR is so big is to make it easier on mobile to tap & select it.

01zulfi:
For large screens, the horizontal line will now be `2px` tall. For smaller screens, the previous `5px` value is kept.

Do let me know if you had something else in mind, @thecodrr

Starship-Headlight:
@thecodrr  @01zulfi sorry for bringing this up after the closure of the PR. But I wanted to know your opinions on this. This can be implemented in both mobile and desktop apps: 

Selecting a ruler in the current implementation can be a little challenging. 
Maybe it could have spaces on top and bottom. In that case the ruler can be trimmed further to make it look sleeker. 
It looks like this in Notesnook.
brave_pYAoDo7224

It could be like this:
brave_Ix0OVHoMch

Or this: 
brave_fCoEGSzqnE

The lines in the last two don't look crowded and selecting the ruler is also easier with a lot of clickable space.

PR #9090, opened 2025-12-24
web: add desktop drag region in auth pages

01zulfi:
this part of the auth pages is a drag region now
image


thecodrr:
I don't get it...why not just render the title bar instead?

01zulfi:
https://github.com/streetwriters/notesnook/pull/9093

PR #9093, opened 2025-12-25
web: show titlebar on auth pages

01zulfi:


thecodrr (approved):

thecodrr:
## QA Scope

- Needs to be tested on all Desktop platforms.
- Make sure the custom titlebar is visible in auth related pages (signup/login, password recovery etc.) on Linux
- Make sure the custom titlebar is NOT visible if a native titlebar is being used (e.g. on macOS & Windows)

PR #9159, opened 2026-01-06
web: allow editing note's created date from properties panel

01zulfi:
https://github.com/user-attachments/assets/5c78c028-ab0c-45a6-a13d-bdb8d3ef75ce


thecodrr (changes requested):
I think it shouldn't be possible to set the date in future.

01zulfi:
On the web side, these restrictions are applied
1. Creation date cannot be in the future
2. Creation date cannot be after the last edited date

thecodrr (approved):

Issue #542, opened 2022-04-23
Expiring notes (suggested by TheMatjaz#8264)


candroid-man:
**Feature Request**

> By that I mean setting a time-to-live to a note, so it's moved to trash automatically afterwards. Would be useful for temporary notes like grocery lists, which become useless anyway after a week or two. I tend to forget about deleting such notes.

**Personal Opinion**

This sounds extremely useful, especially if it was paired with another feature; a dedicated spot for temporary notes in the app. Like a folder or tag, and if you create a note using the widget or maybe in the future using a quick settings tile on Android, it would be default be put into temporary notes, and you can change the location whilst creating the note if you want it somewhere else.

ammarahm-ed:
Great idea!

thecodrr:
Let's add this. How it'll work:

Core:

1. We'll add a new property in the note object called `expiryDate`.
2. We'll also need a `cleanup` method in the `Notes` collection that cleans up all the expired notes
3. We will call the `cleanup` method after the `Notes` collection has initialized

Web:

1. There will be a new item in note properties context menu called "Set expiry"
2.  Clicking on it will open a context menu with some preset values like `1 day`, `1 week` etc and it'll also have a last option called `Custom`. Clicking on `Custom` will open a dialog with a number counter. The number counter represents the number of days after which the note will be removed.
3. We'll also add a new task in the scheduler that will run every 24 hours and call the `cleanup` function from the Core.

Mobile:

1. There will be a new button in note properties action sheet called "Set expiry"
2. Clicking on it will open a sheet which will have a few defaults and also a way to set a custom number of days.

alihamuh:
working on it

abhinav700:
@alihamuh is this issue still open?

PR #9167, opened 2026-01-07
Feat/expiring notes

ammarahm-ed:


01zulfi:
https://github.com/user-attachments/assets/45e9d944-a11a-4e97-9481-bc02d5840f53

apps/web/src/app-effects.tsx
```diff
@@ -235,3 +238,10 @@ function getProcessingStatusFromType(type: ProcessingType) {
       return [];
   }
 }
+
+async function scheduleExpiredNotesDeletion() {
+  await TaskScheduler.stop("delete-expired-notes");
+  TaskScheduler.register("delete-expired-notes", "0 0 * * *", async () => {
+    await db.notes.deleteExpiredNotes();
+  });
+}
```
01zulfi: haven't tested this

packages/core/src/api/sync/merger.ts
```diff
@@ -45,6 +46,27 @@ class Merger {
...
+      localItem.deletedBy === "expired"
+    ) {
+      if (
+        remoteItem.expiryDate.dateModified > localItem.expiryDate.dateModified
+      ) {
+        localItem.expiryDate = remoteItem.expiryDate;
+        (localItem as unknown as Note).type = "note";
+        (localItem as unknown as Note).deletedBy = null;
+        (localItem as unknown as Note).dateDeleted = null;
+        (localItem as unknown as Note).itemType = null;
```
thecodrr, line 65:
```suggestion
        localItem.type = "note";
        localItem.deletedBy = null;
        localItem.dateDeleted = null;
        localItem.itemType = null;
```

apps/web/src/components/note/index.tsx
```diff
@@ -848,3 +957,100 @@ async function copyNote(noteId: string, format: "md" | "txt") {
...
+        </PopupPresenter>
+      </Dialog>
+    );
+  }
+);
+
+function setDateOnly(str: string, date: dayjs.Dayjs) {
+  const value = dayjs(str, db.settings.getDateFormat(), true);
+  return date.year(value.year()).month(value.month()).date(value.date());
+}
```
thecodrr: This shouldn't be here.

apps/web/src/components/note/index.tsx
```diff
@@ -612,6 +653,74 @@ export const noteMenuItems: (
...
+          {
+            type: "button",
+            key: "custom",
+            title: "Custom",
+            onClick: async () => {
+              await CustomExpiryDateDialog.show({
+                noteIds: ids
+              });
+            }
+          },
```
thecodrr: These are unnecessary. We have no data to indicate that 1, 7 or 30 days are commonly used as expiry date so adding them here doesn't really speed up user's workflow but actually adds noise. There should be just a single button that opens the set expiry dialog.

apps/web/src/components/note/index.tsx
```diff
@@ -179,6 +189,19 @@ function Note(props: NoteProps) {
...
+                  text={getFormattedDate(note.expiryDate, "date")}
+                  styles={
+                    lessThanOneDayRemaining(note.expiryDate)
+                      ? {
+                          text: { color: "var(--accent-error)" }
+                        }
+                      : {}
+                  }
+                />
+              )}
```
thecodrr: I don't think it makes sense to show all this in `compact` mode. Just show the destruct icon.

apps/web/src/components/note/index.tsx
```diff
@@ -268,6 +291,20 @@ function Note(props: NoteProps) {
...
+
+              {note.expiryDate ? (
+                <IconTag
+                  icon={Destruct}
+                  text={getFormattedDate(note.expiryDate, "date")}
+                  styles={
+                    lessThanOneDayRemaining(note.expiryDate)
+                      ? {
+                          text: { color: "var(--icon-error)" }
+                        }
```
thecodrr: Why do we want to highlight this?

thecodrr (changes requested):

apps/web/src/components/note/index.tsx
01zulfi: This was based on the linked issue comment https://github.com/streetwriters/notesnook/issues/542#issuecomment-1500953688

Now it is just the "Set expiry date" menu option for the note which doesn't has expiry. If expiry date is set, then there are additional "Change" and "Remove" sub menu items

apps/web/src/components/note/index.tsx
01zulfi: I thought it would be nice; although could be distracting, removed

apps/web/src/components/note/index.tsx
```diff
@@ -848,3 +938,96 @@ async function copyNote(noteId: string, format: "md" | "txt") {
       showToast("error", `${strings.failedToCopyNote()}: ${e.message}.`);
   }
 }
+
+type NoteExpiryDateDialogProps = BaseDialogProps<boolean> & {
+  noteId: string;
+  expiryDate?: number | null;
+};
+
+const NoteExpiryDateDialog = DialogManager.register(
```
thecodrr: Move this to its own file.

apps/web/src/components/note/index.tsx
```diff
@@ -298,6 +325,10 @@ export default React.memo(Note, function (prevProps, nextProps) {
 //     .pin(note.id);
 // };
 
+function lessThanOneDayRemaining(date: number) {
```
thecodrr: This is no longer needed right?

PR #9169, opened 2026-01-07
editor: add options to export/import table <> csv

01zulfi:
https://github.com/user-attachments/assets/1141ac58-f934-4d59-a43a-11e9917b9a4a

### DIFF
--- a/packages/intl/locale/en.po
+++ b/packages/intl/locale/en.po
@@ -2729,7 +2729,7 @@ msgstr "Export all notes as pdf, markdown, html or text in a single zip file"
 msgid "Export as{0}"
 msgstr "Export as{0}"
 
-#: src/strings.ts:2623
+#: src/strings.ts:2633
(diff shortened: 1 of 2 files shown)


packages/editor/src/extensions/table/table.ts
```diff
@@ -244,6 +244,16 @@ declare module "@tiptap/core" {
...
+
+      /**
+       * Export the current table as a CSV file
+       */
+      exportTableAsCsv: () => ReturnType;
+
+      /**
+       * Import a CSV file into a new table at the current position
+       */
+      importCsvToTable: () => ReturnType;
```
01zulfi, line 256: Needed to add these command types for checking perms.

Maybe I could implement these commands itself and use `editor.commands.exportTableAsCsv` instead of writing the login in `table/actions.ts` file, but not sure

PR #9180, opened 2026-01-08
web: add referer header with content origin

01zulfi:
* youtube embeds aren't working without adding this header


thecodrr:
Fix confirmed.

PR #9186, opened 2026-01-09
web: add option to preview attachments in attachment manager

01zulfi:
https://github.com/user-attachments/assets/03276718-10c9-4b5d-8440-c0e959f1de41


01zulfi:
There already is a full screen button so I didn't think of having full width&height dialog. I've added it now though

PR #9201, opened 2026-01-14
web: enable toc toggle on new notes

01zulfi:
I think we can keep the toc toggle enabled on new notes. Mostly because when the toc is open and user creates a new note, the toc panel takes up space that cannot be released.


thecodrr:
Haven't we fixed that issue?

01zulfi:
https://github.com/streetwriters/notesnook/pull/9002

That was for properties panel, I think that's a different issue because for new notes the properties panel isn't even shown (it would just take up space)

Issue #9107, opened 2025-12-28
Dropdown icon not visible on Callouts with Blank Title


AlbusPercivalDumbledore:
### What happened?

https://github.com/user-attachments/assets/fde3ce2d-f1c1-474b-9bbf-067cdae2b0e2

### Steps to reproduce the problem

1. Insert a Callout
2. Remove its Title

### Version

3.3.7-c560685-desktop

### Platform/OS

Linux

### Relevant log output

```shell

```

PR #9220, opened 2026-01-15
editor: fix hidden collapse/expand icon when callout title is empty

01zulfi:
Closes https://github.com/streetwriters/notesnook/issues/9107

### DIFF
--- a/packages/editor/styles/styles.css
+++ b/packages/editor/styles/styles.css
@@ -965,6 +965,10 @@ del.diffdel {
   display: none !important;
 }
 
+.ProseMirror div.callout > :is(h1, h2, h3, h4, h5, h6):first-child.empty::after {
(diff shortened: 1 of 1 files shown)


packages/editor/styles/styles.css
```diff
@@ -965,6 +965,10 @@ del.diffdel {
   display: none !important;
 }
 
+.ProseMirror div.callout > :is(h1, h2, h3, h4, h5, h6):first-child.empty::after {
+  display: inline-block !important;
+}
```
thecodrr, line 970: Why do we not make it `inline-block` by default? Does that create issues?

01zulfi: it is `inline-block` by default, but for empty headings we apply `display: none !important`

we need to override for first callout heading

PR #9252, opened 2026-01-21
web: All tabs closed when any note is deleted.

NeedsChloesure:
The logic was missing a check to ensure that only the tabs that correspond to the recently deleted object were being marked for closure. The lack of this check meant that for any deleted note item, all tabs would pass the `if (isDeleted(item) || isTrashItem(item))` check since the open tab loop runs on every tab for every item.

### DIFF
--- a/apps/web/src/stores/editor-store.ts
+++ b/apps/web/src/stores/editor-store.ts
@@ -286,15 +286,7 @@ class EditorStore extends BaseStore<EditorStore> {
         const clearIds: string[] = [];
         for (const session of sessions) {
           if (session.type === "new") continue;
-          if (session.note.id !== item.id) continue;
-          const noteId = isDeleted(item)
-            ? null
-            : item.type === "note"
-            ? item.id
-            : item.type === "tiptap"
-            ? item.noteId
-            : null;
-          if (noteId && session.note.id !== noteId) continue;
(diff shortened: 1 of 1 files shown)


apps/web/src/stores/editor-store.ts
```diff
@@ -286,6 +286,7 @@ class EditorStore extends BaseStore<EditorStore> {
         const clearIds: string[] = [];
         for (const session of sessions) {
           if (session.type === "new") continue;
+          if (session.note.id !== item.id) continue;
```
thecodrr: The check on line 297 is doing the same thing.

thecodrr (changes requested):

apps/web/src/stores/editor-store.ts
NeedsChloesure: I removed the check on line 297 and definition of NoteId. NoteId isn't set when the note is deleted, which makes the check following the definition fail. Why check if the incoming item is deleted before grabbing it's ID? Isn't it just faster to check the IDs up front? Are there cases where we do not have this information available in `item`?

For the potential edge case where a content ID is passed here, wouldn't the note ID still follow (or already gone through)?

thecodrr:
Thank you  so much for this and sorry it took so long to merge!

PR #9283, opened 2026-01-30
fix: unable to change zoom level back to 1.0

### LINKED ISSUE #4035: Unable to change Zoom Level back to 1.0 (from 0.5 or 1.5)
Platform: Desktop (Windows 10)
Impact: Low - Medium - High
Issue: After changing the Zoom Factor from the default 1.0 (up or down) - I am unable to change it back to 1.0
I can type "1.0" in the field and the screen will resort to that view, but if I click out of that window it resorts to the last Zoom level.
I Restarted the app and it will not allow me to change back to 1.0 default zoom level.

**Device information:**
App version: 2.6.13-2f3bddf-desktop
OS: Windows 10 64-bit
Browser: Electron 25.9.8
Pro: true

littleKitchen:
## Summary
Fixed the issue where users couldn't change the zoom level back to 1.0 after changing it to another value.

## Root Cause
The number input for zoom factor used:
1. `defaultValue` (uncontrolled) - which doesn't update the displayed value when the store changes
2. 500ms debounced `onChange` - which doesn't commit the value if the user clicks away quickly

This caused a race condition where:
1. User types 1.0, the visual zoom changes
2. User clicks away before 500ms passes
3. The debounced onChange never fires, value isn't saved
4. On settings refresh, the old value is loaded back

## Fix
1. **Added `onBlur` handler** - Immediately commits the value when the input loses focus, ensuring values are saved even if the user navigates away before the debounce fires
2. **Added `key` prop** - Forces re-render when the stored value changes, ensuring the input always reflects the actual persisted value

Also applied the same fix to text inputs for consistency.

Fixes #4035

### DIFF
--- a/apps/web/src/dialogs/settings/index.tsx
+++ b/apps/web/src/dialogs/settings/index.tsx
@@ -581,6 +581,7 @@ function SettingItem(props: { item: Setting }) {
               case "input":
                 return component.inputType === "number" ? (
                   <Input
+                    key={component.defaultValue()}
                     type={"number"}
                     min={component.min}
                     max={component.max}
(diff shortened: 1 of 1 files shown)


thecodrr:
Thank you!

PR #9289, opened 2026-02-01
web: thumbnail generation for first image and settings

marcolaux:
## Thumbnails

- creates a thumbnail of the first image of a note and saves it with the note object
- reacts on changes in the editor, whatever image comes first becomes the thumbnail
- global setting to turn this off
  - if enabled, a per note setting in the properties is also available
- for previous notes, that lack a thumbnail: it gets generated automatically, when the note gets opened
- in addition the note title is 14px instead of 12px now - makes it easier to read

grafik

grafik

### DIFF
--- a/packages/intl/locale/en.po
+++ b/packages/intl/locale/en.po
@@ -13,7 +13,7 @@ msgstr ""
 "Language-Team: \n"
 "Plural-Forms: \n"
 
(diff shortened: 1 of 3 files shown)


marcolaux:
the commit was a bit long because I did not put the new strings at the end of the file.
it is better now :)

PR #9301, opened 2026-02-04
core: remove orphaned attachments in trash cleanup method

01zulfi:
Closes https://github.com/streetwriters/notesnook/issues/9280


thecodrr:
This looks good but we should add a new endpoint in the server to allow bulk deletion.

thecodrr:
@01zulfi can you update this to use the bulk delete endpoint?

Also I don't think we should `await` the bulk delete request because it might cause a hang if there's a network issue. It should be fire and forget with logging if an error occurs.

PR #9302, opened 2026-02-04
web: disable desktop update checker if auto updates off

### LINKED ISSUE #9285: Disabling update mechanism

Automatic updates can be disabled, but not the update check. It is often not needed, as OS package manager or store do it already. Also it can show update available when it is not yet pushed to the actual repository user gets updates from.

A toggle to disable updates check. Should be toggled on by default when installed on Linux from Flathub or package manager and on Android.

_No response_

### Additional context

_No response_

01zulfi:
Closes #9285 

_

### DIFF
--- a/apps/web/src/hooks/use-auto-updater.ts
+++ b/apps/web/src/hooks/use-auto-updater.ts
@@ -22,7 +22,7 @@ import { checkForUpdate } from "../utils/updater";
 import { AppEventManager, AppEvents } from "../common/app-events";
 import BaseStore from "../stores";
 import createStore from "../common/store";
-import { store as settingStore } from "../stores/setting-store";
+import { useStore as useSettingStore } from "../stores/setting-store";
 
 type CompletedUpdateStatus = { type: "completed"; version: string };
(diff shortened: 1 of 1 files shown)


apps/web/src/hooks/use-auto-updater.ts
```diff
@@ -104,7 +104,7 @@ export function useAutoUpdater() {
     );
 
     checkingForUpdate();
-    checkForUpdate().catch(console.error);
+    checkForUpdate(settingStore.get().autoUpdates).catch(console.error);
```
thecodrr: Are we sure that `autoUpdates` has the updated value here? If not then it won't ever check for updates on startup.

01zulfi: that was a nice catch, thanks!

I'm now using a listener so we can guarantee the latest value. One side-ffect is that update checker will run anytime the value changes, which I don't think should be a huge issue

thecodrr: What if the value changes _before_ we get to the update checker?

PR #9326, opened 2026-02-10
editor: escape html in twitter embed src

01zulfi:


packages/editor/src/extensions/embed/component.tsx
```diff
@@ -204,9 +204,19 @@ function isTwitterX(src: string) {
...
 }
 
+function escapeHtml(unsafe: string): string {
+  return unsafe
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&#039;");
+}
```
thecodrr: Instead of doing this manually, we should use the browser's built in escaping by creating the `a` element using `document.createElement` and setting the `href` attribute.

PR #9328, opened 2026-02-10
web: allow locked notes to stay unlocked if keep vault note unlocked is enabled

01zulfi:


01zulfi:
closing since this is not required anymore, instead a vault lock timer will be added https://github.com/streetwriters/notesnook/pull/9395

PR #9331, opened 2026-02-11
Harden cmd function in scripts/build.mjs against command injection

ot-net:
Updated the cmd function in scripts/build.mjs.

shell: true can allow command injection if arguments include untrusted input.

Switching to shell: false is safer, but may break commands using shell features like pipes (|), chaining (&&), or glob patterns (*.js).

Testing is required before merging the switch to make sure it doesn't break.

### DIFF
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -84,7 +84,11 @@ await Promise.all([
 ]);
 
 async function cmd(...command) {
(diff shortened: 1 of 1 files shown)


thecodrr:
`shell: true` doesn't work with Windows. Thanks for the PR though.

Issue #9336, opened 2026-02-12
Please align the editor elements 🖋️


InsaneEagle-dot:
### What problem are you facing?

This is a small change but it can significantly improve the writing experience in Notesnook.

Please see the two screenshots. One of them has properly aligned elements and another one is from Notesnook.

This screenshot is from Notesnook. The elements are not aligned:

![Image](https://github.com/user-attachments/assets/2b9a9338-834b-4a6e-8e13-4cddcc4ab3fa)

This screenshot has the elements aligned and looks quite pleasing:
![Image](https://github.com/user-attachments/assets/f9a1b335-12b0-433e-bee3-2a92204e9bac)

### What's the solution you'd like to see in Notesnook?

Aligning the elements. Please check the screenshot. The text starts at the same place in every indented element. All the icons are also aligned. It provides the text with a sort of coherence. 

(If possible please allow the ability to change the color of these elements with themes like shown in the screenshot.)

PR #9358, opened 2026-02-16
editor: fix checklist & blockquote indent so text starts at same pos as other lists

01zulfi:
related https://github.com/streetwriters/notesnook/issues/9336

# before

image

# after

image

### DIFF
--- a/packages/editor/styles/styles.css
+++ b/packages/editor/styles/styles.css
@@ -393,6 +393,8 @@ img.ProseMirror-separator {
 .ProseMirror blockquote {
   border-inline-start: 5px solid var(--border);
   padding-inline-start: 15px;
+  margin-inline-start: 20px;
+  margin-inline-end: 0px;
 }
 
 @media screen and (max-width: 480px) {
@@ -720,7 +722,7 @@ p > *::selection {
 .ProseMirror ul.simple-checklist {
   list-style: none;
   padding-inline: 0px !important;
-  padding-inline-start: 24px !important;
+  padding-inline-start: 40px !important;
 }
 
 .ProseMirror li.nested > ul.simple-checklist {


InsaneEagle-dot:
Thank you so very much for this @01zulfi .  It looks really good. 

Just a few questions if you could please tell me. 

Is it possible to move the indented items a little left so that it's not too off from the headings or text as show in the screenshot in the feature request

Is it possible to make the dots and block quote vertical ruler appear in the middle of the check boxes. The current implementation is still just a little off. For example if you drop a straight vertical line in the middle of the checkbox, it should cut the dots and the vertical ruler of blockquotes right in the middle. It's a nitpick but it will make it look more coherent. 

(The numbers in ordered list can be a little off (as it is in the current PR) as it will go from 1 - 1000; 1-99 will make the 99 move in the middle then 100 will move a little to the left and then on 1000 the editor will make the last `0` appear in the right and move the text a little to the right. I hope I am able to explain.)

And maybe the space between the `dot`, `checkbox`,  `number.`  and text can be increased a little to make the text look less congested. Not by much tho. Just a tiny bit. 

Also, did you make it same for outline lists too?  Or all indented elements available?

Again, thanks a lot for this PR 😄

01zulfi:
The purpose of this PR was to have all the text in indented items start from the same positions. 

> Is it possible to move the indented items a little left so that it's not too off from the headings or text as show in the screenshot in the feature request

The level of indentation is a very subjective change. I'll defer this decision to @thecodrr 

> Is it possible to move the indented items a little left so that it's not too off from the headings or text as show in the screenshot in the feature request

I'll look into it

> (The numbers in ordered list can be a little off (as it is in the current PR) as it will go from 1 - 1000; 1-99 will make the 99 move in the middle then 100 will move a little to the left and then on 1000 the editor will make the last 0 appear in the right and move the text a little to the right. I hope I am able to explain.)

Didn't notice this. Could you share a screenshot please?

> And maybe the space between the dot, checkbox, number. and text can be increased a little to make the text look less congested. Not by much tho. Just a tiny bit.

Will defer to @thecodrr 

> Also, did you make it same for outline lists too? Or all indented elements available?

Didn't notice any need for change in outline lists so kept it as is

InsaneEagle-dot:
@01zulfi Please check these. 

image
image
image
image


The numbers expand to left until they reach 3 digits (100). The unit digits stays in the place till 99 and tens digit moves to the right starting 10. Also the text moves a little to the right at this point. At 100 the text moves further right as the 1 in 100 is at the same place as 9 in 98. After which the numbers expand to right side and the text also moves. Making the ordered list always stay indented to the text below or on top of it. 

How it looks in Notesnook 👇:

image

image


In Notesnook the numbers expand to the left and sometimes they even get out of the editor as shown in the last screenshot.

PR #9395, opened 2026-02-24
web: add vault lock timer setting

01zulfi:


apps/web/src/dialogs/settings/vault-settings.tsx
```diff
@@ -50,6 +50,36 @@ export const VaultSettings: SettingsGroup[] = [
...
+        description:
+          strings.lockVaultAfterDesc(),
+        isHidden: () => !useAppStore.getState().isVaultCreated,
+        onStateChange: (listener) =>
+          useAppStore.subscribe((s) => s.vaultLockAfter, listener),
+        components: [
+          {
+            type: "dropdown",
+            options: [
+              { title: "1 minute", value: 1000 * 60 * 1 },
```
thecodrr, line 65: Add an `Immediately` option in which we don't store the password at all.

thecodrr (changes requested):
Add

apps/web/src/dialogs/settings/vault-settings.tsx
01zulfi: If we don't store the password, then when a note is unlocked, it won't be able to be edited and will throw the vault locked error

```
index.mjs:20  2026-02-25T09:09:46.953Z | ERROR | [notesnook-web]: Error: ERR_VAULT_LOCKED
    at Vault.<anonymous> (vault.js:229:23)
    at Generator.next (<anonymous>)
    at fulfilled (http://localhost:3000/@fs/home/zulfi/code/notesnook/notesnook/packages/core/dist/esm/api/vault.js:22:58) {} [62ms]
 ```

thecodrr (approved):

ayesha004tanveer-netizen (changes requested):
@01zulfi 

**Bug Title:**
No notification when vault auto-locks

**Description:**
When you lock the vault manually, the app shows a notification saying “Vault locked.” But if the vault locks automatically using the timer, no notification appears.

**Steps to Reproduce:**

1. Open the app.
2. Lock the vault manually → see the “Vault locked” notification.
3. Go to settings and enable auto-lock (set timer to 1 minute).
4. Wait for 1 minute and see what happens.

**Expected Result:**
A notification should appear when the vault locks automatically,  just like the manual lock.

**Actual Result:**
No notification appears when the vault auto-locks.

Reference Screenshot attached:


Image

ayesha004tanveer-netizen (changes requested):
[screen-capture (88).webm](https://github.com/user-attachments/assets/93761ab5-b720-41cd-8548-566f1de32d53)

@01zulfi 


**Bug Title:** 
Vault locks immediately after selecting auto-lock time

**Description:**
When I select an auto-lock time (like 5 or 10 minutes), the vault locks right away. After unlocking, it locks again after the selected time, which is correct. But it should not lock immediately when I choose the time.

**Steps to Reproduce:**

1. Go to vault settings.
2. Select an auto-lock time (5 or 10 minutes).
3. See that the vault locks immediately.
4. Unlock the vault and wait, the vault locks again after the selected time.

**Expected Result:**
The vault should lock only after the selected time passes.

**Actual Result:**
The vault locks immediately on selecting the option

Bug Video is attached:

PR #9419, opened 2026-02-27
editor: optimize search performance

01zulfi:
* do matchAll per text block instead of matchAll on entire text string

Related https://github.com/streetwriters/notesnook/pull/1788

### DIFF
--- a/packages/editor/src/extensions/search-replace/search-replace.ts
+++ b/packages/editor/src/extensions/search-replace/search-replace.ts
@@ -26,6 +26,7 @@ import {
   Transaction,
   TextSelection
 } from "prosemirror-state";
+import { Node as PmNode } from "prosemirror-model";
 import { SearchSettings } from "../../toolbar/stores/search-store.js";
 import { tiptapKeys } from "@notesnook/common";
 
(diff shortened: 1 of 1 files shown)


packages/editor/src/extensions/search-replace/search-replace.ts
```diff
@@ -104,57 +125,16 @@ function searchDocument(
...
-        lastNode.text += "\n";
-        cursor++;
+  for (const { text, pos } of extractTextWithPosition(doc)) {
+    for (const match of text.matchAll(searchTerm)) {
+      if (match[0] === "") break;
+      if (match.index !== undefined) {
+        results.push({
+          from: pos + match.index,
+          to: pos + match.index + match[0].length
+        });
```
01zulfi, line 135: Got this directly from: https://github.com/streetwriters/notesnook/pull/1788 (There are also other changes that might optimize search further, have decided to include them in a future PR, so we can test this current change first)

The search speed improvement isn't noticeable because previous search was still fast (<100ms for 1000+ matches in a 100k+ words note). Though I've verified with console.time, this change improves the speed anywhere from 3x to 17x.

PR #9429, opened 2026-03-02
web: don't ask for account password for vault deletion for not logged in users

01zulfi:
## QA Scope

Needs to be tested on web & mobile. When user is logged in, it should ask for account password to delete the vault. When user is not logged in, it shouldn't ask for account password to delete the vault

### DIFF
--- a/apps/web/src/common/vault.ts
+++ b/apps/web/src/common/vault.ts
@@ -23,6 +23,7 @@ import { showToast } from "../utils/toast";
 import { VAULT_ERRORS } from "@notesnook/core";
 import { strings } from "@notesnook/intl";
 import { useStore as useAppStore } from "../stores/app-store";
+import { useStore as useUserStore } from "../stores/user-store";
 
 class Vault {
   static async createVault() {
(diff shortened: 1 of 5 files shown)


apps/web/src/common/vault.ts
```diff
@@ -63,29 +64,47 @@ class Vault {
...
+            deleteAllLockedNotes: {
+              text: strings.deleteAllNotes(),
+              default: false
+            }
+          },
+          validate: ({ password }) => {
+            return db.user.verifyPassword(password);
+          }
+        })
+      : showPasswordDialog({
```
thecodrr, line 88: Why not just use a `confirmDialog` here?

thecodrr (changes requested):

apps/web/src/common/vault.ts
01zulfi: `ConfirmDialog` would give a different return type here `result.checks.deleteAllLockedNotes` while `showPasswordDialog` gives `result.deleteAllLockedNotes`. I think it is cleaner to just use the `showPasswordDialog` instead of handling two different return types

thecodrr (approved):

Issue #9421, opened 2026-02-27
search does not display results, despite seemingly finding them


code-consensus:
### What happened?

[v3.0.1](https://blog.notesnook.com/notesnook-v3.0.1/) introduced FTS5 search syntax and [v3.0.18](https://blog.notesnook.com/notesnook-v3.0.18/) made search diacritic "agnostic" -- but the issue is that something must have happened since then and now, because on v3.3.8 these don't seem to be working any longer (I am a new user, and the only version I've ever used is this latest version).

### Diacritics
from v3.0.18 release notes: "Previously, if you searched for `cafe`, it wouldn't return notes that contain `café`. Now, it will return notes that contain either `cafe` or `café`" -- the search doesn't behave this way any longer:

1. I go to a note where there is such content, e.g. the word `también` or `España`.
2. I do Ctrl+F to open the note-only search (_not_ the global database search)
3. I type `tambi` and it finds it and highlights it.
4. I continue typing `tambie` and the highlight disappears and it says 0/0 found.
5. I erase the e and do `tambié`, the term is found and highlighted.

So something must have broken between v3.0.18 and v3.3.8 to make the "diacritic free" search no longer function. Note that another user reported this related to _imported_ notes (issue #9227), except I don't think this has anything to do with only imported notes, because I am seeing this in also completely brand new notes.

### Global search finding, but not _displaying_, results
There are certain search letters, terms, phrases, that seem are being found properly, but _not_ **_displayed_** to the user.

In global search, if _nothing_ is found the program informs you of this in one way or another. 
e.g. if I search for something that is definitively _not_ in my notes, e.g. `zzzzzzffff`:
- On mobile (v3.3.12 f-droid) you get a specific message on what you searched: "TIP No results found for zzzzzzffff".
- On desktop (v3.3.8), I get "TIP Wrap a query in double quotes to search for an exact match."

Going now to my example of "también" above, except this time, don't search in the note, but search in the "global database search" bar (however you call it):

1. I go to the global search, or type Ctrl+F+F
2. I enter the term I know is in my database, here I say again `también`. The search will show the note in the list that has that word.
3. I enter now instead `tambifn`, a slight variation that I know does _not_ exist in the database, and I get the "tip: no results found for tambifn"
4. I enter now `tambien` (the variant without the diacrtic) -- now the search list is _empty_ showing no notes, but there is **NOT** the message that no results were found. It seems as if the app _is_ finding the term, but somehow just not _displaying_ the search results.

[...]

PR #9479, opened 2026-03-10
core: fix diacritics search in db.lookup.notesWithHighlighting

01zulfi:
## Description

Closes https://github.com/streetwriters/notesnook/issues/9421

Fix diacritic agnostic search not working in global search

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests
- [X] Added/updated tests for this change (if needed)

## Platform

- [X] Web
- [X] Mobile
- [X] Desktop

## Sign-off
- [X] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/core/src/api/lookup.ts
+++ b/packages/core/src/api/lookup.ts
@@ -772,25 +772,12 @@ function highlightQueries(
   try {
     const regex = new RegExp(patterns.join("|"), "gi");
     const normalizedText = removeDiacritics(text);
-
-    let hasMatches = false;
-    let matchIdCounter = 0;
-    let result = "";
-    let lastIndex = 0;
-    let m: RegExpExecArray | null;
-
-    while ((m = regex.exec(normalizedText)) !== null) {
-      hasMatches = true;
-      result += text.slice(lastIndex, m.index);
-      result += createSearchResultTag(
-        text.slice(m.index, m.index + m[0].length),
-        `match-${++matchIdCounter}`
-      );
-      lastIndex = m.index + m[0].length;
-    }
-
-    result += text.slice(lastIndex);
-
+    const { result, hasMatches } = highlightRegexMatches(
+      text,
+      normalizedText,
+      regex,
+      0
+    );
     return { text: result, hasMatches };
   } catch (error) {
     return { text, hasMatches: false };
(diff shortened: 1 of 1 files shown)


packages/core/src/api/lookup.ts
```diff
@@ -768,20 +768,31 @@ function highlightQueries(
...
-    });
+      result += text.slice(lastIndex, m.index);
+      result += createSearchResultTag(
+        text.slice(m.index, m.index + m[0].length),
+        `match-${++matchIdCounter}`
+      );
+      lastIndex = m.index + m[0].length;
+    }
+
+    result += text.slice(lastIndex);
```
thecodrr: Why this change? `text.replace` doesn't work here?

packages/core/src/api/lookup.ts
```diff
@@ -1151,25 +1175,30 @@ function getMatchScore(
 }
 
 function textContainsTokens(text: string, tokens: QueryTokens) {
-  const lowerCasedText = text.toLowerCase();
+  const normalizedText = removeDiacritics(text).toLowerCase();
+  const normalizedTokens = {
+    andTokens: tokens.andTokens.map((t) => removeDiacritics(t)),
+    orTokens: tokens.orTokens.map((t) => removeDiacritics(t)),
+    notTokens: tokens.notTokens.map((t) => removeDiacritics(t))
+  };
```
thecodrr: Instead of doing this here, why don't we do it in `transformTokens` once? Right now `removeDiacritics` is basically everywhere. We should use it at source to avoid all this clutter.

packages/core/src/api/lookup.ts
```diff
@@ -768,20 +768,31 @@ function highlightQueries(
 
   const patterns = queries
     .filter((q) => q.length > 0)
-    .map((q) => q.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"));
+    .map((q) => removeDiacritics(q).replace(/[.*+?^${}()|[\]\\]/g, "\\$&"));
```
thecodrr: We won't need to call `removeDiacritics` here if we just remove diacritics at source in `transformTokens`

packages/core/src/api/lookup.ts
```diff
@@ -979,14 +990,27 @@ function highlightHtmlContent(html: string, queries: string[]): string {
   const parser = new Parser(
     {
       ontext(text) {
+        const normalizedText = removeDiacritics(text);
```
thecodrr, line 977: Why not remove diacritics in the source `html`?

packages/core/src/api/lookup.ts
01zulfi: We need to remove diacritics from the text to search. But, we should add the search result tags on the original text instead of the normalized text. That's why we have to run a loop and build the text from scratch. Please let me know if there's a better approach

packages/core/src/api/lookup.ts
01zulfi: Same reason as above. Remove diacritics from source html will alter that html. We only need to remove diacritics for searching purposes, but keep the original one to show to the user

packages/core/src/api/lookup.ts
```diff
@@ -979,14 +987,27 @@ function highlightHtmlContent(html: string, queries: string[]): string {
...
+        while ((m = searchRegex.exec(normalizedText)) !== null) {
+          processed += text.slice(lastIndex, m.index);
+          processed += createSearchResultTag(
+            text.slice(m.index, m.index + m[0].length),
+            `match-${++matchIdCounter}`
+          );
+          lastIndex = m.index + m[0].length;
+        }
+
+        processed += text.slice(lastIndex);
```
thecodrr: Maybe we can move this to a function since the logic is identical to the one we use above.

thecodrr (approved):

PR #9485, opened 2026-03-11
core: add fallback to return the note if search highlighting returns empty values

01zulfi:
## Description

Related https://github.com/streetwriters/notesnook/issues/9421

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [X] Mobile
- [X] Desktop

## Sign-off
- [X] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/core/src/api/lookup.ts
+++ b/packages/core/src/api/lookup.ts
@@ -401,6 +401,28 @@ export default class Lookup {
           );
         }
 
+        /**
+         * If both title and content is missing in the result object, then someting went wrong in the highlighting process.
+         * Instead of returning empty values, fetch the title so the matched note is still returned,
+         * even if highlighting information is unavailable.
+         */
+        const missingTitleAndContentResults = results.filter(
+          (r) => !r.content.length && !r.title.length
+        );
+        if (missingTitleAndContentResults.length > 0) {
+          const ids = missingTitleAndContentResults.map((r) => r.id);
+          const notes = await db
+            .selectFrom("notes")
+            .where("id", "in", ids)
+            .select(["id", "title"])
+            .execute();
+          for (const note of notes) {
+            const result = results.find((r) => r.id === note.id);
+            if (!result || !note.title) continue;
+            result.title = stringToMatch(note.title);
+          }
+        }
+
         return {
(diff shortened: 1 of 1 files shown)


packages/core/src/api/lookup.ts
```diff
@@ -404,6 +404,28 @@ export default class Lookup {
...
+            .where("id", "in", ids)
+            .select(["id", "title"])
+            .execute();
+          for (const note of notes) {
+            const result = results.find((r) => r.id === note.id);
+            if (!result || !note.title) continue;
+            result.title = stringToMatch(note.title);
+          }
+        }
+
```
01zulfi, line 425: Hopefully there aren't any new cases where this fallback is triggered except for in https://github.com/streetwriters/notesnook/issues/9421

Also, this logic is similar to the [`resultsWithMissingTitles`](https://github.com/streetwriters/notesnook/blob/dda36bb036ab8fa442cde0d0a04aad86ea1a72ad/packages/core/src/api/lookup.ts#L378), but that had a different `filter` logic, so I'm not sure what the impact would have been.  That's why I didn't modify that but instead created this new `missingTitleAndContentsResults` array.

packages/core/src/api/lookup.ts
```diff
@@ -404,6 +404,28 @@ export default class Lookup {
...
 
+        /**
+         * If both title and content is missing in the result object, then someting went wrong in the highlighting process.
+         * Instead of returning empty values, fetch the title so the matched note is still returned,
+         * even if highlighting information is unavailable.
+         */
+        const missingTitleAndContentResults = results.filter(
+          (r) => !r.content.length && !r.title.length
+        );
+        if (missingTitleAndContentResults.length > 0) {
```
thecodrr, line 412: I don't think this is a good idea because sometimes there are matches in the HTML markup part which is a false positives. This will show those as matches as well. This should be tested to gauge the negative impact.

Perhaps we can handle this in the title highlighting loop by falling back to including the whole title?

thecodrr:
Closing as this is no longer necessary.

Issue #9422, opened 2026-02-28
link visualization issues


code-consensus:
### What happened?

I would like to note a couple of issues with how links appear that decrease usability.

### The hovering link-submenu is often too "far away" to be able to reach with your mouse
When you hover over a link, you get a pop-up submenu that includes the URL and other buttons to manage the link. The issue is, at least on Windows, that this little submenu is very often "too far" from the mouse cursor, such that upon moving the mouse to go to the menu, the menu disappears.

To replicate the _problematic_ behavior most easily:
1. find a link in a note
2. position your mouse cursor somewhere _below_ that text.
3. move your mouse _slowly_ until you you _just_ hit the link. So as soon as the cursor becomes a finger, stop moving.
4. wait for the submenu to appear.
5. now try to move your cursor to the submenu -- the submenu will instantly disappear.
6. No matter how fast you try to move the mouse, you will never "catch" the submenu before it disappears.

I realize the above process is a bit extreme, but I listed it because that will make the problem happen for sure (for your troubleshooting). But from my own normal usage of the application, it is happening to me almost all the time, e.g. moving to the link from the side, etc.

When I do need to access the submenu, my workaround is to _carefully_ move my cursor _from the top_ of the link, then wait for the menu to appear -- in this instance, the submenu will _not_ disappear as I move to it.

### Bi-directional note links could show _title_ instead of the raw link "nn://..."
When you hover over a bi-directional note link, the submenu that appears is visually exactly the same as when it is an external URL link. That is, you see the URL that the link is pointing to, which in the case of a bi-directional note link it is looks like "nn://note/5660b4fe222...".

Whereas to see the URL of an external link, such as "https://www.google.com/", is instructive for the user, in the case of a bi-directional note link to see the raw URL of the note as internally represented in the Notesnook database ("nn://note/...") does not convey very useful information to the user.

I think a more helpful behavior for the pop-up for specifically a bi-directional link would be instead to show the _title_ of the note that is being linked. This would allow the user to also understand what the link is pointing to in the case of a link text that is not directly related to the linked note's title either (e.g. a note link with text "see <ins>this</ins> note", if the pop-up then showed instead of "nn://note/..." showing "Bean soup recipe").

### Update documentation on the Evernote importation help page
Just on the topic of bi-directional links, the [import help page for Evernote](https://help.notesnook.com/importing-notes/import-notes-from-evernote) states that "Internal note links" are imported into Notesnook (there is a checkmark next to it).

[...]

PR #9487, opened 2026-03-11
editor: remove vertical offset on link hover popup

01zulfi:
## Description

Back in April last year, we moved the y offset of the hover popup from 10 to -10: https://github.com/streetwriters/notesnook/pull/8011

This was done so that when the hover popup opened, the link text wasn't obstructed and remained visible to the user. But, this offset adds an issue as reported in this issue https://github.com/streetwriters/notesnook/issues/9422: from certain positions the offset is so much that navigation onto the popup before it disappearing becomes a challenge. I remember facing this issue back when working on comments https://github.com/streetwriters/notesnook/pull/8984, so I set the offset to 0.

I'm doing the same in this PR, setting the offset to 0. I also wonder if we should add a delay (100ms or something) between navigating away from the element and closing the popup?

Screenshot of hovering from bottom of the link:

**before**

image

**after**

image

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [ ] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/editor/src/toolbar/floating-menus/hover-popup/index.tsx
+++ b/packages/editor/src/toolbar/floating-menus/hover-popup/index.tsx
@@ -125,7 +125,7 @@ export function HoverPopupHandler(props: FloatingMenuProps) {
                 align: "start",
                 location: "top",
                 isTargetAbsolute: true,
-                yOffset: -10,
+                yOffset: 0,
                 xOffset: -30
               }
             });


echoriiku:
Maybe just the 100 ms (or even 150 ms) delay is enough? I personally like the way it _looked_ before.

PR #9518, opened 2026-03-13
web: escape html in note title in diff viewer

01zulfi:
## Description

Fixes https://github.com/streetwriters/notesnook/security/advisories/GHSA-45g3-cv93-q59v

We should escape html in note title in diff viewer because of `dangerouslySetInnerHTML`

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [ ] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/web/src/components/diff-viewer/index.tsx
+++ b/apps/web/src/components/diff-viewer/index.tsx
@@ -35,6 +35,7 @@ import { UnlockView } from "../unlock";
 import { getFormattedDate } from "@notesnook/common";
 import { diff } from "diffblazer";
 import { strings } from "@notesnook/intl";
+import { escapeUTF8 } from "entities";
 
 type DiffViewerProps = { session: ConflictedEditorSession | DiffEditorSession };
 function DiffViewer(props: DiffViewerProps) {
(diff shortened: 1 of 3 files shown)


apps/web/src/components/diff-viewer/index.tsx
```diff
@@ -419,3 +420,12 @@ async function createCopy(note: Note, content: ContentItem, title?: string) {
...
 }
+
+function escapeHtml(s: string) {
+  return s
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&#039;");
+}
```
thecodrr: Instead of this, let's use `entities` package.

thecodrr (changes requested):

apps/web/src/components/diff-viewer/index.tsx
01zulfi: Done, using `escapeUTF8` from `entities` now

PR #9565, opened 2026-03-24
web: fix command palette highlighting being escaped

01zulfi:
## Description

Fix command palette highlighting being escaped

# before

image

# after

image

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [ ] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
+++ b/apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
@@ -249,7 +249,7 @@ export const CommandPaletteDialog = DialogManager.register(
                 >
                   <Text
                     variant="subBody"
-                    dangerouslySetInnerHTML={{ __html: escapeUTF8(label) }}
+                    dangerouslySetInnerHTML={{ __html: label }}
                   />
                 </Box>
               );
@@ -306,7 +306,7 @@ export const CommandPaletteDialog = DialogManager.register(
                         overflow: "hidden"
                       }}
                       dangerouslySetInnerHTML={{
-                        __html: escapeUTF8(command.title)
+                        __html: command.title
                       }}
                     />
                   </Flex>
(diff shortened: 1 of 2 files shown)


apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -451,9 +451,13 @@ function useDatabaseFuzzySearch() {
...
         const items = collections[type];
         if (!items) continue;
+        const escapedItems = items.map((item) => ({
+          ...item,
+          title: escapeUTF8(item.title)
+        }));
         for (const item of fuzzy(
           query,
-          items,
+          escapedItems,
```
thecodrr: You also need to escape the query so escaped characters can be matched.

thecodrr (changes requested):
I am thinking this kind of usage can still be problematic. What if we switch to using: https://interweave.dev/

I think it'll help us get rid of all `dangerouslySetInnerHTML`s.

thecodrr:
Scratch that. Interweave is unmaintained.

apps/web/src/dialogs/command-palette/command-palette-dialog.tsx
```diff
@@ -451,9 +451,14 @@ function useDatabaseFuzzySearch() {
         const type = _type as keyof typeof collections;
         const items = collections[type];
         if (!items) continue;
+        const escapedItems = items.map((item) => ({
+          ...item,
+          title: escapeUTF8(item.title)
+        }));
```
thecodrr: We should escape the items once on fetch (in `updateCollections`) instead of on each key stroke.

01zulfi:
> Can you try replacing these escapeUTF8 with a DOMPurify based solution?

I looked into [DOMPurify](https://github.com/cure53/DOMPurify), it is mostly used for sanitization. I'm not sure we want that as we just want to escape user generated html?

thecodrr:
True. Let's merge this and explore it later.

PR #9583, opened 2026-03-26
editor: fix table column toolbar not showing in correct position

01zulfi:
## Description

This PR fixes two bugs:

1. The table column toolbar wasn't showing at the correct position. This was introduced when we removed Simplebar, but continued to find the simplebar container to figure out the position
2. When a new column was created, the table wouldn't scroll to it if the new column was not visible. This is fixed for both web and mobile

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [X] Mobile
- [X] Desktop

## Sign-off
- [x] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/editor/src/extensions/table/actions.ts
+++ b/packages/editor/src/extensions/table/actions.ts
@@ -178,6 +178,7 @@ function selectRow(
       : currentCellIndex + rect.map.width;
   const pos = rect.map.map[nextCellIndex - 1];
   tr.setSelection(new TextSelection(tr.doc.resolve(rect.tableStart + pos + 1)));
+  tr.scrollIntoView();
   return true;
 }
 


packages/editor/src/extensions/table/component.tsx
```diff
@@ -62,6 +60,35 @@ export function TableComponent(
...
+      } else if (cellRect.left < containerRect.left) {
+        scrollContainer.scrollLeft -= containerRect.left - cellRect.left;
+      }
+    }
+
+    editor.on("selectionUpdate", onSelectionUpdate);
+    return () => {
+      editor.off("selectionUpdate", onSelectionUpdate);
+    };
+  }, []);
```
thecodrr: This feels largely unnecessary. Just do `.scrollIntoView()` for the `tr` in `selectColumn`/`selectRow`.

packages/editor/src/extensions/table/component.tsx
```diff
@@ -285,19 +310,21 @@ function TableColumnToolbar(props: TableToolbarProps) {
...
       columnToolsRef.current.style.top = `${pos.top}px`;
     }
 
+    const scrollContainer = table?.current?.parentElement;
+
     editor.on("selectionUpdate", onSelectionUpdate);
+    scrollContainer?.addEventListener("scroll", onSelectionUpdate);
     return () => {
       editor.off("selectionUpdate", onSelectionUpdate);
+      scrollContainer?.removeEventListener("scroll", onSelectionUpdate);
```
thecodrr, line 296: This works but I am not sure about the performance implications...

01zulfi: What about firing an event for this case? Then we can call this handler only when that event is raised instead of scrolling

packages/editor/src/extensions/table/actions.ts
```diff
@@ -192,6 +192,7 @@ function selectColumn(
     direction === "prev" ? currentCellIndex - 1 : currentCellIndex + 1;
   const pos = rect.map.map[nextCellIndex - 1];
   tr.setSelection(new TextSelection(tr.doc.resolve(rect.tableStart + pos + 1)));
+  tr.scrollIntoView();
```
thecodrr, line 196: Can you also add it for `selectRow`?

packages/editor/src/extensions/table/component.tsx
thecodrr: Not sure how we'll raise the event...

01zulfi: when a new column is added, we raise a browser event, similar to how we do `permissionrequest` event in the editor package

thecodrr: It won't be accurate. We want an event that operates based on the DOM changes instead of changes in ProseMirror state. Let's keep this for now. We can look into it later if it becomes an issue.

PR #9709, opened 2026-04-14
web: fix RCE via stored XSS in note html export

01zulfi:
## Description

Fixes https://github.com/streetwriters/notesnook/security/advisories/GHSA-fjm8-jg78-89h4

* sandbox the export iframe
* escape fields when building html template

## QA Scope

Should be tested on both web and mobile.

If a note's title and headline is a HTML, it should be correctly exported as HTML.
For example, set `RCE test  again` as note's title and headline and export the note as PDF/HTML.

Also use similar HTML strings to name colors and tags and then link them to the note. Then export the note as PDF/HTML and it should work as expected

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [X] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/web/src/common/export.ts
+++ b/apps/web/src/common/export.ts
@@ -54,6 +54,7 @@ export async function exportToPDF(
     iframe.style.height = "0";
     iframe.style.border = "0";
     iframe.loading = "eager";
+    iframe.sandbox.add("allow-same-origin", "allow-modals");
 
     iframe.onload = async () => {
       if (!iframe.contentWindow) return;
--- a/packages/core/src/utils/templates/html/template.ts
+++ b/packages/core/src/utils/templates/html/template.ts
@@ -19,6 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import { TemplateData } from "../index.js";
 import { formatDate } from "../../date.js";
+import { escapeUTF8 } from "entities";
 
 export function template(data: TemplateData) {
   return `<!DOCTYPE html>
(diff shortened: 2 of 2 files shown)


apps/web/src/common/export.ts
```diff
@@ -54,6 +54,7 @@ export async function exportToPDF(
     iframe.style.height = "0";
     iframe.style.border = "0";
     iframe.loading = "eager";
+    iframe.sandbox.add("allow-same-origin", "allow-scripts", "allow-modals");
```
01zulfi: These are the minimum permissions I found which are required for the export to work, let me know if this needs any changes

packages/core/src/utils/templates/html/template.ts
thecodrr: Colors and tags should be HTML escaped too, I think.

thecodrr (changes requested):

apps/web/src/common/export.ts
thecodrr: We can't add `allow-scripts` as that makes the whole sandbox thing pointless.

thecodrr (changes requested):

apps/web/src/common/export.ts
01zulfi: removed

thecodrr (approved):

PR #9718, opened 2026-04-16
web: improve attaching attachments UX

01zulfi:
## Description

This PR improves the UX of attaching bulk attachments by showing a unified dialog with all states of the attaching progress. Single files will also be attached through this flow. Encryption and attaching to editor hasn't been changed as well.

Example of the new dialog: (compression & encryption manually throttled)

https://github.com/user-attachments/assets/df2077e9-2064-4a9e-9995-e0f3363156d1

## Type of Change
- [ ] Bug fix
- [X] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [ ] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed


apps/web/src/components/editor/picker.ts
```diff
@@ -124,6 +136,64 @@ export async function reuploadAttachment(
...
+        });
+        file = new FileWithURI([compressed], file.name, {
+          lastModified: file.lastModified,
+          type: file.type
+        });
+      } catch (e) {
+        yield {
+          type: "error",
+          index: i,
+          error: (e as Error).message || "Compression failed"
```
thecodrr: Use intl strings here.

apps/web/src/components/editor/picker.ts
```diff
@@ -35,69 +35,81 @@ import Config from "../../utils/config";
...
-  return await attachFiles(files, type === "*/*");
+  if (!files || files.length === 0) return;
+
+  return await attachFilesWithBatchDialog(files, editorId, type === "*/*");
 }
 
-export async function attachFiles(
+export async function attachFilesWithBatchDialog(
   files: File[],
+  editorId: string,
```
thecodrr: I don't think we should couple the editor logic with this. The previous behavior of returning the `attachments` array was good. That allowed the user of `attachFiles` to decide what to do with the attachments.

thecodrr (changes requested):
This is a good implementation but I am left wondering about the actual utility of a specially designed batch attachment dialog instead of just using the TaskManager to show progress for multiple attachments. The only difference I see is the ability to toggle compression for each attachment separately but I think we can just add support for multiple files in our ImagePickerDialog.

That will simplify a lot of things without us having to change the whole attachment flow.

01zulfi:
> using the TaskManager to show progress for multiple attachments.

I did not think about that, will look into it

> we can just add support for multiple files in our ImagePickerDialog.

that is true, when a user uploads multiple images, the ImagePickerDialog can be repurposed to support multiple entries

But when user uploads different kind of attachments in bulk, the experience could still be fragmented with these changes suggested. Imagine a user uploads 2 images and 3 text files, the ImagePickerDialog will first open to ask for compression, and then the TaskManager's dialog will open for encryption. With a unified dialog, we can provide a consistent UX

apps/web/src/components/editor/picker.ts
01zulfi: The new `attachFiles` still returns only the attachments. I added the editor coupling to pass the `editorId` to the dialog as I wanted to have the `editor.attachFile` happen inside the dialog rather than outside of it

thecodrr: Let's keep the editor logic outside. It'll keep things explicit.

thecodrr (approved):

PR #9737, opened 2026-04-20
desktop: Stop urgent reminder causes segfault & app freeze on Linux.

NeedsChloesure:
## Description
This patch prevents a segfault on Linux by blocking off the `shell.beep` call due to an Electron issue https://github.com/electron/electron/issues/49436. The Linux app will consume all the available memory, then segfaults after running out of available memory to consume. 

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

Tests were not added or modified. I don't think they're necessary. 

## Platform

- [ ] Web
- [ ] Mobile
- [x] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/web/src/stores/reminder-store.ts
+++ b/apps/web/src/stores/reminder-store.ts
@@ -31,6 +31,7 @@ import {
 } from "@notesnook/core";
 import { store as noteStore } from "./note-store";
 import { ReminderPreviewDialog } from "../dialogs/reminder-preview-dialog";
+import { getPlatform } from "../utils/platform";
 
 class ReminderStore extends BaseStore<ReminderStore> {
   reminders: VirtualizedGrouping<Reminder> | undefined = undefined;
@@ -101,9 +102,7 @@ function scheduleReminder(id: string, reminder: Reminder, cron: string) {
       return;
     }
 
-    if (IS_DESKTOP_APP && !IS_LINUX) {
-      // why not do this on Linux? Attempting to bring the app to the front causes the app to freeze here and leak memory. 
-      // Once there is no more memory left to leak, the app segfaults.
+    if (IS_DESKTOP_APP) {
       const tag = await desktop?.integration.showNotification.query({
         title: reminder.title,
         body: reminder.description ?? "",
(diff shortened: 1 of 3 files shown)


apps/web/src/stores/reminder-store.ts
```diff
@@ -101,7 +101,9 @@ function scheduleReminder(id: string, reminder: Reminder, cron: string) {
       return;
     }
 
-    if (IS_DESKTOP_APP) {
+    if (IS_DESKTOP_APP && !IS_LINUX) {
+      // why not do this on Linux? Attempting to bring the app to the front causes the app to freeze here and leak memory. 
+      // Once there is no more memory left to leak, the app segfaults.
```
thecodrr: The guard should be on line 122 no:

```
 await desktop?.integration.bringToFront.query();
```

apps/web/vite.config.ts
```diff
@@ -42,6 +42,7 @@ const isBeta = version.includes("-beta");
 const isTesting =
   process.env.TEST === "true" || process.env.NODE_ENV === "development";
 const isDesktop = process.env.PLATFORM === "desktop";
+const isLinux = process.platform === "linux";
```
thecodrr: This is a build constant i.e. it'll be `darwin` if linux desktop app is built from macOS. Better to use the `getPlatform()` from `utils/platform.ts`

thecodrr (changes requested):

apps/web/src/stores/reminder-store.ts
NeedsChloesure: I was wrong about the actual cause of the issue, I just assumed it was this behavior because of Wayland being used on my machine. The attempt to bring the window to the front is completely fine and benign and does nothing, the real cause is the use of `shell.beep()` in the desktop notification flow, see electron/electron#49436

This issue is fixed in Electron 40 (allegedly), but I wasn't able to easily get that running so I gave up there.

I added the check to the responsible app/desktop file instead.

apps/web/src/stores/reminder-store.ts
```diff
@@ -31,6 +31,7 @@ import {
 } from "@notesnook/core";
 import { store as noteStore } from "./note-store";
 import { ReminderPreviewDialog } from "../dialogs/reminder-preview-dialog";
+import { getPlatform } from "../utils/platform";
```
thecodrr:
```suggestion
```

thecodrr (approved):

PR #9779, opened 2026-04-30
common: don't prefix attachment hash in filenames during export

### LINKED ISSUE #9427: Markdown import issues: images imported perfectly but the log makes you think otherwise + minor DATE bug
\### What happened?  
  
\### importing Markdown with images/files works flawlessly, but the log presented to the user makes you think it didn't import them  
When you import a Markdown file in a zip that includes attachments (e.g. like in the format as Notesnnook exports it), Notesnook imports it \*\*completely successfully\*\* \_including the files/images/etc\_.  
  
But the importer screen shows you something like this, which, to me, made me think that it was \_not\_ importing the attachments (emphasis added):  
\`\`\`  
Unpacking 1 file(s) ...  
Unpacked into 4 ...  
Preprocessing complete ...  
SKIPPING md\_note.zip ...  
SKIPPING attachments ...  
SKIPPING 9c4f2a7b6e1d3f08-attach1.jpg ...  
SKIPPING 5b7e1c9a3d2f4b0e-attach2.jpg ...  
Processing Example --- note-to-export.md ...  
Pushing Example - note to export into data  
\`\`\`  
And then it gives the \_final\_ message: "Import completed. 1 notes \*\*\*successfully\*\*\* imported.0 errors occurred" (note I didn't make a typo, there is no space between the period and zero "imported.0", so must be a typo in code).  
  
This log message is confusing, because while it said "1 notes successfully imported", I thought it had failed to import all the images (it "skipped" them). I thought "ah this didn't \_fully\_ work I guess", but then, as I mentioned above, \*\*all the images were correctly imported\*\* and the note looks perfect.  
  
\### date importation bug with Frontmatter input (using Notesnook's \_own\_ export)  
In testing the above, I noticed that Notesnook switches days for months from the Frontmatter part of the Markdown file. And I mean for a note that was exported by \_Notesnook\_ itself, \*\*not\*\* from another program. See steps below  
  
\### Steps to reproduce the problem  
  
\### steps for the image import incorrect messaging  
\- \*\*First\*\* get a Markdown file with images in zip format.  
\- I just took an existing Notesnook note that had two images and exported it to Markdown + Frontmatter from within the program.  
\- Go to the Notesnook importer, select Markdown and drop in the file, and click Start importing.  
\- You will have the "Logs" in front of you of the sort I mentioned above "skipping xyz.jpg", but the note will be perfectly imported  
  
\### steps for the date importation bug  
1\. Create a new note, e.g. today March 1st 2026 from inside Notesnook.  
2\. Export it with the option Markdown + Frontmatter.  
3\. The markdown file will have lines like this at the top: \`created\_at: 01-03-2026 11:14 AM\` and \`updated\_at: 01-03-2026 11:15 AM\`  
4\. I then deleted the note and deleted it from the trash to just have a clean situation.  
5\. Import the exported note back into Notesnook.  
6\. Then note will be imported successfully, but its dates will be \_swapped\_: The note will say \_January\_ \*\*3rd\*\*, \_not\_ March 1st. i.e. the importer expected the date where the month was, and the month where the date was.  
  
But it was Notesnook \_itself\_ that exported this very same note, so it should already be in the order that the Notesnook importer would expect. So either Notesnook is exporting the date incorrectly (month should be first, not day), or importing the date incorrectly (month should be second).  
  
\### suggestion for imported Markdown from Notesnook's own-export for filenames  
I noticed in the above process that for the Markdown with images that I imported (for which the Markdown was created by Notesnook itself, not using another program), Notesnook named the imported images with the filenames as the were named from inside the zip file (in my case "9c4f2a7b6e1d3f08-attach1.jpg" & "5b7e1c9a3d2f4b0e-attach2.jpg").  
  
Notesnook when it exported the note as Markdown, didn't name the images with the original filenames ("attach1.jpg" & "attach2.jpg"), but rather with the xxh64 hash put in front of the actual filename -- this export behavior of Notensook is fully logical, because you can have two images with the same name across different notes (or even the same note), so it gets differentiated by the xxh64 hash so as not to create any filename conflicts in the zip file. So the exporting to Markdown in this regard is working well.  
  
The minor issue here is that, upon importing this Notesnook exported-note back into Notesnook, Notesnook will name the files \_inside the program\_ as: 9c4f2a7b6e1d3f08-attach1.jpg & 5b7e1c9a3d2f4b0e-attach2.jpg (\_not\_ attach1.jpg & attach2.jpg).  
  
What is the problem? If you later export that note \_again\_ from Notesnook, now the files in the new export will get named:  
9c4f2a7b6e1d3f08-9c4f2a7b6e1d3f08-attach1.jpg  
5b7e1c9a3d2f4b0e-5b7e1c9a3d2f4b0e-attach2.jpg  
  
The case where this would be relevant is if you needed to import a note(s) that were exported from Notesnook, back into Notesnook. For example, another Notesnook user exported a note, and shares it with you -- you then make substantial edits, and export it back to Markdown and share it back with them. Or you have some own notes of yours exported into Markdown from Notesnook, and you need to restore them into your account. If you do this a couple of times, you'll have filenames like 5b7e1c9a3d2f4b0e-5b7e1c9a3d2f4b0e-5b7e1c9a3d2f4b0e-attach2.jpg.  
  
\*\*Suggestion\*\*  
Perhaps the importer can have some check associated with attachments of a Markdown zip, and if the attachment being imported has a filename that includes a xxh64 hash as the prefix to an actual filename, then it detects that this as a Notesnook naming convention, and so then the filename that gets registered inside the database drops the hash part (e.g. the imported filename wouldn't be "9c4f2a7b6e1d3f08-attach1.jpg" but "attach1.jpg"). Of course, if the filename has no "real" filename associated with it (e.g. the filename itself is something like 5b7e1c9a3d2f4b0e.jpg), then the importer could just name it 5b7e1c9a3d2f4b0e.jpg inside the database.  
  
\### Version  
  
v3.3.8  
  
\### Platform/OS  
  
Windows  
  
\### Relevant log output  
  
\`\`\`shell  
Unpacking 1 file(s) ...  
Unpacked into 4 ...  
Preprocessing complete ...  
Skipping md\_note.zip ...  
Skipping attachments ...  
Skipping 9c4f2a7b6e1d3f08-attach1.jpg ...  
Skipping 5b7e1c9a3d2f4b0e-attach2.jpg ...  
Processing Example --- note-to-export.md ...  
Pushing Example - note to export into data  
\`\`\`  
  
  
Issue created on GitHub By [code-consensus](https://github.com/code-consensus)

01zulfi:
## Description

Closes https://github.com/streetwriters/notesnook/issues/9427

Now when a note is exported, the attachment filenames won't have the hash prefix.

## QA Scope

Needs to be tested on web only.

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [X] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/common/src/utils/export-notes.ts
+++ b/packages/common/src/utils/export-notes.ts
@@ -102,7 +102,10 @@ export async function* exportNotes(
 
   // case where the user has a notebook named attachments
   const attachmentsRoot = pathTree.add("attachments", "underscore");
-  const pendingAttachments: Map<string, Attachment> = new Map();
+  const resolvedAttachments: Map<
+    string,
+    { path: string; attachment: Attachment }
+  > = new Map();
 
   for (const [id] of notePathMap) {
     const note = await database.notes.note(id);
@@ -120,7 +123,8 @@ export async function* exportNotes(
         unlockVault: options.unlockVault,
         format,
         attachmentsRoot,
-        pendingAttachments,
+        resolvedAttachments,
+        attachmentTree: pathTree,
         resolveInternalLink: (link) => {
           const internalLink = parseInternalLink(link);
           if (!internalLink) return link;
@@ -152,7 +156,7 @@ export async function* exportNotes(
     }
   }
 
-  for (const [path, attachment] of pendingAttachments) {
+  for (const { path, attachment } of resolvedAttachments.values()) {
     yield <ExportableAttachment>{
       type: "attachment",
       path,
@@ -178,13 +182,18 @@ export async function* exportNote(
   });
   const ext = FORMAT_TO_EXT[options.format];
   const path = [filename, ext].join(".");
-  const pendingAttachments: Map<string, Attachment> = new Map();
+  const resolvedAttachments: Map<
+    string,
+    { path: string; attachment: Attachment }
+  > = new Map();
+  const attachmentTree = new PathTree();
 
   try {
     const content = await exportContent(note, {
       format,
       attachmentsRoot,
-      pendingAttachments,
+      resolvedAttachments,
+      attachmentTree,
       unlockVault: options.unlockVault
     });
     if (!content) return false;
@@ -197,7 +206,7 @@ export async function* exportNote(
       ctime: new Date(note.dateCreated)
     };
 
-    for (const [path, attachment] of pendingAttachments) {
+    for (const { path, attachment } of resolvedAttachments.values()) {
       yield <ExportableAttachment>{
         type: "attachment",
         path,
@@ -225,7 +234,8 @@ export async function exportContent(
 
     // TODO: remove these
     attachmentsRoot?: string;
-    pendingAttachments?: Map<string, Attachment>;
+    resolvedAttachments?: Map<string, { path: string; attachment: Attachment }>;
+    attachmentTree?: PathTree;
     resolveInternalLink?: ResolveInternalLink;
   }
 ) {
@@ -234,7 +244,8 @@ export async function exportContent(
     unlockVault,
     resolveInternalLink,
     attachmentsRoot,
-    pendingAttachments,
+    resolvedAttachments,
+    attachmentTree,
     disableTemplate
   } = options;
   const rawContent = await database.content.findByNoteId(note.id);
@@ -274,7 +285,7 @@ export async function exportContent(
 
   if (
     attachmentsRoot &&
-    pendingAttachments &&
+    resolvedAttachments &&
     format !== "txt" &&
     format !== "pdf"
   ) {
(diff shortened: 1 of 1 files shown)


packages/common/src/utils/export-notes.ts
```diff
@@ -226,6 +233,8 @@ export async function exportContent(
     // TODO: remove these
     attachmentsRoot?: string;
     pendingAttachments?: Map<string, Attachment>;
+    attachmentTree?: PathTree;
+    attachmentPathMap?: Map<string, string>;
```
thecodrr: I think having both `pendingAttachments` and `attachmentPathMap` is a bit redundant.

packages/common/src/utils/export-notes.ts
```diff
@@ -286,15 +297,30 @@ export async function exportContent(
...
+            format
+          );
+          continue;
+        }
+
+        const baseFilename = attachment.filename || attachment.hash;
+        const rawAttachmentPath = join(attachmentsRoot, baseFilename);
+        const attachmentPath = attachmentTree
+          ? attachmentTree.add(rawAttachmentPath)
+          : rawAttachmentPath;
```
thecodrr: This can be simplified to:

const attachmentPath = attachmentTree?.add(rawAttachmentPath) || rawAttachmentPath.

packages/common/src/utils/export-notes.ts
```diff
@@ -286,15 +297,30 @@ export async function exportContent(
...
-        const attachmentPath = join(attachmentsRoot, filename);
+        if (attachmentPathMap?.has(attachment.hash)) {
+          sources[attachment.hash] = resolveAttachment(
+            elements,
+            attachment,
+            attachmentPathMap.get(attachment.hash)!,
+            format
+          );
+          continue;
+        }
```
thecodrr: Is this added so that the same attachment across multiple notes links to the same file on export?

thecodrr (changes requested):

packages/common/src/utils/export-notes.ts
01zulfi: yep

PR #9788, opened 2026-05-01
editor: add input & menu to select different font sizes

01zulfi:
## Description

image

## QA Scope

Test on web and mobile.

## Type of Change
- [ ] Bug fix
- [X] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [X] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/editor/src/toolbar/tools/font.tsx
+++ b/packages/editor/src/toolbar/tools/font.tsx
@@ -32,16 +32,12 @@ import { strings } from "@notesnook/intl";
 import { Flex, Input } from "@theme-ui/components";
 import { ToolButton } from "../components/tool-button.js";
 import { ResponsivePresenter } from "../../components/responsive/index.js";
+import { FONT_SIZE_BOUNDS } from "@notesnook/common";
 
 const FONT_SIZES = [
   8, 9, 10, 11, 12, 14, 16, 18, 20, 22, 24, 28, 32, 36, 42, 48, 56, 64, 72, 80,
   96, 120
 ];
-
-const FONT_SIZE_BOUNDS = {
-  MIN: 8,
-  MAX: 120
-};
 
 export function FontSize(props: ToolProps) {
   const { editor } = props;
@@ -65,6 +61,7 @@ export function FontSize(props: ToolProps) {
       inputRef.current.value = String(currentSize);
       return;
     }
+
     const clamped = Math.min(
       FONT_SIZE_BOUNDS.MAX,
       Math.max(FONT_SIZE_BOUNDS.MIN, parsedValue)
--- a/packages/editor/src/toolbar/tools/font.tsx
+++ b/packages/editor/src/toolbar/tools/font.tsx
@@ -203,7 +203,7 @@ export function FontSize(props: ToolProps) {
               return;
             }
 
-            applyInputValue();
+            if (inputRef.current) inputRef.current.value = String(currentSize);
             setIsMenuOpen(false);
           }}
         />
--- a/packages/editor/src/toolbar/tools/font.tsx
+++ b/packages/editor/src/toolbar/tools/font.tsx
@@ -17,22 +17,27 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { ToolProps } from "../types.js";
-import { Editor } from "../../types.js";
-import { Dropdown } from "../components/dropdown.js";
-import { MenuItem } from "@notesnook/ui";
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import {
-  useToolbarStore,
-  useToolbarLocation
-} from "../stores/toolbar-store.js";
-import { getFont, getFontById, getFonts } from "../../utils/font.js";
-import { CodeBlock } from "../../extensions/code-block/index.js";
+import { FONT_SIZE_BOUNDS, toTitleCase } from "@notesnook/common";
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/menu/index.tsx
```diff
@@ -226,7 +226,12 @@ function MenuContainer(props: PropsWithChildren<MenuContainerProps>) {
           {title}
         </Text>
       )}
-      <ScrollContainer suppressScrollX>{children}</ScrollContainer>
+      <ScrollContainer
+        suppressScrollX
+        onWheel={(e) => e.stopPropagation()}
+      >
+        {children}
+      </ScrollContainer>
```
01zulfi, line 237: scrolling in the font sizes menu would also scroll the editor toolbar

this change fixes that

thecodrr (changes requested):
I think we should go with the Google Docs UX i.e. an input with a dropdown menu as well:

Image

packages/editor/src/toolbar/tools/font.tsx
```diff
@@ -21,45 +21,232 @@ import { ToolProps } from "../types.js";
...
+import { ResponsivePresenter } from "../../components/responsive/index.js";
+
+const FONT_SIZES = [
+  8, 9, 10, 11, 12, 14, 16, 18, 20, 22, 24, 28, 32, 36, 42, 48, 56, 64, 72, 80,
+  96, 120
+];
+
+const FONT_SIZE_BOUNDS = {
+  MIN: 8,
+  MAX: 120
```
thecodrr: Let's use 1px as min and 400px as max, as used by Google Docs.

packages/editor/src/toolbar/tools/font.tsx
```diff
@@ -21,45 +21,232 @@ import { ToolProps } from "../types.js";
...
+       */
+      if (containerRef.current?.contains(target)) return;
+      if ((target as HTMLElement).closest?.(".popup-presenter")) return;
+
+      setIsMenuOpen(false);
+    }
+
+    document.addEventListener("mousedown", handleMouseDown);
+    return () => document.removeEventListener("mousedown", handleMouseDown);
+  }, [isMenuOpen]);
```
thecodrr, line 103: Why not just make the menu `blocking`?

packages/ui/src/components/menu/use-focus.ts
```diff
@@ -93,12 +94,16 @@ export function useFocus(
           default:
             break;
         }
-        if (nextIndex !== i) moveItemIntoView(nextIndex);
+        if (nextIndex !== i) {
+          moveItemIntoView(nextIndex);
+          const focusedItem = items[nextIndex];
+          if (focusedItem) onKeyboardFocus?.(focusedItem);
```
thecodrr, line 100: I don't fully understand what `onKeyboardFocus` is for.

thecodrr (changes requested):

packages/editor/src/toolbar/tools/font.tsx
01zulfi: I want to be able to interact with the input field without closing the menu

packages/ui/src/components/menu/use-focus.ts
01zulfi: In Google Docs, when we navigate with arrowup/arrowdown in the font sizes menu, it sets the menu item as the input value. I replicated that behavior here.

For that I needed a callback to run when a menu item is focused via a keyboard press i.e. arrowup/arrowdown

thecodrr: Side effect: up/down keys change the value and if you close the menu by clicking away, it applies the value to the text instead of cancelling/discarding it. The change should only be applied in 2 cases:

1. User presses Enter in input
2. User clicks/selects a menu item

thecodrr (changes requested):
Side effect: up/down keys change the value and if you close the menu by clicking away, it applies the value to the text instead of cancelling/discarding it. The change should only be applied in 2 cases:

- User presses Enter in input
- User clicks/selects a menu item

01zulfi:
> Side effect: up/down keys change the value and if you close the menu by clicking away, it applies the value to the text instead of cancelling/discarding it. The change should only be applied in 2 cases: 1. User presses Enter in input 2. User clicks/selects a menu item

Updated behavior:

∙ * if input had last interaction, on enter set input value as font size
∙ * else if menu had last interaction (arrow up/down), on enter set focused value as font size

thecodrr:
> > Side effect: up/down keys change the value and if you close the menu by clicking away, it applies the value to the text instead of cancelling/discarding it. The change should only be applied in 2 cases: 1. User presses Enter in input 2. User clicks/selects a menu item
> 
> Updated behavior:
> 
> ∙ * if input had last interaction, on enter set input value as font size ∙ * else if menu had last interaction (arrow up/down), on enter set focused value as font size

Not sure if I understand correctly but won't arrow up/down automatically change the input value? Why do we need special handling for both of these cases? Why can't we just handle the `Enter` as using the input value no matter what?

01zulfi:
Oh my bad, looks like I misunderstood. I thought we should remove the input value changing by navigating the menu with  arrow up/down.

I'll bring that back, and set the input value as font size only if the user hits `Enter`

thecodrr (approved):

thecodrr (approved):

PR #9879, opened 2026-05-19
web: fix day picker to allow selecting all months within min/max range

01zulfi:
## Description

web: fix day picker to allow selecting all months within min/max range

## QA Scope

Needs to tested on web only. Need to verify the following pickers work as expected (with correct disabled states and allowing correct selection within min/max range):
1. Picker in notes expiry dialog
2. Picker when editing note's creation date
3. Picker when creating reminders
4. and other pickers in the webapp

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [ ] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/web/src/components/day-picker/index.tsx
+++ b/apps/web/src/components/day-picker/index.tsx
@@ -16,6 +16,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
+
 import { useDatePicker } from "@rehookify/datepicker";
 import { Box, Button, Flex, Text } from "@theme-ui/components";
 import { useState } from "react";
@@ -31,6 +32,7 @@ type DayPickerProps = {
 export function DayPicker(props: DayPickerProps) {
   const { selected, sx, maxDate, minDate, onSelect } = props;
   const [selectedDates, onDatesChange] = useState<Date[]>([selected]);
+
   const {
     data: { weekDays, months, years, calendars },
     propGetters: { dayButton, addOffset, subtractOffset, setOffset }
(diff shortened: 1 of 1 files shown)


apps/web/src/components/day-picker/index.tsx
```diff
@@ -158,39 +170,73 @@ export function DayPicker(props: DayPickerProps) {
...
+                maxDate.getDate()
+              )
+            : undefined;
+          const cleanMin = minDate
+            ? new Date(
+                minDate.getFullYear(),
+                minDate.getMonth(),
+                minDate.getDate()
+              )
+            : undefined;
```
thecodrr: No need to do this inside `.map`. Do it once outside and use the values inside `.map`.

thecodrr (changes requested):
I am not sure this fixes what we think it fixes. I tried reproducing this issue in production but failed. Here's a screenshot:

Image

As you can see, I can select dates upto May 20, 2027 which means the min/max limit is working correctly.

I could not, however, jump between years which is the actual bug.

01zulfi:
@thecodrr, thanks for pointing out that we can still navigate via chevron buttons. 

I've revisited the fix, and I believe that we just need to clamp dates during the month/year dropdown selectors.

apps/web/src/components/day-picker/index.tsx
```diff
@@ -31,6 +32,24 @@ type DayPickerProps = {
...
+
+    if (normalizedMaxDate && d > normalizedMaxDate) {
+      return normalizedMaxDate;
+    }
+    if (normalizedMinDate && d < normalizedMinDate) {
+      return normalizedMinDate;
+    }
+
+    return d;
+  }
```
thecodrr: This function should be outside the component with `minDate` and `maxDate` passed as parameters.

thecodrr (changes requested):

thecodrr (approved):

PR #9899, opened 2026-05-25
web: compress image in realtime for better UX in attach files dialog

01zulfi:
## Description

web: compress image in realtime for better UX in attach files dialog

## QA Scope

Test on web only. Needs comprehensive testing on compression in the dialog with all options in image compression behavior settings

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [ ] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/web/src/dialogs/attach-files-dialog.tsx
+++ b/apps/web/src/dialogs/attach-files-dialog.tsx
@@ -17,10 +17,9 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import Dialog from "../components/dialog";
-import { ScrollContainer } from "@notesnook/ui";
-import { Box, Flex, Image, Switch, Text } from "@theme-ui/components";
+import { Flex, Image, Switch, Text } from "@theme-ui/components";
 import { formatBytes } from "@notesnook/common";
 import { BaseDialogProps, DialogManager } from "../common/dialog-manager";
 import { strings } from "@notesnook/intl";
@@ -32,7 +31,6 @@ import { ImageCompressionOptions } from "../stores/setting-store";
 import { Attachment } from "@notesnook/editor";
 import {
   CheckCircle,
-  Loading,
   File as FileIcon,
   CloseCircle
 } from "../components/icons";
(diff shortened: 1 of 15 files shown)


01zulfi (commented):
I feel like there might be a way to improve the `useEffects` in the `attach-files-dialog`, but not sure how. @thecodrr, do take a look

apps/web/src/dialogs/attach-files-dialog.tsx
```diff
@@ -108,32 +160,46 @@ export const AttachFilesDialog = DialogManager.register(
     }, []);
 
     useEffect(() => {
-      if (showCompressionPrompt || processingRef.current) return;
+      (async () => {
+        await Promise.all(
+          fileStates.map(async (state, index) => {
+            await compress(index, state);
+          })
+        );
```
thecodrr: Using `Promise.all` here is a bad idea. We should probably use something like `p-queue` or `p-map` with limited concurrency to avoid locking the main thread if there are a lot of images.

apps/web/src/dialogs/attach-files-dialog.tsx
```diff
@@ -79,8 +81,58 @@ export const AttachFilesDialog = DialogManager.register(
...
+    const [isCompressionComplete, setIsCompresionComplete] = useState(false);
     const processingRef = useRef(false);
 
+    async function compress(index: number, state: FileState) {
+      if (state.compressedFile) return;
+      if (!state.compress) return;
+
+      setFileStates((prev) =>
+        prev.map((s, idx) =>
+          idx === index ? { ...s, status: "compressing" } : s
```
thecodrr: If we have the `index`, why not just update the item at the index instead of mapping over the whole array?

apps/web/src/dialogs/attach-files-dialog.tsx
```diff
@@ -108,32 +160,46 @@ export const AttachFilesDialog = DialogManager.register(
...
 
-      processingRef.current = true;
+    useEffect(() => {
+      if (
+        !isCompressionComplete ||
+        showCompressionPrompt ||
+        processingRef.current
+      ) {
+        return;
+      }
```
thecodrr: Splitting file compression and other processing across 2 `useEffect`s is a bad idea. Just do them sequentially in a single `useEffect`. That will also help us get rid of these unnecessary checks and, perhaps, a few `useState`s.

apps/web/src/dialogs/attach-files-dialog.tsx
```diff
@@ -108,32 +160,46 @@ export const AttachFilesDialog = DialogManager.register(
...
 
-      const shouldCompress: boolean[] = fileStates.map((s) => !!s.compress);
+      processingRef.current = true;
 
       (async () => {
         const attachments: Attachment[] = [];
-        let hasError = false;
+        let hasError = fileStates.some((s) => s.status === "error");
```
thecodrr: `hasError` is needed at the end so why evaluate it here?

apps/web/src/dialogs/attach-files-dialog.tsx
```diff
@@ -108,32 +160,46 @@ export const AttachFilesDialog = DialogManager.register(
...
+        const validIndices = fileStates
+          .map((s, i) => (s.status === "error" ? -1 : i))
+          .filter((i) => i !== -1);
+        const filesToAttach = validIndices.map((i) => {
+          const state = fileStates[i];
+          return state.compress && state.compressedFile
+            ? state.compressedFile
+            : state.file;
+        });
```
thecodrr: These 2 can be merged. We don't really need `validIndices` because one good thing of using `for..await` is that the output order = input order.

thecodrr (changes requested):

thecodrr (approved):

PR #9923, opened 2026-06-01
web: refresh tag & notebook filtered lists on item rename

01zulfi:
## Description

web: refresh tag & notebook filtered lists on item rename

## QA Scope

Web only

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [ ] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/web/src/views/notebooks.tsx
+++ b/apps/web/src/views/notebooks.tsx
@@ -38,6 +38,7 @@ export function Notebooks() {
   const roots = useStore((store) => store.notebooks);
   const [filteredNotebooks, setFilteredNotebooks] =
     useState<VirtualizedGrouping<NotebookType>>();
+  const inputRef = useRef<HTMLInputElement>(null);
   const treeRef =
     useRef<
       VirtualizedTreeHandle<{ notebook: NotebookType; totalNotes: number }>
@@ -61,6 +62,13 @@ export function Notebooks() {
 
   useEffect(() => {
     treeRef.current?.refresh();
+
+    const query = inputRef.current?.value.trim();
+    if (!query) return;
+
+    (async () => {
+      setFilteredNotebooks(await db.lookup.notebooks(query).sorted());
+    })();
   }, [roots]);
 
   return (
@@ -165,6 +173,7 @@ export function Notebooks() {
         )}
       </Box>
       <Input
+        ref={inputRef}
         variant="clean"
         placeholder="Filter notebooks..."
         sx={{ borderTop: "1px solid var(--border)", mx: 0 }}
--- a/apps/web/src/views/tags.tsx
+++ b/apps/web/src/views/tags.tsx
@@ -23,22 +23,30 @@ import Placeholder from "../components/placeholders";
 import { db } from "../common/db";
 import { ListLoader } from "../components/loaders/list-loader";
 import { Flex, Input } from "@theme-ui/components";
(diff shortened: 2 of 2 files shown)


apps/web/src/views/notebooks.tsx
```diff
@@ -63,6 +64,18 @@ export function Notebooks() {
...
+    const query = inputRef.current?.value.trim();
+    if (!query) {
+      setFilteredNotebooks(undefined);
+      return;
+    }
+
+    (async () => {
+      setFilteredNotebooks(await db.lookup.notebooks(query).sorted());
+    })();
+  }, [roots]);
```
thecodrr, line 72: I think these 2 should be combined. Otherwise, a race condition can occur.

apps/web/src/views/notebooks.tsx
```diff
@@ -165,11 +178,12 @@ export function Notebooks() {
...
       </Box>
       <Input
+        ref={inputRef}
         variant="clean"
         placeholder="Filter notebooks..."
         sx={{ borderTop: "1px solid var(--border)", mx: 0 }}
-        onChange={debounce(async (e) => {
-          const query = e.target.value.trim();
+        onChange={debounce(async () => {
+          const query = inputRef.current?.value.trim();
```
thecodrr: This seems unnecessary.

apps/web/src/views/notebooks.tsx
```diff
@@ -63,6 +64,18 @@ export function Notebooks() {
     treeRef.current?.refresh();
   }, [roots]);
 
+  useEffect(() => {
+    const query = inputRef.current?.value.trim();
+    if (!query) {
+      setFilteredNotebooks(undefined);
+      return;
+    }
```
thecodrr: Why are we changing state here?

apps/web/src/views/tags.tsx
```diff
@@ -23,23 +23,34 @@ import Placeholder from "../components/placeholders";
...
 
   useEffect(() => {
     store.refresh();
   }, []);
 
+  useEffect(() => {
+    const query = inputRef.current?.value.trim();
+    if (!query) {
+      setFilteredTags(undefined);
+      return;
```
thecodrr: Why are we changing state here?

thecodrr (changes requested):

apps/web/src/views/notebooks.tsx
01zulfi: > I think these 2 should be combined

What 2 needs to be combined?

apps/web/src/views/notebooks.tsx
01zulfi: if query is undefined, shouldn't we set filtered to `undefined`?

apps/web/src/views/notebooks.tsx
thecodrr: The 2 useEffects

apps/web/src/views/notebooks.tsx
thecodrr: That would/should already be happening in the `onChange` event.

thecodrr (approved):

PR #9925, opened 2026-06-02
Allow any width/height value in EmbedSizeOptions when inserting embed

ammarahm-ed:
## Description

Allow any width/height value in EmbedSizeOptions when inserting embeds. Previously changing width would remove the height value and vice versa. This decouples width/height change by user so they can be any value.

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [x] Web
- [x] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/editor/src/toolbar/popups/embed-popup.tsx
+++ b/packages/editor/src/toolbar/popups/embed-popup.tsx
@@ -76,6 +76,7 @@ export function EmbedPopup(props: EmbedPopupProps) {
           let _src = src;
           let _width = size.width;
           let _height = size.height;
+
           if (embedSource === "code") {
             const document = new DOMParser().parseFromString(src, "text/html");
             if (document.getElementsByTagName("iframe").length <= 0)
@@ -103,11 +104,12 @@ export function EmbedPopup(props: EmbedPopupProps) {
           }
 
           const convertedUrl = convertUrlToEmbedUrl(_src);
-          if (!convertedUrl && embedSource === "url") {
-            return setError("Please provide a valid embed url.");
-          }
 
           if (convertedUrl) _src = convertedUrl;
+
+          if (!_src && embedSource === "url") {
+            return setError("Please provide a valid embed url.");
+          }
 
           if (_src.startsWith("javascript:")) {
             return setError("Embedding javascript code is not supported.");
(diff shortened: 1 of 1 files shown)


packages/editor/src/toolbar/popups/embed-popup.tsx
```diff
@@ -38,33 +38,30 @@ export type EmbedPopupProps = {
...
-      setHeight(size.height);
-      if (onSizeChanged) onSizeChanged(size);
+      const hasNewWidth = Number.isFinite(newWidth);
+      const hasNewHeight = Number.isFinite(newHeight);
+
+      if (!hasNewWidth && !hasNewHeight) return;
+      setSize((size) => {
+        const newSize = {
+          width: hasNewWidth ? ((newWidth || 0) as number) : size.width,
+          height: hasNewHeight ? ((newHeight || 0) as number) : size.height
```
thecodrr, line 58: Setting width/height to 0 shouldn't be allowed.

thecodrr (changes requested):

packages/editor/src/toolbar/popups/embed-popup.tsx
ammarahm-ed: User can set it in dialog and save but we have min width/height values set on the embed so it never collapses completely.

thecodrr (approved):

packages/editor/src/toolbar/popups/embed-popup.tsx
```diff
@@ -17,17 +17,18 @@ You should have received a copy of the GNU General Public License
...
-import { Input, Textarea } from "@theme-ui/components";
-import { Embed, EmbedSizeOptions } from "../../extensions/embed/index.js";
-import { convertUrlToEmbedUrl } from "@social-embed/lib";
 import { InlineInput } from "../../components/inline-input/index.js";
-import { Tabs, Tab } from "../../components/tabs/index.js";
-import { strings } from "@notesnook/intl";
+import { Tab, Tabs } from "../../components/tabs/index.js";
+import { Embed, EmbedSizeOptions } from "../../extensions/embed/index.js";
+import { Popup } from "../components/popup.js";
+import isURL from "validator/lib/isURL.js";
```
thecodrr: I don't think this is necessary. Our utility `isValidURL` operates based on if a browser can accept a URL which is more accurate. If a URL _looks_ invalid but browser is accepting it then it is a valid URL.

thecodrr (changes requested):

thecodrr (approved):

PR #9931, opened 2026-06-02
mobile: replace Input with FormInput in VaultDialog for inline field …

kashaf-ansari-dev:
## Description
Replace Input with FormInput in VaultDialog to move field-level validation inline. Password required and confirm password match errors now appear directly under the relevant field instead of being shown through ToastManager.

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

This is a UI-layer refactor with no logic changes.

## Platform

- [ ] Web
- [x] Mobile
- [ ] Desktop

### DIFF
--- a/apps/mobile/app/components/dialogs/vault/index.tsx
+++ b/apps/mobile/app/components/dialogs/vault/index.tsx
@@ -839,7 +839,6 @@ export const VaultDialog: React.FC = () => {
                 autoCapitalize="none"
                 testID={notesnook.ids.dialogs.vault.pwdAlt}
                 secureTextEntry
-                validationType="confirmPassword"
                 autoComplete="password"
                 returnKeyLabel="Create"
                 returnKeyType="done"
--- a/apps/mobile/app/components/dialogs/vault/index.tsx
+++ b/apps/mobile/app/components/dialogs/vault/index.tsx
@@ -256,12 +256,7 @@ export const VaultDialog: React.FC = () => {
         });
         close();
       } catch (e) {
-        close();
-        ToastManager.show({
-          heading: strings.passwordIncorrect(),
-          type: "error",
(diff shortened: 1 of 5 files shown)


apps/mobile/app/components/dialogs/vault/index.tsx
```diff
@@ -849,70 +793,68 @@ export const VaultDialog: React.FC = () => {
...
 
-              <Input
+              <FormInput
+                name="confirmPassword"
+                formRef={formRef}
                 fwdRef={confirmPassRef}
                 autoCapitalize="none"
                 testID={notesnook.ids.dialogs.vault.pwdAlt}
                 secureTextEntry
                 validationType="confirmPassword"
```
ammarahm-ed:
```suggestion
```
Not required

ammarahm-ed (changes requested):

ammarahm-ed (approved):

packages/core/src/api/vault.ts
```diff
@@ -203,9 +202,9 @@ export default class Vault {
     );
 
     if (password) {
-      this.password = password;
-      if (!(await this.exists())) await this.create(password);
+      await this.unlock(password).catch(() => {});
```
ammarahm-ed: use `try/catch`.

apps/mobile/app/components/dialogs/vault/index.tsx
ammarahm-ed, line 498: One more issue we need to fix is to ensure that when password is changed, biometric credentials are reset if user does not enable biometricUnlock.

thecodrr (approved):

ammarahm-ed (approved):

PR #9972, opened 2026-06-11
Remove sync conflict threshold

thecodrr:
## Description

This should fix all the cases where a conflicting edit from another device resulted in content getting unintentially and wrongly replaced. A side effect of this change is that users will experience a lot more conflicts but I think that is better than silently keeping one or the other copy. To trigger a conflict when there's an actual change, I have also improved the HTML equality check. Since

## QA Scope

Needs to be tested on web & mobile. Specifically test the case where edits are being made on 2 devices simultaneously.

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [x] Web
- [x] Mobile
- [x] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/core/src/api/sync/merger.ts
+++ b/packages/core/src/api/sync/merger.ts
@@ -33,7 +33,6 @@ import { InboxItemsHistoryErrorContext } from "../../types.js";
 import { z } from "zod";
 import { sanitizeHtml } from "../../utils/html-parser.js";
 
-const THRESHOLD = process.env.NODE_ENV === "test" ? 2 * 1000 : 60 * 1000;
 class Merger {
   logger = logger.scope("Merger");
   constructor(private readonly db: Database) {}
@@ -93,7 +92,7 @@ class Merger {
       // we can just replace the conflicted content
       const conflicted = localItem.conflicted
         ? "conflict"
-        : isContentConflicted(localItem, remoteItem, THRESHOLD);
+        : isContentConflicted(localItem, remoteItem);
 
       if (conflicted === "merge") return remoteItem;
       else if (!conflicted) return;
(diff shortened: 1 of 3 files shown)


packages/core/src/api/sync/merger.ts
```diff
@@ -148,21 +146,8 @@ export function isContentConflicted(
...
-    if (
-      timeDiff < conflictThreshold ||
-      isHTMLEqual(localItem.data, remoteItem.data)
-    ) {
-      if (remoteItem.dateModified > localItem.dateModified) {
-        return "merge";
-      }
-      return;
+    if (isHTMLEqual(localItem.data, remoteItem.data)) {
+      return remoteItem.dateEdited > localItem.dateEdited ? "merge" : undefined;
```
mrjo118, line 150: If you're going to remove the conflict threshold, wouldn't it be safer to implement a change id of some sort to detect conflicts instead of relying on timestamps? If there are even small amounts of local time drift between devices, there's a possibility that the conflict detection wont work correctly for changes made close together, if a user has the app open on 2 devices at the same time.

I worked on an implementation like this for Joplin last year, which falls back to a timestamp based check when the new column is not yet populated. The PR for the change was closed because a spec had not been discussed at the time and there was some related work in the pipeline which it should align with. But the implementation works and is conceptually sound. You can see the code for that here https://github.com/laurent22/joplin/pull/13617

thecodrr: > there's a possibility that the conflict detection wont work correctly for changes made close together

That should be covered by the `isHTMLEqual` check.

Issue #10045, opened 2026-07-01
(FR) Note creation during bidirectional note linking


the0s1s:
### What problem are you facing?

One thing that I miss from tools like Obsidian and Capacities is having the option to create a note if it doesn't exist during bidirectional linking. It's something I and many others rely on during PKM workflows.

### What's the solution you'd like to see in Notesnook?

Offer the option to create a note during bidirectional linking if it doesn't exist.

### What alternatives have you considered?

The alternative right now is to manually create the target note and then go back to the original note and link it.

### Additional context

I am willing to work on this feature and submit a pull request. I already have a working implementation in place on my personal Notesnook fork.

thecodrr:
@the0s1s that'd be cool if it integrates neatly into the current backlinking UX. Feel free to open a PR.

the0s1s:
PR sent in. Thanks.

PR #10050, opened 2026-07-02
web: create a note directly from the internal link picker

the0s1s:
## Description
The internal-link picker only searched existing notes, so typing the name of a
note that does not exist left an empty list with no way forward.

This adds a "Create note" action to the picker. When the typed name matches no
existing note (an exact, case-insensitive title check against non-trashed
notes), the picker offers to create that note and insert the `nn://note` link
in a single step.

Implementation notes:
- The existence check uses a SQLite `LIKE` match (no wildcards), which is a
  case-insensitive full-string comparison, filtered to exclude trashed notes.
  The fuzzy full-text lookup that populates the list cannot answer "does an
  exact note with this title exist", so the check is a separate query.
- On click, the action calls `db.notes.add({ title })`, refreshes the note
  store, and closes the dialog with the new note's `nn://note` link.

## Type of Change
- [ ] Bug fix
- [ x ] Feature

## Visuals
- [ X] Attached relevant screenshots / screen recording / GIF
image

## Testing
- [ ] Ran all E2E tests
- [x ] Added/updated tests for this change (if needed)

Tests were added: `apps/web/__e2e__/note-linking.test.ts` covers both paths:
one test types a title matching no note, asserts the create action appears,
creates the note, and confirms an `nn://note` link is inserted into the current
note and the new note appears in the list; the second types an existing note's
title and asserts the create action is not offered.

E2E and integration suites were not run to green locally. The web e2e suite
could not load until a separate cross-package Playwright fix (see the
`fix/web-e2e-duplicate-playwright` branch), and the local machine cannot extract
Playwright browser binaries. The suite does load and discover these tests; they
are expected to run in CI where browsers are provisioned.

## Platform
- [x ] Web
- [ ] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed


the0s1s:
Related to issue #10045

packages/intl/src/strings.ts
```diff
@@ -525,6 +525,7 @@ $day$: Current day (eg. Monday)`,
   newNotebook: () => t`New notebook`,
   newInternalLink: () => t`Link to note`,
   editInternalLink: () => t`Edit internal link`,
+  createNoteToLink: (title: string) => t`Create note "${title}"`,
```
01zulfi: Ideally we should add new strings at the end of the file.

Plus, can you run `npm run build` in the `intl` package to update the locale files?

01zulfi (changes requested):

packages/intl/src/strings.ts
the0s1s: The requested changes have been committed. Let me know if you need anything else.

apps/web/src/dialogs/note-linking-dialog.tsx
```diff
@@ -221,19 +227,65 @@ export const NoteLinkingDialog = DialogManager.register(
...
+                  // lookup above is fuzzy full-text, so it can't answer this on its own.
+                  // LIKE (no wildcards) is a case-insensitive full-string match in SQLite.
+                  const existing = await db
+                    .sql()
+                    .selectFrom("notes")
+                    .select("id")
+                    .where("type", "!=", "trash")
+                    .where("title", "like", query)
+                    .limit(1)
+                    .execute();
```
thecodrr: I don't like this here. I think we should use `db.lookup.notes` with a targeted query only for titles (e.g `title: <query>`). Or perhaps, the whole search here should be title based instead of title + content based?

thecodrr (changes requested):

apps/web/src/dialogs/note-linking-dialog.tsx
the0s1s: I changed it. Thanks for your patience.

PR #10052, opened 2026-07-03
mobile: set default font size 1-400

kashaf-ansari-dev:
## Description
Added the range of default font size in settings/editor 1-400

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

No ui or functionality change

## Platform

- [ ] Web
- [x] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/mobile/app/screens/settings/settings-data.tsx
+++ b/apps/mobile/app/screens/settings/settings-data.tsx
@@ -206,7 +206,7 @@ export const settingsGroups: SettingSection[] = [
               : status === SubscriptionStatus.ACTIVE
                 ? strings.subRenewOn(expiryDate)
                 : status === SubscriptionStatus.CANCELED ||
-                    status === SubscriptionStatus.PAUSED
+                  status === SubscriptionStatus.PAUSED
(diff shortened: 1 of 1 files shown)


apps/mobile/app/screens/settings/settings-data.tsx
```diff
@@ -862,8 +862,8 @@ export const settingsGroups: SettingSection[] = [
             name: strings.defaultFontSize(),
             description: strings.defaultFontSizeDesc(),
             type: "input-selector",
-            minInputValue: 8,
-            maxInputValue: 120,
+            minInputValue: 1,
+            maxInputValue: 400,
```
01zulfi: these bounds are already defined in the `common` package, I think we should use from there  https://github.com/streetwriters/notesnook/blob/b2936301754ebe3574073771e0b7b6ceabc878c1/packages/common/src/utils/font.ts#L20

01zulfi (changes requested):

PR #10054, opened 2026-07-03
core: move inbox item processing outside the sync loop

01zulfi:
## Description

core: move inbox item processing outside the sync loop

## QA Scope

- Test on web & mobile
- Ensure a large number of inbox items do not cause the sync to get stuck or error out
- Ensure all the inbox items get processed properly and show up after the sync completes. 

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [X] Web
- [X] Mobile
- [X] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/core/src/api/index.ts
+++ b/packages/core/src/api/index.ts
@@ -58,7 +58,6 @@ import {
   IFileStorage,
   IStorage,
   KVStorageAccessor,
-  PendingSyncItemsAccessor,
   StorageAccessor
 } from "../interfaces.js";
 import TokenManager from "./token-manager.js";
@@ -166,8 +165,9 @@ class Database {
 
   private _kv = new KVStorage(this.databaseReady.promise);
   kv: KVStorageAccessor = () => this._kv;
-  private _pendingSyncItems = new PendingSyncItems(this.databaseReady.promise);
-  pendingSyncItems: PendingSyncItemsAccessor = () => this._pendingSyncItems;
+  pendingSyncItems = new PendingSyncItems(
+    this.sql as unknown as DatabaseAccessor<RawDatabaseSchema>
+  );
   private _config: ConfigStorage = new ConfigStorage(
     this.databaseReady.promise
   );
--- a/packages/core/src/api/index.ts
+++ b/packages/core/src/api/index.ts
@@ -58,6 +58,7 @@ import {
   IFileStorage,
   IStorage,
   KVStorageAccessor,
+  PendingSyncItemsAccessor,
   StorageAccessor
 } from "../interfaces.js";
 import TokenManager from "./token-manager.js";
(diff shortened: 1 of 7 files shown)


packages/core/src/api/sync/index.ts
```diff
@@ -159,6 +159,7 @@ export class Sync {
   devices;
   private conflictedNoteIds: string[] = [];
   private uncachedAttachments: DownloadableFile[] = [];
+  private pendingInboxItems: SyncInboxItem[] = [];
```
thecodrr: We can't store it in memory. It'll need to be persisted so that in case of an interruption data is not lost.

thecodrr (changes requested):

packages/core/src/api/sync/index.ts
01zulfi: Won't it just get synced again in that case?

thecodrr: I mean what if the user closes the app while items have downloaded but not yet processed?

packages/core/src/api/index.ts
```diff
@@ -164,6 +166,8 @@ class Database {
 
   private _kv = new KVStorage(this.databaseReady.promise);
   kv: KVStorageAccessor = () => this._kv;
+  private _pendingSyncItems = new PendingSyncItems(this.databaseReady.promise);
+  pendingSyncItems: PendingSyncItemsAccessor = () => this._pendingSyncItems;
```
thecodrr: Any reason we are making this lazy?

thecodrr (changes requested):

packages/core/src/api/index.ts
01zulfi: hmm, not really. I just followed the pattern for kv storage https://github.com/streetwriters/notesnook/pull/10054/changes#diff-0920a6ece979373823bf5d4384cff2d0838b9684112aadac38160a907f2eb57bL165

thecodrr: KV Storage needs to be lazy since it is sometimes accessed before the database is initialised so it waits for the db using the `LazyDatabaseAccessor` but that's not required for PendingSyncItems.

packages/core/src/database/migrations.ts
```diff
@@ -496,6 +496,18 @@ export class NNMigrationProvider implements MigrationProvider {
...
             .execute();
         }
+      },
+      "a-2026-07-06": {
+        async up(db) {
+          await db.schema
+            .createTable("pendingsyncitems")
+            .ifNotExists()
+            .addColumn("id", "text", (c) => c.primaryKey().unique().notNull())
+            .addColumn("type", "text")
```
thecodrr, line 506: Let's add an index on `type` column.

thecodrr (changes requested):

thecodrr (approved):

PR #10088, opened 2026-07-12
feat: Add Recent Searches feature in search bar

### LINKED ISSUE #9437: Add support for recent searches

Currently, the app does not store any recently searched terms making it harder to quickly search a frequently needed term.

The recent searches should be shown on this screen when you tap on `Search in Notes` etc.

Image

The recent searches need to have the following functions:

1. There should be a X button to delete a specific item.
2. We should show a total of 10 recent searches.
3. There should be a button at the top of the list to clear all recent searches.

A rough wireframe looks like this:

Image

None

### Additional context

_No response_

afsuyadi:
## Description

Closes #9437 

1. Need to have a feature of _recent searches_, which can save up to 10 searched keywords.
2. Everytime a user types a keyword and then press _ENTER_, that keyword will be saved into the _recent searches_ feature.
3. User is able to press _Clear All_ button to remove all the keywords.

## Type of Change
- [  ] Bug fix
- [✔] Feature

## Visuals
- [✔] Attached relevant screenshots / screen recording / GIF
- [  ] N/A (not a feature or no UI changes)

## Testing
- [  ] Ran all E2E tests
- [  ] Ran all integration tests
- [  ] Added/updated tests for this change (if needed)
- [✔] N/A (tests not needed — explanation provided below)

The test are purely from UI testing.
image

After changes:
[Screencast from 12-07-26 11:31:44.webm](https://github.com/user-attachments/assets/3e2b21de-0f77-4590-9ef1-563e51286e58)

_(Additional note: Please ignore the weird symbols above the default strings. I don't know why it appears on my Notenook's local)_

## Platform

- [✔] Web
- [✔] Mobile
- [✔] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/mobile/app/screens/search/index.tsx
+++ b/apps/mobile/app/screens/search/index.tsx
@@ -24,22 +24,31 @@ import {
   VirtualizedGrouping
 } from "@notesnook/core";
 import { strings } from "@notesnook/intl";
+import { useThemeColors } from "@notesnook/theme";
 import React, { useEffect, useRef, useState } from "react";
+import { ScrollView, Text, TouchableOpacity, View } from "react-native";
 import { DatabaseLogger, db } from "../../common/database";
 import List from "../../components/list";
+import { IconButton } from "../../components/ui/icon-button";
 import SelectionHeader from "../../components/selection-header";
 import { useNavigationFocus } from "../../hooks/use-navigation-focus";
 import { ToastManager, eSubscribeEvent } from "../../services/event-manager";
 import { NavigationProps } from "../../services/navigation";
 import useNavigationStore from "../../stores/use-navigation-store";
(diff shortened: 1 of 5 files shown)


afsuyadi:
Hi @ammarahm-ed , for this PR's issue, I think I need a detailed requirement regarding on how to saved searches work. 

At my first change, every time I type a letter onto the search bar, it directly saves it, even though I haven't finished typing the word. (Example: I intend to search "notes", but when I type the first letter "n" and be idle for a few seconds, it directly saves that "n" letter as saved searches)

I then modify the change because it feels inconvenient, so that the typed words will be saved ONLY if it's **entered**. So If I typed "notes", and then saved it, it will appear in the recent searches.

Which one do you prefer?

PR #10107, opened 2026-07-15
web: redesign auth pages

01zulfi:
## Type of Change
- [ ] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [ ] Web
- [ ] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed


apps/web/src/components/error-text/index.tsx
```diff
@@ -59,3 +59,42 @@ export function ErrorText(props: ErrorTextProps) {
     </Flex>
   );
 }
+
+export function AuthErrorText(props: ErrorTextProps) {
```
thecodrr, line 63: Is this temporary?

apps/web/src/stores/user-store.ts
```diff
@@ -38,6 +38,12 @@ class UserStore extends BaseStore<UserStore> {
   user?: User = undefined;
   counter = 0;
 
+  private _pendingLogin?: {
+    email: string;
+    password: string;
+    sessionExpired: boolean;
+  };
```
thecodrr, line 45: What's the point of this?

apps/web/src/stores/user-store.ts
```diff
@@ -112,30 +118,41 @@ class UserStore extends BaseStore<UserStore> {
...
+          this._pendingLogin = undefined;
+          await db.user.authenticatePassword(
+            email,
+            password,
+            undefined,
+            sessExp
+          );
+        }
 
+        Config.set("encryptBackups", true);
```
thecodrr, line 155: I don't like this. We should use the proper way which is to add a new login method in the `core` which uses `password` grant as discussed earlier.

thecodrr (changes requested):

apps/web/src/components/error-text/index.tsx
01zulfi: Yes, this will eventually be replace by `ErrorText` once all its usages are redesigned.

apps/web/src/stores/user-store.ts
01zulfi: To facilitate the new flow without introducing any new changes to `core`

apps/web/src/stores/user-store.ts
01zulfi: There are some backend changes required to enable the new login flow. This was discussed with @ammarahm-ed, where it was decided that we will add the `core` changes after the backend has been sorted out. For now, I've hardcoded the flow on the `web` side (similar process is also done for mobile as well)

thecodrr:
This also doesn't match with the design.

apps/web/src/stores/user-store.ts
thecodrr: You are right. Let's keep it like this then.

thecodrr (changes requested):
The hover here looks weird:

Image

Secondly, set `line-height` to `100%` for buttons. That will fix the extra height issue.

01zulfi:
@thecodrr done

thecodrr:
@01zulfi line-height is still not set to 100%

thecodrr (changes requested):
A few UX issues:

1. Logging in with a wrong password shows the "password incorrect" error on the 2FA page instead of taking the user back to the email/password view. I think we should fix this properly in the backend by changing the login logic.
2. Pressing "Go back" from the 2FA screen doesn't retain the email/password

thecodrr (approved):

PR #10166, opened 2026-07-28
mobile: fix images sometimes not saving after insert

### LINKED ISSUE #10091: Uncompressed image not displaying in note
### What happened?

User is running into an issue where large images are intermittently disappearing from my notes,
The Issue:
When I attach a large, uncompressed image to a note, it looks fine at first. However, when I come back to that note later, the picture is completely missing.
Key Details:
 * No Error Messages: No errors or warnings pop up when uploading the image, and there is no indication of a sync failure later. The image just vanishes silently.
 * Note Content: The rest of the note text and formatting remains perfectly intact—only the image disappears.

Max image size is 3Mb 
Device:  OnePlus 12

--

### Version

3.4.5

### Platform/OS

Android

### Relevant log output

```shell

```

### LINKED ISSUE #10135: Photo taken via in-app camera (or selected via disk/URL) is permanently lost — not saved to gallery, not in Notesnook attachments — when note title fails to auto-generate on Android
### What happened?
On Android, inserting a file via "Attachment" auto-populates the note title with a timestamp, and the note saves correctly. Inserting an image via the "Image" option (tested across all three sub-options: upload from disk, take a photo, attach from URL) does not auto-populate the title. The title only auto-generates if the user taps an empty area of the note afterward, which is treated as a new-line trigger. If the user instead taps directly on the image or presses Back without tapping an empty area first, the note disappears entirely — and the image is not recoverable, either from Manage Attachments or, in the camera case, from the device Gallery.

### Steps to reproduce (Case A — Bug triggers, confirmed on all 3 sub-options):
1. Open Notesnook and create a new note.
2. Tap "Image" and choose disk / camera / URL.
3. Confirm the image is inserted into the note body.
4. Tap directly on the image, or press Back without tapping any empty area first.
5. Return to the notes list. Note is gone.
6. Check Manage Attachments. Image is not listed.
7. (Camera only) Check device Gallery, photo is not there either.

### Steps to reproduce (Case B — Workaround, no bug):
1. Repeat steps 1–3 above.
2. Tap an empty area of the note (e.g. below the image).
3. Title auto-populates.
4. Press Back, then see that note, title, and image all save correctly.

### Expected behavior:
The note title should auto-populate as soon as an image is inserted, regardless of where the user taps afterward, consistent with the "Attachment" flow. Captured photos should also persist independently of note save state.

### Actual behavior:
Title stays empty unless the user taps an empty area first. If not, the note and image are both lost with no recovery path in Manage Attachments or the Gallery.

### Environment:
App version: 3.4.7
OS: Android 16
Permissions granted to app: Notification, Camera (Storage/Media permission is unavailable)

### Possible cause:
The app currently only holds Camera and Notification permissions, no Storage/Media permission. Captured images may be held in a temporary cache location rather than persisted, so if the title-trigger never fires, that file may be discarded with nothing retained.

kashaf-ansari-dev:
## Description
Fixes a bug where content changes tagged ignoreEdit: true (most notably image/attachment inserts) were sometimes silently not saved to the note.
Closes https://github.com/streetwriters/notesnook/issues/10135
Closes https://github.com/streetwriters/notesnook/issues/10091

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [ ] Web
- [x] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/mobile/app/screens/editor/tiptap/use-editor.ts
+++ b/apps/mobile/app/screens/editor/tiptap/use-editor.ts
@@ -260,7 +260,6 @@ export const useEditor = (
       id,
       data,
       type,
-      ignoreEdit,
       sessionHistoryId: currentSessionHistoryId,
       tabId,
       pendingChanges,
@@ -342,11 +341,6 @@ export const useEditor = (
         };
 
         noteData.title = title;
-
-        if (ignoreEdit) {
-          DatabaseLogger.log("Ignoring edits...");
-          noteData.dateEdited = note?.dateEdited;
-        }
 
         if (data) {
           noteData.content = {
@@ -984,7 +978,6 @@ export const useEditor = (
       title,
       content,
       type,
-      ignoreEdit,
       noteId,
       tabId,
       pendingChanges,
@@ -995,7 +988,6 @@ export const useEditor = (
       title?: string;
       content?: string;
       type: string;
-      ignoreEdit: boolean;
       tabId: string;
       pendingChanges?: boolean;
       sourceNoteId?: string;
@@ -1005,7 +997,6 @@ export const useEditor = (
         `saveContent... title: ${!!title}, content: ${!!content}, noteId: ${noteId}`
       );
       if (
-        ignoreEdit ||
         lock.current ||
         (currentLoadingNoteId.current &&
           currentLoadingNoteId.current === noteId)
@@ -1014,7 +1005,6 @@ export const useEditor = (
 
           lock.current: ${lock.current}
           currentLoadingNoteId.current: ${currentLoadingNoteId.current}
-          ignoreEdit: ${ignoreEdit}
         `);
         if (lock.current) {
           setTimeout(() => {
@@ -1051,7 +1041,6 @@ export const useEditor = (
         data: content,
         type: "tiptap",
         id: noteId,
-        ignoreEdit,
         sessionHistoryId: noteId ? editorSessionHistory.get(noteId) : undefined,
         tabId: tabId,
         pendingChanges,
@@ -1080,7 +1069,7 @@ export const useEditor = (
             saveNote(params);
           }
         },
-        ignoreEdit ? 0 : 150
+        150
       );
     },
     [editorSessionHistory, withTimer, onChange, saveNote]
--- a/packages/editor-mobile/src/hooks/useEditorController.ts
+++ b/packages/editor-mobile/src/hooks/useEditorController.ts
@@ -216,8 +216,7 @@ export function useEditorController({
         return;
       }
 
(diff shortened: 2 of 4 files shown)


apps/mobile/app/screens/editor/tiptap/use-editor.ts
```diff
@@ -949,7 +949,6 @@ export const useEditor = (
         `saveContent... title: ${!!title}, content: ${!!content}, noteId: ${noteId}`
       );
       if (
-        ignoreEdit ||
```
ammarahm-ed, line 1008: We cannot remove this, this can cause unintended automatic edits in user's note content.

ammarahm-ed (changes requested):

packages/editor-mobile/src/hooks/useEditorController.ts
```diff
@@ -216,6 +216,13 @@ export function useEditorController({
         logger("info", "Edit skipped, tab is in loading state");
         return;
       }
+
+      const timerPending = !!timers.current.change;
+      if (ignoreEdit && timerPending) {
+        logger("info", "Ignoring ignoreEdit update, a save is already pending");
+        return;
+      }
+
```
ammarahm-ed:
```suggestion
      if (ignoreEdit) {
        logger("info", "Ignoring ignoreEdit update, a save is already pending");
        return;
      }

```
We always skip ignoreEdit. It gets discarded when reaches React Native Side from webview.

packages/editor-mobile/src/hooks/useEditorController.ts
```diff
@@ -224,12 +231,14 @@ export function useEditorController({
         tabRef.current.session?.noteId
       );
       if (!editor) return;
-      if (typeof timers.current.change === "number") {
-        clearTimeout(timers.current?.change);
+      if (timerPending) {
+        clearTimeout(timers.current?.change as any);
       }
```
ammarahm-ed: Revert this.

packages/editor-mobile/src/hooks/useEditorController.ts
```diff
@@ -224,12 +231,14 @@ export function useEditorController({
...
 
       timers.current.change = setTimeout(async () => {
         htmlContentRef.current = editor.getHTML();
+        timers.current.change = null;
+
```
ammarahm-ed: Revert.

ammarahm-ed (changes requested):

ammarahm-ed (approved):

PR #10168, opened 2026-07-28
mobile: preserve line breaks in shared plain text

### LINKED ISSUE #9207: On android, appending to a note through the sharing function should start with a new line, not space
When I come across something I want to save to notesnook, I use the android share function and append to a note, the appended content appear in the note with a space in between, which make appending through multiple time creates a long line, looks badly format.
My suggestion is new content appended should be with a new line character before it, not space, so that each appended new content stay on their own line.

_______________
**Device information:**
App version: 3.3.12
Platform: android
Device: OnePlus-CPH2465-14
Pro: true
Logged in: yes
Github Release: Yes

kashaf-ansari-dev:
## Description
Fixes an issue where shared plain text containing line breaks was rendered as a single paragraph in the share editor. The plain text to HTML conversion now preserves line breaks by converting each line into a separate paragraph before HTML encoding.
Fixes https://github.com/streetwriters/notesnook/issues/9207

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

Verified manually by sharing multiline text from another app and confirming that line breaks and paragraph formatting are preserved in the share editor and saved note.

## Platform

- [ ] Web
- [x] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/mobile/app/share/share.tsx
+++ b/apps/mobile/app/share/share.tsx
@@ -97,9 +97,10 @@ function makeHtmlFromUrl(url: string) {
 function makeHtmlFromPlainText(text: string) {
   if (!text) return "";
 
-  return `<p>${encodeHTML5(text)
-    .replace(/[\n]+/g, "\n")
-    .replace(/(?:\r\n|\r|\n)/g, "</p><p>")}</p>`;
+  return `<p>${text
+    .split(/\r\n|\r|\n/)
+    .map((line) => encodeHTML5(line))
+    .join("</p><p>")}</p>`;
 }
 
 type DefaultNote = {
--- a/apps/mobile/app/share/share.tsx
+++ b/apps/mobile/app/share/share.tsx
@@ -97,10 +97,10 @@ function makeHtmlFromUrl(url: string) {
 function makeHtmlFromPlainText(text: string) {
   if (!text) return "";
 
-  return `<p>${text
+  return text
     .split(/\r\n|\r|\n/)
-    .map((line) => encodeHTML5(line))
-    .join("</p><p>")}</p>`;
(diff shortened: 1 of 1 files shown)


apps/mobile/app/share/share.tsx
```diff
@@ -97,9 +97,10 @@ function makeHtmlFromUrl(url: string) {
 function makeHtmlFromPlainText(text: string) {
   if (!text) return "";
 
-  return `<p>${encodeHTML5(text)
-    .replace(/[\n]+/g, "\n")
-    .replace(/(?:\r\n|\r|\n)/g, "</p><p>")}</p>`;
+  return `<p>${text
+    .split(/\r\n|\r|\n/)
+    .map((line) => encodeHTML5(line))
+    .join("</p><p>")}</p>`;
```
ammarahm-ed: I think this will break if there were more than 2 lines with `\r\n` in them. Instead each line should be wrapped in p tag I think.

ammarahm-ed (changes requested):

apps/mobile/app/share/share.tsx
kashaf-ansari-dev: Tested it this was already correct: join("p tags") closes each line's tag and opens the next one in between, with the outer template literal adding the opening and closing for the whole string. So for N lines you get N tags either way, not just for 2.

But now i made it more explicit, now each line is wrapped in its own p tag inside the map, the result is same

ammarahm-ed: Yes you are right, it was working fine before too.

PR #10232, opened 2026-08-13
mobile: add locale detection and language switching support

kashaf-ansari-dev:
## Description
- Detect and activate device locale on app launch via Intl API, falling back to English for unsupported locales
- Load compiled @notesnook/intl catalogs (en, fr, es, de, it) and activate the saved or detected locale in index.js
- Persist user-selected language via SettingsService (appLanguage)

## Type of Change
- [ ] Bug fix
- [x] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [ ] Web
- [x] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed


apps/mobile/app/app.tsx
```diff
@@ -63,6 +63,7 @@ const App = (props: { configureMode: "note-preview" }) => {
   useAppEvents();
   //@ts-ignore
   globalThis["IS_MAIN_APP_RUNNING"] = true;
+  useSettingStore((state) => state.settings.appLanguage);
```
ammarahm-ed: Is this needed if we are doing a full app reload?

apps/mobile/globals.js
```diff
@@ -28,21 +27,38 @@ Object.defineProperty(global, "Buffer", {
...
-  });
+import SettingsService from "./app/services/settings";
+
+const localeCatalogs = {
+  en: require("@notesnook/intl/dist/locales/$en.json").messages,
+  fr: require("@notesnook/intl/dist/locales/$fr.json").messages,
+  es: require("@notesnook/intl/dist/locales/$es.json").messages,
+  de: require("@notesnook/intl/dist/locales/$de.json").messages,
+  it: require("@notesnook/intl/dist/locales/$it.json").messages
+};
```
ammarahm-ed: We are loading all locales eagerly, instead these should be getters so on first access the locale gets loaded and cached. But first add a small `console.time` to determine how long it takes right now to load all 5 locales and whether lazily loading makes any difference or not.

apps/mobile/globals.js
```diff
@@ -28,21 +27,38 @@ Object.defineProperty(global, "Buffer", {
...
+
+  i18n.load(localeCatalogs);
+  i18n.activate(targetLang);
+  setI18nGlobal(i18n);
+  return targetLang;
 }
 
-i18n.activate("en");
-setI18nGlobal(i18n);
+initLocale();
```
ammarahm-ed: I think that we should move this to common/locale where we do all of this work, in globals.js we only call `initLocale`. a more clean approach.

apps/web/src/dialogs/settings/behaviour-settings.ts
```diff
@@ -28,13 +28,70 @@ import { TimeFormat, DayFormat } from "@notesnook/core";
...
+                message: strings.restartAppToApplyChanges
+                  ? strings.restartAppToApplyChanges()
+                  : "Restart the app to apply the changes.",
+                positiveButtonText: strings.restartNow
+                  ? strings.restartNow()
+                  : "Restart now",
+                negativeButtonText: strings.cancel
+                  ? strings.cancel()
+                  : "Cancel"
+              });
```
ammarahm-ed, line 65: If `strings.changeLanguage()` exists always, why we need a ternary operator here?

apps/web/src/utils/config.ts
```diff
@@ -24,10 +24,10 @@ function set<T>(key: string, value: T) {
   return value;
 }
 
-function get<T>(key: string, def?: T): T {
+function get<T = any>(key: string, def?: T): T {
   const value = window.localStorage.getItem(key);
   if (!value && def !== undefined) return def;
-  return value ? tryParse(value) : def;
+  return (value ? tryParse(value) : def) as T;
 }
```
ammarahm-ed, line 31: This is not needed, you should pass the correct `default` value. which is `en`.

apps/web/src/stores/setting-store.ts
```diff
@@ -110,6 +112,7 @@ class SettingStore extends BaseStore<SettingStore> {
       titleFormat: db.settings.getTitleFormat(),
       trashCleanupInterval: db.settings.getTrashCleanupInterval(),
       profile: db.settings.getProfile(),
+      appLanguage: Config.get<string | undefined>("appLanguage", undefined),
```
ammarahm-ed: The default appLanguage is not `undefined`, it is `en` and should be so.

apps/web/src/index.ts
```diff
@@ -44,14 +44,51 @@ if (theme) {
...
+]).then(([targetMessages, enMessages]) => {
+  const catalogs: Record<string, Messages> = {
+    [targetLang]: targetMessages
+  };
+  if (enMessages) {
+    catalogs.en = enMessages;
+  }
+
+  i18n.load(catalogs);
+  i18n.activate(targetLang);
```
ammarahm-ed: This should get refactored to a common location in the web app

ammarahm-ed: Why we need to load all languages if the app is going to use a single language at a time?

packages/intl/vite.config.mts
```diff
@@ -56,8 +56,7 @@ export default defineConfig({
       }
     }),
     dts({
-      exclude: ["**/locales/*.json"],
-      rollupTypes: true
```
ammarahm-ed, line 60: Why change this?

ammarahm-ed (changes requested):

apps/mobile/app/components/list/card.tsx
```diff
@@ -41,6 +42,7 @@ export const Card = ({
   );
   const announcements = useMessageStore((state) => state.announcements);
   const fontScale = Dimensions.get("window").fontScale;
+  const freshStrings = getLocalizedMessageStrings(messageBoardState.id);
```
ammarahm-ed: Rename is variable. `freshStrings` is too generic, a variable should describe what it's value is, in this case `localizedMessage` would be a better pick.

apps/mobile/app/components/list/card.tsx
```diff
@@ -90,22 +92,23 @@ export const Card = ({
...
               style={{
-                flexWrap: "nowrap",
+                flexWrap: "wrap",
                 flexShrink: 1
               }}
               size={AppFontSize.sm}
               color={colors.primary.heading}
             >
-              {messageBoardState.actionText}
+              {freshStrings?.actionText || messageBoardState.actionText}
```
ammarahm-ed: we probably don't need fallback value since localizedMessage strings should always be available.

apps/mobile/app/components/list/card.tsx
```diff
@@ -90,22 +92,23 @@ export const Card = ({
...
             </Paragraph>
             <Paragraph color={colors.secondary.paragraph} size={AppFontSize.xs}>
-              {messageBoardState.message}
+              {freshStrings?.message || messageBoardState.message}
```
ammarahm-ed: we don't need fallback value.

apps/mobile/app/components/side-menu/side-menu-home.tsx
ammarahm-ed: We can simply remove the `pro` object and use values for title/onPress directly here.

apps/mobile/app/components/side-menu/side-menu-home.tsx
```diff
@@ -48,9 +48,10 @@ const pro = {
       context: "logged-in"
     });
   }
-};
+});
 
 export function SideMenuHome() {
+  const pro = getProItem();
```
ammarahm-ed: Remove this. all we need is title/onPress so we can move this inline.

apps/mobile/app/screens/settings/home.tsx
```diff
@@ -35,6 +35,8 @@ const keyExtractor = (item: SettingSection) => item.id;
 const Home = ({
   navigation
 }: NativeStackScreenProps<RouteParams, "SettingsHome">) => {
+  const settingsGroups = useSettingsGroups();
```
ammarahm-ed, line 38: memoize this

apps/mobile/app/services/message.tsx
```diff
@@ -98,13 +98,33 @@ const APP_MESSAGES: Message[] = [
...
+    const msg = builder();
+    return {
+      message: msg.message as string,
+      actionText: msg.actionText as string
+    };
+  }
+  return null;
 }
 
 export function setRateAppMessage() {
```
ammarahm-ed: I think there is a more simple and cleaner solution for this.

1. setMessageById on sets the `id` of the message in useMessageStore, not the full message so it never becomes stale.
2. In our component we use getMessageById() to get the message full with data & actions.
3. We won't need `getLocalizedMessageStrings` since I think it is duplicated work.

apps/desktop/src/utils/locale.ts
thecodrr, line 1: We should share the locale selection + locale init logic across platforms. It's 99% the same so makes no sense to duplicate it everywhere. It will also make it easier to add new languages.

Let's move this utility to `@notesnook/intl` and use it from there on all platforms.

apps/desktop/src/api/os-integration.ts
```diff
@@ -136,6 +137,13 @@ export const osIntegrationRouter = t.router({
 
     config.backupDirectory = result.filePaths[0];
   }),
+  setAppLanguage: t.procedure
+    .input(z.string())
+    .mutation(({ input: language }) => {
+      JSONStorage.set("appLanguage", language);
```
thecodrr: We don't use `JSONStorage` directly. Add `appLanguage` in `config.ts` and use `config` instead.

thecodrr (changes requested):

packages/intl/src/generated/locale-map.ts
ammarahm-ed, line 1: Add `generated/` to `.gitignore` so these files are never commited and always generated on build automatically

ammarahm-ed (changes requested):

packages/intl/scripts/generate-locale-loaders.mjs
```diff
@@ -0,0 +1,103 @@
...
+if (!existsSync(generatedDir)) {
+  mkdirSync(generatedDir, { recursive: true });
+}
+writeFileSync(
+  path.join(generatedDir, "locale-map.ts"),
+  generateWebLocaleMap()
+);
+
+// 2. Write Mobile locale loaders
+const mobileLocaleDir = path.join(
```
ammarahm-ed: Mobile locale files should also be in `src/generated` too instead of being written in the app dir.

packages/intl/scripts/generate-locale-loaders.mjs
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  return `${LICENSE_HEADER}
+export const LOCALE_LOADERS = {
+${entries.join(",\n")}
+};
+`;
+}
+
+// 1. Write Web/Desktop locale map
```
ammarahm-ed, line 83: src/generated should be gitignored and automatically generated on every build.

ammarahm-ed (changes requested):

PR #10269, opened 2026-08-21
core: fix quoted & repeated tag/color filters in search query

01zulfi:
## Description

core: fix quoted & repeated tag/color filters in search query
* fix extra quotes being added for multi word tag/color filter
* fix multiple tag/color filters not working

## QA Scope

Web and mobile. The following behavior needs to be tested:

1. Multi word tag and color filters should work. These filters are represented by enclosing the tag/color name in double quotes. For example: `tag:"the diary"` will search notes only in `the diary` tag
2. Multiple tag and color filters should work. For example: `tag:one tag:two` will search notes only in `one` or `two` tags

## Type of Change
- [X] Bug fix
- [ ] Feature

## Visuals
- [ X] N/A (not a feature or no UI changes)

## Testing
- [ ] Ran all E2E tests
- [X] Added/updated tests for this change (if needed)

## Platform

- [X] Web
- [X] Mobile
- [x] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/core/src/api/lookup.ts
+++ b/packages/core/src/api/lookup.ts
@@ -155,6 +155,13 @@ export default class Lookup {
           : [];
 
       const defaultVault = await this.db.vaults.default();
+
+      /**
+       * `notes` is a FilteredSelector whose `where` method mutates the selector.
+       * Thus the selector passed into `notesWithHighlighting` is mutated, which can cause unexpected results.
+       * To avoid this, we clone the selector before `where`.
+       */
+      notes = notes.clone();
       notes = notes.where((eb) => {
         const exprs = [];
         const tagsFilter = this.db.relations
--- a/packages/core/src/utils/query-transformer.ts
+++ b/packages/core/src/utils/query-transformer.ts
@@ -38,6 +38,12 @@ type FieldPhraseNode = {
   value: QueryNode;
 };
 
+type FieldToken = {
+  field?: string;
+  fieldOccurrence?: number;
+  token: string;
+};
+
 type OperatorNode = {
   type: "AND" | "OR" | "NOT";
 };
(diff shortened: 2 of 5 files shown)


packages/core/src/utils/query-transformer.ts
```diff
@@ -177,24 +198,31 @@ function tokenizeWithFields(
...
+      field !== currentField ||
+      (isArrayField && fieldOccurrence !== currentOccurrence);
+
+    if (shouldStartNewGroup) {
       if (currentTokens.length > 0) {
         groups.push({ field: currentField, tokens: currentTokens });
         currentTokens = [];
       }
       currentField = field;
+      currentOccurrence = fieldOccurrence;
```
thecodrr, line 225: Why do we care about occurrence?

thecodrr (approved):

packages/core/src/utils/query-transformer.ts
01zulfi: Because array fields were getting grouped together if they appeared next to each other. So, occurrence is being  compared to preserve separate array fields

thecodrr: What if both array fields appear the same number of times e.g. `tag: a color: a tag: b color: b`?

01zulfi: That works as expected with `tag: ['one','two']` and `color:['one','two']`

I did found another issue where one letter tag filters aren't working

packages/core/src/api/lookup.ts
```diff
@@ -112,6 +112,8 @@ export default class Lookup {
     notes: FilteredSelector<Note>,
     sortOptions?: SortOptions
   ): Promise<VirtualizedGrouping<HighlightedResult>> {
+    notes = notes.clone();
```
thecodrr: Not sure but is it a good idea to fix this here instead of doing it on the mobile side? I mean ideally the passed parameters shouldn't be mutated as that creates side effects (like we observed) so this is good in principle. However, cloning on every query does add a minimal cloning cost.

thecodrr (approved):

packages/core/src/api/lookup.ts
01zulfi: I thought about fixing it on mobile and passing in a fresh new selector every time like on web. But I think that would be an indirect fix as the parameter shouldn't be mutated at all

thecodrr (approved):

PR #10273, opened 2026-08-22
mobile: add support for the pebble index 01 ring

tannermeade:
## Description
Add an on-device Android receiver for Pebble Index 01 so Notesnook can save ring recordings without a webhook or internet.

Pebble delivers audio + transcription (and optional reminder metadata) through an explicit intent to `PebbleIndexCaptureService`. The capture is written to a private inbox, then Headless JS creates a note, attaches the `.m4a` as a playable `<audio>` block, and — when the capture is a reminder — creates a Notesnook reminder on that same note (`once` at the given deadline, or `permanent` if there is no due time). After a successful ingest, a send-sync is run so other signed-in devices get the note without opening the phone app.

Settings live under **Productivity → Pebble Index 01**:
- Receive Index captures (default on)
- Keep ready in background (optional FGS so Headless JS is warm)
- Notebook for Index notes
- Notebook for Index reminders
- Reminder alert style (silent / vibrate / urgent)

The service is exported but permission-gated (`RECEIVE_INDEX_CAPTURE`) and UID-checked to `coredevices.coreapp`. No new network permission is required for the capture path; existing Notesnook sync is used only after the note exists locally.

The PR for the Pebble Core app is here: https://github.com/coredevices/mobileapp/pull/369

## Type of Change
- [ ] Bug fix
- [x] Feature

## Visuals

Settings UI is a new Productivity screen (toggles + notebook pickers + alert style). Screenshots of **Settings → Productivity → Pebble Index 01** and a resulting note with audio + reminder can be added if desired.

## Testing
- [ ] Ran all E2E tests

This is an Android inter-app intent + Headless JS path that depends on a physical Index 01 / Pebble Core install. Existing mobile E2E does not cover third-party explicit intents or Notifee scheduling from a headless task. Behavior was verified on-device: single note per capture, playable audio attachment, reminder `once` vs `permanent`, notebook routing, background notification with Unpin, and send-sync to other devices.

## Platform
- [ ] Web
- [x] Mobile
- [ ] Desktop

Android only. iOS is unchanged.

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/mobile/android/.project
+++ b/apps/mobile/android/.project
@@ -16,12 +16,12 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1622976477541</id>
+			<id>1787398695494</id>
 			<name></name>
 			<type>30</type>
 			<matcher>
 				<id>org.eclipse.core.resources.regexFilterMatcher</id>
-				<arguments>node_modules|.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
+				<arguments>node_modules|\.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
 			</matcher>
 		</filter>
 	</filteredResources>
--- a/apps/mobile/android/.settings/org.eclipse.buildship.core.prefs
+++ b/apps/mobile/android/.settings/org.eclipse.buildship.core.prefs
@@ -1,11 +1,11 @@
-arguments=
+arguments=--init-script /var/folders/cf/lx4530nn29n9wpz2x8z1qm9c0000gn/T/db3b08fc4a9ef609cb16b96b200fa13e563f396e9bb1ed0905fdab7bc3bc513b.gradle
 auto.sync=false
 build.scans.enabled=false
 connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)
 connection.project.dir=
 eclipse.preferences.version=1
 gradle.user.home=
-java.home=/usr/lib/jvm/java-15-openjdk
+java.home=/opt/homebrew/Cellar/openjdk@17/17.0.20.1/libexec/openjdk.jdk/Contents/Home
 jvm.arguments=
 offline.mode=false
 override.workspace.settings=true
--- a/apps/mobile/android/.settings/org.eclipse.buildship.core.prefs
+++ b/apps/mobile/android/.settings/org.eclipse.buildship.core.prefs
@@ -1,11 +1,11 @@
-arguments=--init-script /var/folders/cf/lx4530nn29n9wpz2x8z1qm9c0000gn/T/db3b08fc4a9ef609cb16b96b200fa13e563f396e9bb1ed0905fdab7bc3bc513b.gradle
+arguments=
 auto.sync=false
 build.scans.enabled=false
 connection.gradle.distribution=GRADLE_DISTRIBUTION(WRAPPER)
 connection.project.dir=
 eclipse.preferences.version=1
 gradle.user.home=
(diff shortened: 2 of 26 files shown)


tannermeade:
Apparently, Core Devices doesn't want the PR on their end because "in the nearish future there will be a better way".

thecodrr:
@tannermeade Thank you for the PR. I am not sure such a niche and specific device support belongs inside Notesnook. Is there really no better/alternative way? We do have the [Inbox API](https://notesnook.com/help/inbox-api/getting-started) for this sort of stuff.

tannermeade:
> @tannermeade Thank you for the PR. I am not sure such a niche and specific device support belongs inside Notesnook. Is there really no better/alternative way? We do have the [Inbox API](https://notesnook.com/help/inbox-api/getting-started) for this sort of stuff.

This is the initial PR, before it gets merged into Notesnook it'll be best to generalize the problem/solution.

The Inbox API is a server/internet-based solution that requires internet. The problem space for this PR is local-only, no internet requirement creation of notes. It basically requires a client-side offline-first Inbox API, which in this case uses Android's intent system to facilitate inter-app communication.

The use case here is Pebble's products provide a unique note creation system and are making it very open to be integrated with other apps. This is also going to be useful for other types of note creation devices such as heypocket.com and more. These types of devices usually work by having their own app that facilitates the device communications, and can be integrated with other local apps and server APIs.

Many of the use cases for these solutions are they work reliably without always being connected to the internet. This is why integrating via the existing Inbox API is a non-solution for many, and inefficient (syncing to the internet, then downloading the note that already exists on the phone just not in the desired app).

thecodrr:
> which in this case uses Android's intent system to facilitate inter-app communication.

That works but it'll also allow any malicious app to create all sort of content because this is not authenticated. There are no safety guards. Secondly, even if we add a generalized and local Intent based API, it'll still require a lot of changes on the vendor side (or in Notesnook) to support sending/receiving things from/to Notesnook. Taking over the burden to maintain something like this (especially when we don't have any such device to test on) will need considerable effort & resources. I am still not sure if this is the best approach.

We could provide a local authenticated HTTP server (much like the Inbox API) or something else that'd allow other apps to send things to Notesnook. I think that approach has a lot more use cases but not sure if that'll work for your specific use case.

tannermeade:
> > which in this case uses Android's intent system to facilitate inter-app communication.
> 
> That works but it'll also allow any malicious app to create all sort of content because this is not authenticated. There are no safety guards. Secondly, even if we add a generalized and local Intent based API, it'll still require a lot of changes on the vendor side (or in Notesnook) to support sending/receiving things from/to Notesnook. Taking over the burden to maintain something like this (especially when we don't have any such device to test on) will need considerable effort & resources. I am still not sure if this is the best approach.
> 
> We could provide a local authenticated HTTP server (much like the Inbox API) or something else that'd allow other apps to send things to Notesnook. I think that approach has a lot more use cases but not sure if that'll work for your specific use case.

Ya, the solution should be secure and maintainable.

The protocol options are:
- local authenticated HTTP server
- intents
- file system

The local authenticated HTTP server would work for this use case as long as it's long lived (works in the background reliably) which getting that to work reliably across different Android devices can be a pain. (see https://dontkillmyapp.com)

The original motivation to use intents is they can be on-demand and short-lived, so they don't require keeping a socket open + drop any packets if the socket closes.

The other protocol would be using the file system. Many android-based Obsidian integrations send files by writing them to the file system. For Notesnook, that might look like pointing it to a folder and auto-importing notes. This may or may not work with iOS though due to how it sandboxes apps.

I'd think either intent or file system could be more battery efficient and durable to the app being killed.

PR #10321, opened 2026-09-04
mobile: check if email is confirmed on confirm-email banner press

kashaf-ansari-dev:
## Description
Fixes an issue where the "Please confirm your account" / "Sync is disabled" message banner continues to display on Android even after the user has confirmed their email in an external browser.

In cases where network reconnection latency upon returning to the app causes the background check to miss or delay, clicking the "Confirm email" banner now first checks whether the user's account has already been confirmed:
- **If confirmed:** Directly clears the banner, refreshes the token, starts sync, and triggers the celebration walkthrough without opening an unnecessary bottom sheet.
- **If not confirmed (or offline):** Opens the standard confirmation bottom sheet with instructions and the "Resend email" button.

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

This change adjusts the event handler on the mobile message banner to check user confirmation status before presenting the sheet. Verified manually on device across account creation, backgrounding, external browser confirmation, and returning to the app.

## Platform
- [ ] Web
- [x] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/apps/mobile/app/services/message.tsx
+++ b/apps/mobile/app/services/message.tsx
@@ -34,6 +34,9 @@ import SettingsService from "./settings";
 import { Update } from "../components/sheets/update";
 import { GithubVersionInfo } from "../utils/github-version";
 import { CheckVersionResponse } from "react-native-check-version";
+import { db } from "../common/database";
+
+let isCheckingEmail = false;
 
 const APP_MESSAGES: Message[] = [
   {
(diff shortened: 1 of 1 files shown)


apps/mobile/app/services/message.tsx
```diff
@@ -91,7 +94,21 @@ const APP_MESSAGES: Message[] = [
...
+        if (user?.isEmailConfirmed) {
+          clearMessage();
+          return;
+        }
+      } catch (e) {
+        /* empty */
+      } finally {
+        isCheckingEmail = false;
+      }
+
```
ammarahm-ed, line 116: We should show email confirmed sheet to user here if email is confirmed & hide the card.

ammarahm-ed (changes requested):

PR #10351, opened 2026-09-09
editor: add callout to escapable nodes at document start

### LINKED ISSUE #10341: Callout at the start of a document cannot be escaped with ArrowUp (not in escapableNodesIfAtDocumentStart)
### What happened?

When a note starts with a callout block, there is no way to place the cursor *before* it to insert content above the callout.

The editor already handles this case for other block types: the `Quirks` extension turns **ArrowUp** at the start of the document into "insert an empty paragraph before this node" via `escapableNodesIfAtDocumentStart`. That list currently contains code blocks, tables, blockquotes and lists, but not `callout`:

```ts
// packages/editor/src/index.ts
escapableNodesIfAtDocumentStart: [
  CodeBlock.name,
  Table.name,
  Blockquote.name,
  ...LIST_NODE_TYPES
]
```

So for a callout at the top of the document, ArrowUp does nothing, and pressing Enter at the start of the callout's heading only splits the heading *inside* the callout. The only workaround I found is to type the new content at the bottom, select all callouts, cut, and paste them below it.

Adding `Callout.name` to that list should be enough; the `escapeNode` handler is generic.

1. Create a new note.
2. Insert a callout as the very first block and give it a title.
3. Press Ctrl+Home (or click at the start of the callout title).
4. Press ArrowUp. Expected: an empty paragraph appears above the callout (same as for a code block or table in this position). Actual: nothing happens.
5. Press Enter instead. Actual: a new empty heading is created inside the callout rather than above it.

Compare with a code block or table in the same position: ArrowUp creates the paragraph above as expected.

### Version

3.4.7 (web, self-hosted build from `master` @ 4c4ac1c18)

### Platform/OS

Web (Chrome on Windows 10). The keymap lives in the shared editor package, so desktop should be affected identically.

### Relevant log output

```
n/a — no error, the key press is simply not handled.
```

mohdsultan18:
## Description
Added `Callout.name` to `escapableNodesIfAtDocumentStart` array so pressing ArrowUp at the start of a document inserts an empty paragraph above a callout block, matching existing behavior for code blocks, tables, blockquotes and lists.

Fixes #10341

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

One-line config change — added a node name to an existing array

## Platform

- [x] Web
- [ ] Mobile
- [x] Desktop

## Sign-off
- [x] QA passed
- [x] UI/UX passed

### DIFF
--- a/packages/editor/src/index.ts
+++ b/packages/editor/src/index.ts
@@ -365,6 +365,7 @@ const useTiptap = (
             CodeBlock.name,
             Table.name,
             Blockquote.name,
+            Callout.name,
             ...LIST_NODE_TYPES
           ]
         }),


thecodrr:
Thanks!

PR #10363, opened 2026-09-11
Add localization

thecodrr:
## QA Scope

- Test on web/desktop/mobile
- Ensure changing language changes all the strings in the app
- Ensure system language is correctly acknowledged on first launch
- Errors are currently NOT localized and should be ignored.


packages/intl/src/locale.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  let sysLocale = systemLocale;
+  if (!sysLocale) {
+    try {
+      sysLocale = Intl.DateTimeFormat().resolvedOptions().locale;
+    } catch {
+      sysLocale = "en";
+    }
+  }
```
thecodrr, line 46: @kashaf-ansari-dev this is unnecessary. Providing system locale should be mandatory.

PR #10364, opened 2026-09-11
Fix/locale detection

kashaf-ansari-dev:
## Description
Previously resolveTargetLocale silently fell back to Intl.DateTimeFormat().resolvedOptions().locale when systemLocale
wasn't provided. This masked callers (mobile, web) never passing the actual device/browser locale, and Intl's resolved locale isn't a reliable source of the user's language on all platforms.

systemLocale is now a required param on both resolveTargetLocale and InitLocaleOptions, forcing each platform to explicitly resolve and pass its own locale.

## Type of Change
- [x] Bug fix
- [ ] Feature

## Visuals

## Testing
- [ ] Ran all E2E tests

## Platform

- [x] Web
- [x] Mobile
- [ ] Desktop

## Sign-off
- [ ] QA passed
- [ ] UI/UX passed

### DIFF
--- a/packages/intl/src/locale.ts
+++ b/packages/intl/src/locale.ts
@@ -26,8 +26,8 @@ import { LOCALE_LOADERS } from "./generated/loaders.mobile";
 export { localeMap, LOCALE_LOADERS };
 
 export function resolveTargetLocale(
-  savedLanguage?: string | null,
-  systemLocale?: string
+  savedLanguage: string | null | undefined,
+  systemLocale: string
 ): string {
   if (
     savedLanguage &&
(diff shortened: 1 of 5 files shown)


packages/intl/src/locale.ts
```diff
@@ -36,45 +34,65 @@ export function resolveTargetLocale(
...
-
-  defaultI18n.load({ [targetLang]: messages });
+function activateLocale(targetLang: string) {
   defaultI18n.activate(targetLang);
   setI18nGlobal(defaultI18n);
+}
+
+export function initLocaleSync(options: InitLocaleOptions): string {
+  const targetLang = resolveAndSaveLocale(options);
+  defaultI18n.load(localeCatalogs);
```
thecodrr: What's the reason for having cached locale loading here rather than only loading the `targetLang`?

thecodrr (changes requested):

thecodrr (changes requested):
@kashaf-ansari-dev remove the `@lingui` libraries from all apps and then test once to make sure everything works correctly. The only usage of `@lingui/*` libraries should be in `@notesnook/intl`

thecodrr:
@kashaf-ansari-dev I am also noticing that `editor-mobile`, `monograph` and `theme-builder` are still using the old logic for `i18n`. Kindly update those as well.

thecodrr:
@kashaf-ansari-dev Also I don't think we need this in the mobile app:

```
<I18nProvider i18n={i18n as any}>
```

We don't use the React helpers provided by `@lingui` anywhere.

apps/mobile/app/components/ui/reminder-time/index.tsx
```diff
@@ -24,9 +24,28 @@ import { useThemeColors } from "@notesnook/theme";
...
+  if (result.startsWith("Upcoming: ")) {
+    result = `${strings.upcoming()}: ${result.slice(10)}`;
+  } else if (result.startsWith("Last: ")) {
+    result = `${strings.last()}: ${result.slice(6)}`;
+  }
+  return result
+    .replace("Today", strings.today())
+    .replace("Tomorrow", strings.tomorrow())
+    .replace("Yesterday", strings.yesterday());
+}
```
ammarahm-ed: This should be localized in `core`, intl is available there so localize the strings inside `getFormattedReminderTime`.

